PR #181, opened 2026-05-01
feat(xlsx): external workbook references — read & roundtrip

productdevbook:
## Summary

External workbook references (`[N]Sheet!Ref`) were previously silently dropped on roundtrip — the `xl/externalLinks/externalLinkN.xml` files survived in the ZIP, but the `<externalReferences>` block in `xl/workbook.xml` and the `Type=\".../externalLink\"` relationships in `xl/_rels/workbook.xml.rels` were both regenerated without them, leaving orphan parts that Excel ignores.

This PR makes external links first-class: read into a typed model, preserved on roundtrip, and re-declared in all three places that point to them.

## API

```ts
import { readXlsx } from \"hucre\";

const wb = await readXlsx(buf);
for (const link of wb.externalLinks ?? []) {
  console.log(link.target, link.targetMode, link.sheetNames);
  for (const sheet of link.sheetData) {
    for (const cell of sheet.cells) {
      console.log(cell.ref, cell.type, cell.value);
    }
  }
}

// Standalone parser, useful when you already have the XML strings
import { parseExternalLink } from \"hucre\";
const link = parseExternalLink(externalLinkXml, externalLinkRelsXml);
```

## Model

```ts
interface ExternalLink {
  target: string;                    // e.g. \"../External.xlsx\"
  targetMode?: \"External\" | \"Internal\";
  sheetNames: string[];              // declaration order
  sheetData: ExternalSheetData[];    // cached cell values per external sheet
  definedNames?: ExternalDefinedName[];
}

interface ExternalCachedCell {
  ref: string;                       // A1-style
  type: \"n\" | \"s\" | \"b\" | \"e\" | \"str\";
  value: string | number | boolean;
}
```

The 1-based index in `workbook.externalLinks` matches the `[N]` prefix in formulas like `[1]Sheet1!A1`.

## What changed in the writer side

| Hook | Change |
| --- | --- |
| `writeContentTypes` | new `externalLinkIndices?: number[]` → emits `Override` per link |
| `writeWorkbookXml` | new `externalLinkRels` arg → emits `<externalReferences>` block |
| `writeWorkbookRels` | new `externalLinkRels` arg → emits `Type=\".../externalLink\"` relationships with rIds that follow worksheets / styles / sharedStrings / theme / vbaProject / FeaturePropertyBag |
| `roundtrip` | scans `_rawEntries` for `xl/externalLinks/externalLinkN.xml` and threads the indices through |

## Test plan

[...]

PR #180, opened 2026-05-01
feat(a11y): WCAG 2.1 AA helpers — audit, alt text, summary

productdevbook:
## Summary

Adds an `a11y` namespace from the package root for generating WCAG 2.1 AA-compliant spreadsheets. Three concerns:

- **Audit** — `a11y.audit(workbook)` walks the workbook and returns a list of `A11yIssue` records with stable codes, severity, and a pinpoint location (sheet, A1 ref, image/textbox index).
- **Alt text** — `SheetImage` and `SheetTextBox` now carry `altText` and `title`, emitted as `descr=` and `title=` on `xdr:cNvPr` (the OOXML attributes assistive tech reads).
- **Sheet summary** — `sheet.a11y.summary` is promoted to `docProps/core.xml` when the workbook does not already declare a description; screen readers announce it on file open.

## API

```ts
import { writeXlsx, a11y } from "hucre";

const xlsx = await writeXlsx({
  sheets: [{
    name: "Q1 Sales",
    rows: [["Region", "Revenue"], ["EU", 12_400]],
    a11y: { summary: "Quarterly sales by region", headerRow: 0 },
    images: [{ data, type: "png", anchor: { from: { row: 0, col: 3 } },
              altText: "Bar chart showing 47% YoY growth" }],
  }],
});

// Audit (returns A11yIssue[])
for (const issue of a11y.audit(workbook)) {
  console.log(issue.type, issue.code, issue.message, issue.location);
}

// Color contrast helpers (WCAG 2.1 sRGB)
a11y.contrastRatio("0969DA", "FFFFFF");  // ≈ 4.93 (passes AA)
a11y.relativeLuminance("808080");
```

## Issue codes

| code | severity | when |
| --- | --- | --- |
| `no-doc-title` | info | `properties.title` missing |
| `no-doc-description` | warning | no description and no `sheet.a11y.summary` |
| `empty-sheet` | info | sheet has no rows or cells |
| `no-header-row` | warning | populated sheet with neither `a11y.headerRow` nor a table |
| `merged-header-row` | warning | merge range overlaps the marked header row |
| `missing-alt-text` | error / warning | image without alt text (error) or textbox without alt text (warning) |
| `low-contrast` | warning | font/fill contrast below `minContrast` (default 4.5) |
| `blank-row-in-data` | info | blank row between populated rows — screen readers may stop |

`audit(workbook, { skipContrast, minContrast, contrastSampleLimit })` lets callers tune the contrast pass.

## Test plan

[...]

PR #182, opened 2026-05-01
feat(xlsx): threaded comments — read & roundtrip (Excel 365+)

### LINKED ISSUE #158: feat: threaded comments — Excel 365+ modern comment system
## Summary

Support threaded comments (Excel 365+), distinct from legacy comments/notes.

## Technical details

Stored in `xl/threadedComments/threadedComment1.xml`:
- Thread-based replies via `parentId`
- Author info: `userId`, `providerId`, `displayName`
- Timestamps on each comment
- Excel auto-creates legacy comment placeholders for backward compat

## Scope

- [ ] Read threaded comments
- [ ] Write threaded comments
- [ ] Reply threading (parentId)
- [ ] Author + timestamp metadata
- [ ] Auto-generate legacy comment placeholders
- [ ] Round-trip preservation

## References

- MS-XLSX: https://learn.microsoft.com/en-us/openspecs/office_standards/ms-xlsx/e0fb917a-1107-409a-852f-13b47aea70dc

productdevbook:
## Summary

Adds first-class support for the **modern threaded comment** format introduced with Excel 365, distinct from the legacy VML-anchored comment system that hucre already handles.

Both formats can coexist on the same sheet (Excel writes a legacy stub for backward compat), so this PR treats them as independent surfaces — `cell.comment` for legacy notes, `sheet.threadedComments` for threads.

## API

```ts
import { readXlsx } from \"hucre\";

const wb = await readXlsx(buf);
console.log(wb.persons);  // [{ id, displayName, userId?, providerId? }, ...]

for (const sheet of wb.sheets) {
  for (const comment of sheet.threadedComments ?? []) {
    if (!comment.parentId) console.log(\"thread root at\", comment.ref);
    else console.log(\"  reply\", comment.text, \"to\", comment.parentId);
    for (const m of comment.mentions ?? []) console.log(\"  @\" + m.mentionPersonId);
  }
}
```

```ts
// Standalone parsers
import { parsePersons, parseThreadedComments } from \"hucre\";
```

## What changed

| Surface | Change |
| --- | --- |
| `Workbook.persons` | new optional `ThreadedCommentPerson[]` |
| `Sheet.threadedComments` | new optional `ThreadedComment[]` |
| `parsePersons` / `parseThreadedComments` | exported parsers |
| `reader.ts` | walks `Type=\".../person\"` (workbook) and `Type=\".../threadedComment\"` (per-sheet) |
| `writeContentTypes` | new `threadedCommentSheetIndices` + `hasPersons` options |
| `writeWorkbookRels` | new `hasPersons` parameter |
| `roundtrip.ts` | scans `_rawEntries` for both parts and re-emits the references in content-types, workbook rels, and each sheet's rels |

## Test plan

- [x] 2209 tests pass (9 new); `pnpm test` green (lint + typecheck + vitest)
- [x] `parsePersons`: required attrs, optional userId/providerId, skips entries without id+displayName
- [x] `parseThreadedComments`: thread root + reply chain, mentions, `done` accepts `1` and `true`, skips entries without id/personId, empty mentions returned as `undefined`
- [x] End-to-end: a full XLSX with threaded comments roundtrips through `openXlsx → saveXlsx` and a re-read recovers the same persons + thread structure
- [x] Output declares Override for `/xl/persons/person.xml` and `/xl/threadedComments/threadedComment1.xml`, the `Type=\".../person\"` rel in workbook.xml.rels, and the `Type=\".../threadedComment\"` rel in sheet rels

## Out of scope

Synthesizing threadedComments + persons from a model on a fresh `writeXlsx` call (i.e. without an existing source file) needs an XML serializer for both bodies. The infrastructure for the references is now in place — adding the body writers is a follow-up.

References:
- [MS-XLSX] Threaded Comments schema — https://learn.microsoft.com/en-us/openspecs/office_standards/ms-xlsx/adb84732-9fc8-48b6-bddc-6b0bcdaad940
- [MS-XLSX] Threaded Comments overview — https://learn.microsoft.com/en-us/openspecs/office_standards/ms-xlsx/e0fb917a-1107-409a-852f-13b47aea70dc

Closes #158

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -463,6 +463,57 @@ export interface SheetTextBox {
   };
 }
 
+// ── Threaded Comments (Excel 365+) ─────────────────────────────────
+
+/**
+ * A person who can author or be mentioned in threaded comments.
+ * Stored in the workbook-wide `xl/persons/person.xml` part.
+ */
+export interface ThreadedCommentPerson {
+  /** Stable GUID identifying this person within the workbook. */
+  id: string;
+  /** Display name shown in Excel's comment pane (required by the schema). */
+  displayName: string;
+  /** Identity-system user id, e.g. the Azure AD object id. */
+  userId?: string;
+  /** Identity provider name, e.g. "AD" or "PeoplePicker". */
+  providerId?: string;
+}
+
+/**
+ * An `@person` mention inside a threaded comment's text. Indices are
+ * UTF-16 code-unit offsets into the comment text.
+ */
+export interface ThreadedCommentMention {
+  mentionPersonId: string;
+  mentionId: string;
+  startIndex: number;
+  length: number;
+}
+
+/**
+ * A single message in a thread on `xl/threadedComments/threadedCommentN.xml`.
+ * Top-level messages declare a `ref`; replies omit it and link to their
+ * parent through `parentId`.
+ */
+export interface ThreadedComment {
+  id: string;
+  /** A1-style cell ref. Required for thread roots, omitted for replies. */
+  ref?: string;
+  /** GUID matching a {@link ThreadedCommentPerson.id}. */
+  personId: string;
+  /** GUID of the parent comment when this is a reply. */
+  parentId?: string;
+  /** ISO-8601 timestamp from the `dT` attribute. */
+  date?: string;
+  /** Comment body. */
+  text: string;
+  /** Whether the thread is marked resolved. */
+  done?: boolean;
+  /** `@person` mentions inside the text. */
+  mentions?: ThreadedCommentMention[];
+}
+
 // ── Image ──────────────────────────────────────────────────────────
 
 export interface SheetImage {
@@ -594,6 +645,12 @@ export interface Sheet {
   sparklines?: Sparkline[];
   /** Text boxes (shapes with text) */
   textBoxes?: SheetTextBox[];
+  /**
+   * Excel 365 threaded comments for this sheet. Stored physically in
+   * `xl/threadedComments/threadedCommentN.xml` and resolved against
+   * the workbook-wide person list (`Workbook.persons`).
+   */
+  threadedComments?: ThreadedComment[];
 }
 
 // ── Workbook Properties ────────────────────────────────────────────
@@ -633,6 +690,11 @@ export interface Workbook {
     lockStructure?: boolean;
     lockWindows?: boolean;
   };
+  /**
+   * Workbook-wide person directory referenced from threaded comments.
+   * Each `ThreadedComment.personId` resolves against this list.
+   */
+  persons?: ThreadedCommentPerson[];
 }
 
 // ── Read Options ───────────────────────────────────────────────────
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
## Review

Thorough read + roundtrip implementation for Excel 365's modern threaded-comment surface. Architecture (separate `Workbook.persons` + `Sheet.threadedComments` independent of legacy `Cell.comment`) is the right call — Excel writes both formats side by side and tying them together would have leaked complexity into the legacy reader.

### Notes / sanity checks

- **`mentionpersonId` lowercase 'p' is intentional.** Caught my eye because every other attribute in the schema is camelCase, but the MS-XLSX threadedcomments XSD really does spell it `mentionpersonId`. Test fixtures match what Excel actually writes, so this is correct.
- **`done="true"` handling** matches the schema's `xsd:boolean` while still tolerating the `"1"` form Excel emits in practice. The "any other value → undefined (i.e. not done)" semantics are sensible for a tri-state-feeling boolean.
- **`relIdNum` fallback to `0`** in `roundtrip.ts` works in practice because rIds are always `rId{N}`-shaped. If a malformed relationship file ever sneaks in, the worst case is rId1 collision — easily debuggable, fine.
- **Sheet-index-based probe (`xl/threadedComments/threadedComment{N}.xml`) for raw entries.** Relies on Excel's convention that thread parts share the worksheet's 1-based index. This holds for files Excel itself writes; if a third-party tool used a different naming, the part would be dropped from the rels on roundtrip. Worth flagging in a follow-up but not a blocker.
- **`hasPersons || undefined` in content-types options** — minor stylistic note: the `=== false` case never happens since `hasPersons` is `boolean | undefined`, but `|| undefined` keeps the option bag tidy. Fine as-is.

### Fixes pushed

The PR was conflicting against `main` after the a11y (#180) and external-workbook-links (#181) merges. Resolved cleanly by:

1. Combining the new `Sheet.{threadedComments, a11y}` and `Workbook.{persons, externalLinks}` fields rather than picking one side.
2. In `index.ts`, keeping all three new export blocks.
3. In `content-types-writer.ts`, ordering Override emission as threadedComments → persons → externalLinks (no semantic dependency, but groups Excel-365 parts together).
4. In `reader.ts`, parsing persons (workbook-wide) and externalLinks (workbook-wide) as separate steps 7c / 7d.
5. In `workbook-writer.ts`, extending `writeWorkbookRels` to take both `hasPersons` and `externalLinkRels`, and incrementing `nextRid` after the person rel so external-link rIds don't collide.
6. In `roundtrip.ts`, threading `hasPersons` through `computeExternalLinkRelStart` so external-link rIds start *after* the person rel.
7. Updated README roadmap entry: threaded comments read+roundtrip is shipped, write-from-scratch is the remaining piece.

### Verification

- `pnpm test` — all 2255 tests pass (78 files), 0 lint errors, typecheck clean.
- `pnpm build` — 687 kB / 165 files, `dist/xlsx/threaded-comments-reader.mjs` emits correctly.

LGTM, merging once CI is green.

PR #183, opened 2026-05-02
feat(xlsx): slicers & timeline filters — read & roundtrip

productdevbook:
## Summary

Adds first-class **read** support and full **roundtrip preservation** for the slicer (Excel 2010+) and timeline slicer (Excel 2013+) parts that hucre previously dropped on save.

Before this PR the slicer / timeline body files survived inside the ZIP, but the references that wired them into the workbook were silently regenerated without the declarations, so Excel saw orphan parts and dropped the slicers / timelines on next open. This PR wires them back into `[Content_Types].xml`, `xl/_rels/workbook.xml.rels`, the workbook's `extLst`, and each sheet's `_rels`.

## API

```ts
import { readXlsx, openXlsx, saveXlsx } from "hucre";

const wb = await readXlsx(buf);

// Workbook-level cache definitions.
console.log(wb.slicerCaches);    // SlicerCache[]
console.log(wb.timelineCaches);  // TimelineCache[]

// Per-sheet slicer / timeline instances.
for (const sheet of wb.sheets) {
  for (const s of sheet.slicers ?? [])    console.log(s.name, s.cache, s.caption);
  for (const t of sheet.timelines ?? [])  console.log(t.name, t.cache, t.level);
}

// Standalone parsers.
import {
  parseSlicers,
  parseSlicerCache,
  parseTimelines,
  parseTimelineCache,
} from "hucre";
```

## Model

```ts
interface Slicer {
  name: string;
  cache: string;
  caption?: string;
  columnCount?: number;
  style?: string;
  sortOrder?: string;
  rowHeight?: number;
}

interface SlicerCache {
  name: string;
  sourceName?: string;
  pivotTables?: SlicerCachePivotTable[];   // pivot-table-sourced caches
  tableSource?: SlicerCacheTableSource;    // x15:tableSlicerCache
}

interface Timeline {
  name: string;
  cache: string;
  caption?: string;
  level?: string;
  showHeader?: boolean;
  showSelectionLabel?: boolean;
  showTimeLevel?: boolean;
  showHorizontalScrollbar?: boolean;
  style?: string;
}

interface TimelineCache {
  name: string;
  sourceName?: string;
  pivotTables?: SlicerCachePivotTable[];
}
```

## What changed in the writer side

| Hook | Change |
| --- | --- |
| `ContentTypesOptions` | new `slicerIndices` / `slicerCacheIndices` / `timelineIndices` / `timelineCacheIndices` options — emit Override per part |
[...]

PR #184, opened 2026-05-02
feat(xlsx): WPS DISPIMG cell-embedded images — read & roundtrip

### LINKED ISSUE #179: feat: Support reading/writing images embedded in cells
## 📌 Feature Description
Add support for detecting, reading, and writing images embedded within worksheet cells, including:
1.  Reading images stored in cells (including position, size, and image data) from existing workbooks
2.  Writing images into cells with configurable placement (inline, floating, or anchored to cells)
3.  Basic image metadata handling (e.g., dimensions, format preservation)

## 🎯 Use Case & Motivation
This feature would enable users to:
-  Process workbooks that contain product photos, signatures, or QR codes embedded in cells
-  Generate reports with inline images (e.g., charts, logos, or visual indicators) directly from data
-  Preserve visual content when reading and modifying existing Excel files

Currently, the library does not handle cell-embedded images, which limits use cases for data visualization and report generation workflows that rely on visual assets in spreadsheets.

## 📋 Expected Behavior
-  When reading an XLSX file, images in cells should be accessible via a clear API (e.g., `worksheet.getCell(row, col).getImages()`)
-  When writing an XLSX file, users should be able to add images to specific cells with options for sizing and positioning
-  Support common image formats (PNG, JPEG, etc.) as defined in the Excel specification

## 📎 Additional Context
This enhancement would align with other spreadsheet libraries' capabilities and expand the library's use cases for business and reporting applications.

productdevbook:
## Summary

Adds first-class **read** support and full **roundtrip preservation**
for the WPS-style `xl/cellimages.xml` part — the workbook-level
registry that backs `=_xlfn.DISPIMG("<id>", 1)` formulas.

Real-world XLSX files produced by WPS / Kingsoft Office routinely
carry this part, and recent Excel versions round-trip it. Before this
PR the body XML and media binaries survived in the ZIP, but the
`workbook.xml.rels` and `[Content_Types].xml` references that wired
them in were silently regenerated without them — so on save the
relationships disappeared and Excel/WPS lost the link from the
DISPIMG formula to the bytes.

## API

```ts
import { readXlsx, openXlsx, saveXlsx } from "hucre";

const wb = await readXlsx(buf);
for (const img of wb.cellImages ?? []) {
  console.log(img.id, img.type, img.description, img.data.byteLength);
}

// Cells reference each entry through `=_xlfn.DISPIMG("<id>", 1)` —
// match the formula's first argument against `CellImage.id` to look
// up the binary.
```

```ts
// Standalone parsers, for callers that already have the XML strings.
import {
  parseCellImages,
  assembleCellImages,
  REL_CELL_IMAGES,
} from "hucre";

const refs = parseCellImages(cellImagesXml);
const images = assembleCellImages(refs, mediaMap);
```

## Model

```ts
export interface CellImage {
  /** DISPIMG id — value of the `name` attribute on `xdr:cNvPr`. */
  id: string;
  /** Image binary, extracted from the package media folder. */
  data: Uint8Array;
  /** Image format inferred from the media file extension. */
  type: "png" | "jpeg" | "gif" | "svg" | "webp";
  /** Optional human-readable description (`descr` attribute). */
  description?: string;
}
```

`CellImage` is workbook-wide rather than sheet-scoped — that mirrors
the file format, where `xl/cellimages.xml` is a single registry that
every sheet's DISPIMG formulas resolve against.

## What changed in the writer side

| Hook | Change |
| --- | --- |
| `ContentTypesOptions` | new `hasCellImages` flag — emits Override for `/xl/cellimages.xml` |
| `writeWorkbookRels` | new trailing `hasCellImages` parameter — emits `Type=".../cellimage"` with a unique rId past sheets / styles / theme / macros / FeaturePropertyBag / persons / externalLinks |
| `roundtrip` | detects the part in `_rawEntries`, threads the flag through, and force-preserves the media files the sibling rels actually reference (otherwise the `xl/media/image*` sweep would drop them) |

## Test plan

- [x] 14 new tests; `pnpm test` (lint + typecheck + vitest) green; `pnpm build` clean
- [x] `parseCellImages`: required attrs, optional `descr` omitted (not empty string), skips entries missing `<xdr:pic>`, missing DISPIMG name, or missing embed rId
- [x] `assembleCellImages`: drops refs whose embed rId has no media entry; dedupes by id (first occurrence wins); preserves description
- [x] Reader integration: full XLSX with two cellImages (PNG + JPEG, one with `descr`, one without) read into `workbook.cellImages` with intact bytes and correct types; absent part leaves `workbook.cellImages` undefined
- [x] End-to-end roundtrip: `openXlsx → saveXlsx` preserves `xl/cellimages.xml`, `xl/_rels/cellimages.xml.rels`, and the media files; a re-read recovers the same model byte-for-byte
- [x] Output declares the `cellimage` rel in `workbook.xml.rels`, the Override in `[Content_Types].xml`
- [x] No-cellimages workbooks emit clean rels — no spurious cellimage entries

## Out of scope

Synthesizing a `cellimages.xml` from a model on a fresh `writeXlsx`
call (without an existing source file) needs an XML serializer for
the body and the rels; this PR only handles the read + preserve side.
The infrastructure for the references is now in place — adding the
body writer is a natural follow-up.

References:
- WPS Office `etCustomData` namespace — `http://www.wps.cn/officeDocument/2017/etCustomData`
- WPS cellimage relationship type — `http://www.wps.cn/officeDocument/2017/relationships/cellimage`

Closes #179

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,6 +343,34 @@ used by formulas like `[1]Sheet1!A1`. Cached `t="s"` values stay as
 shared-string indices into the _external_ workbook (which hucre cannot
 dereference); resolved strings live in the linked file.
 
+### Cell-Embedded Images (WPS DISPIMG)
+
+WPS Office (and recent Excel versions) embed images inside cells via a
+workbook-level `xl/cellimages.xml` registry referenced from
+`=_xlfn.DISPIMG("<id>", 1)` formulas. hucre reads the registry into a
+typed `workbook.cellImages` array and re-declares the part on
+`saveXlsx` so the DISPIMG link survives round-trips — without this the
+relationship and content-type override are dropped and the formula
+loses its target.
+
+```ts
+import { readXlsx } from "hucre";
+
+const wb = await readXlsx(buf);
+for (const img of wb.cellImages ?? []) {
+  console.log(img.id, img.type, img.description, img.data.byteLength);
+}
+
+// Standalone parsers when you already have the XML strings.
+import { parseCellImages, assembleCellImages, REL_CELL_IMAGES } from "hucre";
+const refs = parseCellImages(cellImagesXml);
+const images = assembleCellImages(refs, mediaMap);
+```
+
+Synthesizing a `cellimages.xml` from a model on a fresh `writeXlsx`
+call (without an existing source file) is a follow-up — for now the
+read + roundtrip-preserve side is in place.
+
 ### Unified API
 
 Auto-detect format and work with simple helpers:
@@ -808,6 +836,8 @@ Zero dependencies. Pure TypeScript. The ZIP engine uses `CompressionStream`/`Dec
 | `XlsxStreamWriter`                 | Incremental row-by-row XLSX writing; auto-splits past `maxRowsPerSheet` |
 | `XLSX_MAX_ROWS_PER_SHEET`          | Excel hard row limit (1,048,576) — exported constant                    |
 | `parseExternalLink(xml, relsXml?)` | Parse `xl/externalLinks/externalLinkN.xml` → `ExternalLink`             |
+| `parseCellImages(xml)`             | Parse `xl/cellimages.xml` → `ParsedCellImageRef[]` (WPS DISPIMG)        |
+| `assembleCellImages(refs, media)`  | Combine parsed refs with resolved media bytes → `CellImage[]`           |
 
 ### ODS
 
@@ -951,6 +981,7 @@ Contributions are welcome! Please [open an issue](https://github.com/productdevb
 - Checkboxes (Excel 2024+)
 - VBA/macro injection
 - Slicers & timeline filters
+- WPS DISPIMG cell-embedded images — synthesize from a fresh write (read + roundtrip already supported)
 - R1C1 notation support
 - Accessibility helpers (WCAG 2.1 AA)
 
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -780,6 +780,30 @@ export interface ExternalLink {
   definedNames?: ExternalDefinedName[];
 }
 
+// ── Cell-Embedded Images (WPS DISPIMG / cellimages) ───────────────
+
+/**
+ * An image embedded inside a cell via the WPS Office cellimages mechanism
+ * (also recognized by recent Excel versions). The image is referenced from
+ * a cell formula `=_xlfn.DISPIMG("<id>", 1)` and the binary lives in the
+ * package as a regular media part. Unlike `SheetImage` (which is anchored
+ * to a drawing rectangle on a sheet), a `CellImage` is workbook-wide and
+ * can be referenced from any number of cells.
+ */
+export interface CellImage {
+  /**
+   * Stable image identifier as it appears inside the DISPIMG formula
+   * (`name` attribute on `xdr:cNvPr`). For example `"ID_2A8C..."`.
+   */
+  id: string;
+  /** Image binary, extracted from the package media folder. */
+  data: Uint8Array;
+  /** Image format inferred from the media file extension. */
+  type: SheetImage["type"];
+  /** Optional human-readable description (`descr` attribute). */
+  description?: string;
+}
+
 // ── Workbook ───────────────────────────────────────────────────────
 
 export interface Workbook {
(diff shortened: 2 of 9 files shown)


productdevbook (commented):
Review summary:

Verified locally: typecheck and `pnpm build` pass; the full vitest suite (469 files / 13549 tests, including 14 new for this PR) passes. The lint warnings shown locally come from sibling agent worktrees inside `.claude/worktrees/`, not from this PR.

Things I checked:
- `parseCellImageEntry` correctly drops entries that are missing `<xdr:pic>`, missing `cNvPr/@name`, or missing `blip/@r:embed`; it also accepts an unprefixed `embed` attribute as a fallback. Empty `descr` is left as `undefined`, not an empty string.
- `assembleCellImages` first-occurrence dedupe by `id` is the right call for malformed registries.
- In `roundtrip.ts`, `collectCellImageMediaPaths` walks the original `xl/_rels/cellimages.xml.rels` and force-preserves each `Target` against the `xl/media/imageN.*` regeneration sweep — without this the media bytes would be lost because the per-sheet drawing pass owns that prefix.
- `writeWorkbookRels` now bumps `nextRid` past the externalLinks block before emitting `cellimage`, so the new rId stays unique. The unit test asserts `cellImagesId > every other id` on the most realistic ordering.
- `readXlsx` matches the WPS namespace exactly (`r.type === REL_CELL_IMAGES`) instead of the loose `matchesRelType` helper — appropriate because that namespace is fixed and the loose helper would over-match generic suffixes.
- README updated in this same branch (commit pushed): added a "Cell-Embedded Images (WPS DISPIMG)" section mirroring the External Workbook References style, two new rows in the XLSX API table, and a roadmap line.

Acknowledged limitation (already in the PR description): synthesizing `xl/cellimages.xml` from a fresh `writeXlsx` (without an existing source file) needs an XML body writer; this PR is read + preserve only, which matches the established pattern.

LGTM — will squash-merge once CI is green on the README commit.

productdevbook (commented):
Follow-up review (after merging main into the branch):

Re-verified locally on the merged branch: typecheck and `pnpm build` are clean; the full vitest suite (81 files / 2302 tests, including 14 for this PR) passes.

What changed in the merge:
- Main now carries #183 (slicers / timelines) and #185 (pivot tables), both of which extended the same workbook-rels and content-types writer surface this PR touches. Resolved by keeping all four relationship families (cellimages, pivot, slicer, timeline) and stitching them through the shared `nextWorkbookRelId` allocator in `roundtrip.ts`.
- Pivot / slicer / timeline relationships are pre-assigned by the caller via `nextWorkbookRelId++`, while cellimages keeps its internal `nextRid`-based emission inside `writeWorkbookRels`. To stop those two counters from colliding, the caller now bumps `nextWorkbookRelId` by 1 when `hasCellImages` is true, right after `externalLinkRels` and before `pivotCacheRels`. That mirrors the order children are pushed inside `writeWorkbookRels` and keeps every rId unique.
- `writeWorkbookRels` signature gained `pivotCacheRels`, `slicerCacheRels`, `timelineCacheRels` (from #185 / #183), with `hasCellImages` moved to the trailing position. Updated `test/cell-images.test.ts` to pass the new optional positionals as `undefined` so the call still selects the cellimage path.
- README sections for cell-images, slicers, and pivot tables now coexist; the API table picked up all three sets of standalone parsers; the roadmap line was kept on a single line under the merged form.

Already covered in the original review (still valid):
- `parseCellImageEntry` drops malformed entries (no `<xdr:pic>`, no `cNvPr/@name`, no `blip/@r:embed`); accepts unprefixed `embed` as a fallback; leaves empty `descr` as `undefined`.
- `assembleCellImages` first-occurrence dedupe by id is the right behavior for malformed registries.
- `collectCellImageMediaPaths` walks the original `xl/_rels/cellimages.xml.rels` and force-preserves each `Target` against the `xl/media/imageN.*` regeneration sweep — without this the media bytes would be lost because the per-sheet drawing pass owns that prefix.
- `readXlsx` matches the WPS namespace exactly (`r.type === REL_CELL_IMAGES`) instead of the loose `matchesRelType` helper — appropriate because that namespace is fixed and the loose helper would over-match generic suffixes.

Acknowledged caveat (already in the PR description): synthesizing a fresh `cellimages.xml` from a model on `writeXlsx` (without an existing source file) needs an XML body writer; this PR is read + preserve only.

LGTM with the merge fixes — will squash-merge once CI is green.

PR #188, opened 2026-05-02
feat(xlsx): native pivot table creation — Phase 1

### LINKED ISSUE #159: feat: pivot table creation
## Summary

Add programmatic pivot table creation. The most complex XLSX feature — even openpyxl only supports read-only.

## Technical details

Three interconnected XML parts:

1. **pivotTableDefinition** (`xl/pivotTables/`) — field axes, data fields, layout
2. **pivotCacheDefinition** (`xl/pivotCache/`) — source range, field definitions, shared items
3. **pivotCacheRecords** — cached source data values

Key elements: `<pivotFields>` (row/col/page/value axes), `<dataFields>` (aggregation: sum, count, average, max, min, product, stdDev, var), `<rowFields>`, `<colFields>`, `<pageFields>`.

## API

```typescript
const xlsx = await writeXlsx({
  sheets: [{
    name: 'Data',
    rows: [['Region', 'Product', 'Revenue'], ...data],
    pivotTables: [{
      name: 'SalesPivot',
      sourceRange: 'A1:C100',
      targetSheet: 'Pivot',
      targetCell: 'A1',
      rows: ['Region'],
      columns: ['Product'],
      values: [{ field: 'Revenue', function: 'sum' }],
    }],
  }],
})
```

## Scope

- [ ] pivotCacheDefinition generation (source range, field defs, shared items)
- [ ] pivotCacheRecords generation
- [ ] pivotTableDefinition generation (field axes, data fields)
- [ ] Basic pivot table styles
- [ ] Round-trip preservation (already works)

## References

- ECMA-376 Part 1: PivotTable definitions
- MS-XLSX: https://learn.microsoft.com/en-us/office/open-xml/spreadsheet/working-with-pivottables

productdevbook:
## Summary

Adds first-class **pivot table creation** to `writeXlsx` through the new
`WriteSheet.pivotTables` field. Phase 1 of #159 covers the most common
dashboard shape — a tabular source summarised onto another sheet with
row / column / page / value field axes. **No other open-source JS
library can author pivot tables reliably** — ExcelJS skips them entirely
and even openpyxl on Python only supports read-only pivots.

## API

```ts
import { writeXlsx } from "hucre";

const xlsx = await writeXlsx({
  sheets: [
    {
      name: "Data",
      rows: [
        ["Region", "Product", "Revenue"],
        ["EU", "A", 100],
        ["EU", "B", 50],
        ["US", "A", 200],
        ["US", "B", 75],
      ],
    },
    {
      name: "Pivot",
      pivotTables: [
        {
          name: "SalesPivot",
          sourceSheet: "Data",
          rows: ["Region"],
          columns: ["Product"],
          values: [{ field: "Revenue", function: "sum" }],
        },
      ],
    },
  ],
});
```

## Model

```ts
interface WritePivotDataField {
  field: string;                       // source column name
  function?: PivotDataFieldFunction;   // sum (default), count, average,
                                       // max, min, product, countNums,
                                       // stdDev, stdDevp, var, varp
  displayName?: string;                // default: "Sum of <field>"
  numberFormat?: string;
}

interface WritePivotTable {
  name: string;
  sourceSheet?: string;                // defaults to host sheet
  sourceRange?: string;                // auto-detected when omitted
  targetCell?: string;                 // default: "A1"
  rows?: string[];                     // row-axis fields
  columns?: string[];                  // column-axis fields
  pages?: string[];                    // page (filter) fields
  values: WritePivotDataField[];       // data-axis fields
  styleName?: string;                  // default: "PivotStyleLight16"
  dataCaption?: string;                // default: "Values"
}
```

## Implementation

- **`src/xlsx/pivot-writer.ts`** — new module that serializes the five
  OOXML parts that back a pivot table:
  - `xl/pivotCache/pivotCacheDefinition{N}.xml` — field declarations
    plus shared-items tables for non-numeric axis fields and
    `containsNumber` / `minValue` / `maxValue` markers for numeric data
    fields.
  - `xl/pivotCache/_rels/pivotCacheDefinition{N}.xml.rels` — links the
    cache definition to its records part.
  - `xl/pivotCache/pivotCacheRecords{N}.xml` — emits one `<r>` per
    source row with `<x v="N"/>` shared-item refs for string fields,
    `<n v="..."/>` literals for numeric fields, and `<m/>` for blanks.
  - `xl/pivotTables/pivotTable{N}.xml` — pivot layout with
    `<pivotFields>`, `<rowFields>` / `<colFields>` / `<pageFields>` /
    `<dataFields>`, and a `pivotTableStyleInfo` block.
  - `xl/pivotTables/_rels/pivotTable{N}.xml.rels` — links the pivot to
    its cache definition.
- **`writer.ts`** orchestrates per-sheet pivot numbering, ZIPs the
  pivot bodies, and registers Override entries in
  `[Content_Types].xml` plus `<pivotCaches>` in `xl/workbook.xml` and
  the matching cache rel in `xl/_rels/workbook.xml.rels`. The existing
  `PivotCacheRef` / `PivotCacheRel` infrastructure (used by roundtrip)
  is reused as-is.
- **`worksheet-writer.ts`** allocates rIds for pivot-table relationships
  on the host sheet's rels file and surfaces them through
  `WorksheetResult.pivotTables`. Pivot tables are referenced only via
  the rels file — there is no `<pivotTable>` element in the worksheet
  body itself.
- Source data is read from either `sheet.rows` (raw 2-D arrays) or
  `sheet.data` + `sheet.columns` (object-style); both shapes normalise
  into the same `CellValue[][]` before resolution.
- Field types are inferred from the source data: a column is `numeric`
  when every non-empty cell is a finite number, otherwise `string` and
  values are routed through the shared-items table. This mirrors what
  Excel does on `Refresh`.
- The numeric layout (row totals, grand totals, value cells) is left
  for Excel to compute on first open via the existing
  `<calcPr fullCalcOnLoad="1"/>` recompute — Phase 1 ships the
  structural skeleton, not pre-computed value cells.

## Out of scope (future phases)

- Pre-computed `<rowItems>` / `<colItems>` expansions for fully
  rendered layouts on first open without a refresh.
- Multiple pivot tables sharing a single cache (each pivot owns one
  cache today).
- Calculated fields, calculated items, grouping (by date, by range),
  and `showDataAs` value transforms (% of total, running total, ...).
- Slicer / timeline integration with newly authored pivots.

## Tests

30 new cases in `test/pivot-write.test.ts`:

- `resolvePivotSource` — header extraction, auto-derived ranges,
  explicit range honour, missing-field error, padding / trimming of
  ragged rows.
- `writePivotTable` — cache definition shape, shared-items collection
  for string fields, numeric `<sharedItems>` markers, record
  serialisation (string-as-index, numeric, blank, integer numerics),
  axis placement, auto-labelled data fields ("Sum of Revenue"),
  displayName overrides, targetCell location, rels files, default and
  override style names.
- End-to-end through `writeXlsx` — five pivot parts in the ZIP,
  Content Types overrides, `<pivotCaches>` in workbook.xml, cache rel
  in workbook.xml.rels, host-sheet rel, no-op when no pivots,
  self-sourcing, multi-pivot index numbering, object-style source
  data, missing-source error, every supported aggregation function,
  re-read recovers the model.

`pnpm test` (lint + typecheck + vitest) and `pnpm build` both pass.

Closes #159

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ import { readXml, writeXml } from "hucre/xml"; // Tabular XML
 | **Write XLSX**        | Yes          | Yes           | Yes             | Yes             | Yes               |
 | **Streaming**         | Read+Write   | Write-only    | No              | const_memory    | SXSSF (write)     |
 | **Charts**            | Round-trip   | 15+ types     | 9 types         | 12+ types       | Limited           |
-| **Pivot tables**      | No           | Read-only     | No              | No              | Limited           |
+| **Pivot tables**      | Read + Write | Read-only     | No              | No              | Limited           |
 | **Cond. formatting**  | Yes (all)    | Yes           | Yes             | Yes             | Yes               |
 | **Sparklines**        | Yes          | No            | Yes             | Yes             | No                |
 | **Formula eval**      | No           | No            | No              | No              | Yes               |
@@ -443,8 +443,52 @@ import { parsePivotTable, parsePivotCacheDefinition, attachPivotCacheFields } fr
 
 `PivotTable.cacheId` matches the workbook-level `cacheId` rather than a
 per-table relationship, so reordering `Workbook.pivotCaches` keeps the
-links sound. Synthesizing a pivot table or cache definition from a
-fresh `writeXlsx` is a follow-up — this is read + preserve only.
+links sound.
+
+`writeXlsx` can also author pivot tables from scratch via the per-sheet
+`pivotTables` field. Hucre emits the pivot cache (definition + cached
+records), the pivot layout, and every required relationship and content
+type. The numeric layout (row totals, grand totals, value cells) is left
+for Excel to compute on first open via the existing `fullCalcOnLoad`
+recompute — Phase 1 ships the structural skeleton, not pre-computed
+value cells.
+
+```ts
+import { writeXlsx } from "hucre";
+
+const xlsx = await writeXlsx({
+  sheets: [
+    {
+      name: "Data",
+      rows: [
+        ["Region", "Product", "Revenue"],
+        ["EU", "A", 100],
+        ["EU", "B", 50],
+        ["US", "A", 200],
+        ["US", "B", 75],
+      ],
+    },
+    {
+      name: "Pivot",
+      pivotTables: [
+        {
+          name: "SalesPivot",
+          sourceSheet: "Data",
+          rows: ["Region"],
+          columns: ["Product"],
+          values: [{ field: "Revenue", function: "sum" }],
+        },
+      ],
+    },
+  ],
+});
+```
+
+Supported aggregation functions: `sum` (default), `count`, `average`,
+`max`, `min`, `product`, `countNums`, `stdDev`, `stdDevp`, `var`,
+`varp`. Pivots can source from their own sheet (omit `sourceSheet`)
+or any sibling sheet, and accept either `rows` (raw 2-D arrays) or
+`columns` + `data` (object-style) source shapes.
 
 ### Charts
 
@@ -1095,7 +1139,6 @@ Contributions are welcome! Please [open an issue](https://github.com/productdevb
 - XLSB binary format read
 - Formula evaluation engine
 - File encryption/decryption (AES-256, MS-OFFCRYPTO)
-- Pivot table creation — synthesize from a fresh write (read + roundtrip already supported)
 - Threaded comments (Excel 365+) — synthesize from a fresh write (read + roundtrip already supported)
 - Checkboxes (Excel 2024+)
 - VBA/macro injection
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -944,6 +944,77 @@ export interface PivotCache {
   hasRecords?: boolean;
 }
 
+/**
+ * A data field placement on a {@link WritePivotTable}.
+ *
+ * `field` names a column in the source data; `function` selects the
+ * aggregation Excel applies (`sum` is the default). `displayName` becomes
+ * the column header on the rendered pivot — it defaults to
+ * `"<Function> of <field>"`, mirroring Excel's auto-label.
+ */
+export interface WritePivotDataField {
+  /** Source column name (must match an entry in the source header row). */
+  field: string;
+  /** Aggregation function. Default: `"sum"`. */
+  function?: PivotDataFieldFunction;
+  /** Optional display name override. Default: e.g. `"Sum of Revenue"`. */
+  displayName?: string;
+  /** Optional number format for aggregated values. Default: General. */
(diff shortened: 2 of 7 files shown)


productdevbook (commented):
Reviewed Phase 1 of the pivot writer. The OOXML wiring (cache definition, cache records, pivot table layout, sheet rels, workbook `<pivotCaches>`, content-type overrides) is structured cleanly and the test coverage on the round-trip path is solid.

**Bug found and fixed** — `buildPivotTableRels` and `buildCacheDefinitionRels` hard-coded `pivotCacheRecords1.xml` and `../pivotCache/pivotCacheDefinition1.xml` regardless of the global pivot index. With two or more pivot tables in a workbook, every pivot beyond the first resolved to the wrong cache definition / records part — even though the parts themselves were named correctly (`pivotCacheRecords2.xml` etc.) — so Excel would silently load the first cache twice on refresh. The existing multi-pivot test only asserted file existence, not rels content, which is why this slipped through.

**Fix** (commit `54ed1f0`):
- Added a `partIndex` parameter to `writePivotTable` that the orchestrator in `writer.ts` populates from the existing `globalIndex` counter.
- Both rels builders now interpolate `partIndex` into the `Target` attribute.
- Extended the "multiple pivot tables with sequentially numbered indices" test to assert `pivotTable2.xml.rels` points at `pivotCacheDefinition2.xml` and `pivotCacheDefinition2.xml.rels` points at `pivotCacheRecords2.xml`.
- Default value is `1` so the public signature stays backward-compatible.

**Docs** — also updated:
- `README.md` Pivot Tables section now describes the `WriteSheet.pivotTables` authoring API with a runnable example, plus the supported aggregation function list.
- Cross-language comparison row updated from "No" to "Read + Write".
- "Pivot table creation" removed from the roadmap.

`pnpm test` (lint + typecheck + 2,345 vitest cases) and `pnpm build` are green locally; CI is re-running on the new commit.

**Other observations** (not blocking, worth noting for Phase 2):
- `buildFieldMetadata` adds an empty `""` to `sharedStrings` for blank cells, but `buildRecordCell` emits `<m/>` for those same cells rather than `<x v="N"/>`. Harmless — the unused entry just sits in the table — but the `count` on `<sharedItems>` ends up one higher than the real referenced item count.
- `numberFormat` on a `WritePivotDataField` is currently dropped (the `numFmtId` is forced to `0` with a comment about it being style-driven). Worth wiring through the styles collector in a follow-up so user-supplied formats survive.
- `cacheId` is derived from `globalIndex - 1` (0-based) while the rest of the code uses 1-based pivot indices. The numbering is correct but the offset is easy to misread; a small helper would help.

Will merge once CI clears.

Issue #77, opened 2026-03-25
Streaming XLSX reader: true streaming without full ZIP in memory


productdevbook:
## Gap Analysis

hucre's `streamXlsxRows()` uses SAX parsing for the worksheet XML (good), but still requires the entire ZIP file in memory upfront because `ZipReader` needs the full buffer. For truly large files (500MB+), we need ZIP streaming.

## Competitors
- **ExcelJS**: Streaming workbook reader that processes the file as a stream
- **SheetJS**: Stream processing support

## Current State
`streamXlsxRows()` does:
1. Load entire ZIP into memory (bottleneck)
2. Extract shared strings + styles (small, fine to keep in memory)
3. SAX-parse worksheet XML row by row (good, low memory per row)

The bottleneck is step 1: the ZIP must be fully loaded.

## Proposed Improvements
1. Support `ReadableStream<Uint8Array>` as input (already in `ReadInput` type but not implemented)
2. Stream ZIP entries without buffering the full archive
3. Process worksheet XML incrementally as ZIP chunks arrive

## Complexity
High — ZIP format stores the central directory at the end of the file, making true streaming difficult. Options:
- Buffer only the worksheet XML entry (skip non-needed entries)
- Use local file headers for streaming (less reliable but workable)
- Accept that shared strings must be fully parsed before worksheet streaming begins

## Considerations
- Most real-world XLSX files are <100MB; the ZIP overhead is acceptable
- The main memory win comes from SAX parsing (already done)
- True ZIP streaming is complex and may not be worth the complexity for v1.0

PR #189, opened 2026-05-02
feat: accept ReadableStream input on every reader entry point

productdevbook:
## Summary

`ReadInput` is already typed as `Uint8Array | ArrayBuffer | ReadableStream<Uint8Array>`, but only `streamXlsxRows` actually honored the stream arm. `readXlsx`, `readOds`, and the unified `read()` / `readObjects()` dispatcher all threw `"ReadableStream input is not supported by the unified read() API. Use readXlsx/readOds directly."` — and `readXlsx` itself just threw `"Unsupported input type. Expected Uint8Array or ArrayBuffer."`. The shorthand helpers (`readXlsxObjects`, `readOdsObjects`) inherited the gap.

This PR fills the gap so every `ReadInput`-typed entry point accepts a stream:

- New `src/_input.ts` exports `bufferReadableStream` and `readInputToUint8Array`. The `ReadableStream` check is duck-typed (`getReader`), so streams from other realms — Node worker threads, undici, browser iframes — are accepted.
- `readXlsx`, `readOds`, and `read()` now `await readInputToUint8Array(input)` instead of the per-file `toUint8Array` helpers.
- `streamXlsxRows` reuses `bufferReadableStream` (its private `bufferStream` was removed) so there is one canonical implementation.
- `readXlsxObjects`, `readOdsObjects`, `readObjects` inherit support automatically since they delegate to the format readers.

Streams are fully buffered before parsing because XLSX and ODS both store the ZIP central directory at the end of the archive — true streaming requires random access. `streamXlsxRows` remains the path for row-level streaming with low per-row memory.

Refs #77 (item 1 of the proposed-improvements list — surface-level API gap. The "true streaming without full ZIP in memory" main goal is intentionally out of scope; that requires a streaming ZIP reader and is called out as high-complexity in the issue.)

## Test plan

- [x] New `test/readable-stream-input.test.ts` covers single-chunk, multi-chunk (64-byte and 256-byte slices), `ArrayBuffer` regression, and unsupported-input error paths across `readXlsx`, `readOds`, `read()`, `readObjects`, `readXlsxObjects`, `readOdsObjects` — 11 cases, all passing.
- [x] `pnpm test` (lint + typecheck + 2371 vitest cases) green.
- [x] `pnpm build` green.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1025,7 +1025,7 @@ Zero dependencies. Pure TypeScript. The ZIP engine uses `CompressionStream`/`Dec
 
 | Function                           | Description                                                                 |
 | ---------------------------------- | --------------------------------------------------------------------------- |
-| `readXlsx(input, options?)`        | Parse XLSX from `Uint8Array \| ArrayBuffer`                                 |
+| `readXlsx(input, options?)`        | Parse XLSX from `Uint8Array \| ArrayBuffer \| ReadableStream<Uint8Array>`   |
 | `writeXlsx(options)`               | Generate XLSX, returns `Uint8Array`                                         |
 | `readXlsxObjects(input, options?)` | Read sheet as `{ data, headers }` — mirror of CSV                           |
 | `writeXlsxObjects(data, options?)` | Write objects to XLSX (auto-derives headers from keys)                      |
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
## Review

Solid, focused fix for a real API gap — `ReadInput` was already typed as `Uint8Array | ArrayBuffer | ReadableStream<Uint8Array>` but only `streamXlsxRows` honored the third arm. Centralizing the normalization in `src/_input.ts` and routing every entry point (`readXlsx`, `readOds`, `read()`, plus the object/streaming shorthands) through `readInputToUint8Array` removes the inconsistency and the duplicated `toUint8Array` helpers. Nice cleanup of `streamXlsxRows`'s private `bufferStream` too — one canonical implementation.

### Strengths

- Duck-typed `isReadableStream` (looks for `getReader`) is the correct call here — `instanceof ReadableStream` would silently fail across realms (Node worker threads, undici, browser iframes).
- `bufferReadableStream` correctly handles the 0-chunk edge case (returns empty `Uint8Array`) and the 1-chunk fast path, and adds a defensive null check on `value` that the previous private helper lacked.
- Error type is `ParseError` with a clear message listing all three accepted shapes; consistent with project conventions.
- Test coverage is thorough: single-chunk, multi-chunk (64-byte and 256-byte slices to exercise chunk boundaries), `ArrayBuffer` regression, unsupported-input error path, all three readers, the unified dispatcher, and both `*Objects` shorthands. 11 cases, all passing.
- Doc comments on each entry point explicitly call out *why* full buffering is required (ZIP central directory at end of archive) and direct users to `streamXlsxRows` for the row-level streaming path that does avoid the full-workbook buffer. Good steering.

### Notes

- Scope is correctly limited to surface-level API gap (item 1 of #77); true streaming without full ZIP in memory is out of scope and that's the right call — it would require a streaming ZIP reader and is a much larger change.
- The PR docstring mentions XLSB but the codebase has no XLSB reader yet, so that's just future-looking comment text — harmless.

### Fixes pushed on top

- Merged latest `main` (which had #188 native pivot tables) into the branch — clean, no conflicts.
- Updated the README API tables for `readXlsx` and `readOds` to advertise `ReadableStream<Uint8Array>` as accepted input. The previous wording said only `Uint8Array | ArrayBuffer`, which would have continued to mislead users.
- Ran `pnpm fmt` to clean up table formatting.

### Verification

- `pnpm test` (lint + typecheck + 2401 vitest cases) green on the merged branch.
- `pnpm build` green.

LGTM. Will merge once CI passes on the new commits.

Issue #136, opened 2026-03-30
Feature request: Chart cloning and composition from template workbooks


MathiasWP:
## Use case

We're building a sustainability compliance platform where users create dashboards with various chart types (line charts, bar charts, donut/pie charts, etc.). When exporting a dashboard to Excel, we want to:

1. Maintain a **template workbook** containing one instance of each supported chart type (line, bar, donut, area, scatter, stacked bar)
2. At export time, **clone specific charts** from the template into a new workbook
3. **Create multiple instances** of the same chart type with different data ranges, axis labels, colors, and titles
4. Position them on a sheet as a composed dashboard

### Example

A user's dashboard has 3 line charts and 1 donut chart. The export should:
- Pick the "line chart" and "donut chart" from the template workbook
- Create 3 instances of the line chart, each pointing at different data ranges with custom axis labels and colors
- Create 1 instance of the donut chart with custom section labels, colors, and sizes
- Place all 4 charts on a sheet in the exported `.xlsx`

## Current state

Hucre's roundtrip API (`openXlsx` → modify → `saveXlsx`) preserves existing charts in a workbook, and `fillTemplate()` can populate data that charts reference. However, there's no way to:

- Read/parse chart definitions from `xl/charts/chartN.xml` into a data model
- Clone a chart and remap its data series to new cell ranges
- Modify chart properties (colors, labels, axis titles) programmatically  
- Insert new chart instances with proper drawing relationships

`copySheetToWorkbook` copies cell data and styles but not chart XML parts or drawing relationships.

### 1. Chart data model

A TypeScript interface representing the common chart properties:

```typescript
interface ChartDefinition {
  type: 'bar' | 'line' | 'pie' | 'doughnut' | 'area' | 'scatter' | 'stackedBar';
  title?: string;
  series: ChartSeries[];
  axes?: { x?: AxisConfig; y?: AxisConfig };
  legend?: LegendConfig;
  position: ChartPosition; // anchor cell, width, height
}

interface ChartSeries {
  name: string;
  dataRange: string;        // e.g. "Sheet1!$B$2:$B$10"
  categoryRange?: string;   // e.g. "Sheet1!$A$2:$A$10"
  color?: string;           // hex RGB
}
```

### 2. Core operations needed

```typescript
// Read charts from an existing workbook
function getCharts(workbook: Workbook): ChartDefinition[];

// Clone a chart with new configuration
[...]

PR #190, opened 2026-05-02
feat(xlsx): surface chart series metadata in parseChart

productdevbook:
## Summary

`parseChart` (and therefore `sheet.charts[i]`) previously surfaced only `kinds`, `seriesCount`, and `title`. The series themselves were unreachable, so users could see *that* a workbook had a chart but not *what* the chart was wired to. That's the gap that blocks the cloning workflow described in #136 (item 1 of the proposed approach: read chart definitions into a data model).

This PR fills that gap on the read side. `Chart.series` is now an array of `ChartSeriesInfo` records that mirror what `ChartSeries` accepts on the write side, so a parsed series can be used as the basis for a new `SheetChart` without remapping field names.

## API

```ts
import { readXlsx, type ChartSeriesInfo } from "hucre";

for (const chart of wb.sheets[0].charts ?? []) {
  for (const s of chart.series ?? []) {
    console.log(s.kind, s.index, s.name, s.valuesRef, s.categoriesRef, s.color);
    // e.g.  "bar"  0  "Revenue"  "Sheet1!$B$2:$B$10"  "Sheet1!$A$2:$A$10"  "1F77B4"
  }
}
```

## Model

```ts
interface ChartSeriesInfo {
  kind: ChartKind;            // matches Chart.kinds
  index: number;              // 0-based position inside its chart-type element
  name?: string;
  valuesRef?: string;         // raw <c:f> formula
  categoriesRef?: string;     // raw <c:f> formula
  color?: string;             // 6-digit RGB hex, upper-cased
}

interface Chart {
  kinds: ChartKind[];
  seriesCount: number;
  title?: string;
  series?: ChartSeriesInfo[]; // NEW — undefined when the chart has zero series
}
```

## Implementation

The reader walks every `<c:ser>` under each chart-type element in `<c:plotArea>` and records:

- **Kind & local index** — combo charts get independent per-kind indexing so the indices match the order Excel writes back when remapping.
- **Name** — literal `<c:v>` first, then `<c:strCache>` cached value, then the bare `<c:f>` formula text as a last resort (matches what other tools display when the cache is missing).
- **Values & categories** — `<c:val>` / `<c:cat>` for category-axis charts, `<c:yVal>` / `<c:xVal>` for scatter and bubble. `<c:numLit>` series (literal embedded data, no formula) intentionally surface no ref.
- **Color** — first `<a:srgbClr val>` under `<c:spPr><a:solidFill>`, normalized to upper-case 6-digit hex; malformed values are dropped instead of leaking through.

Refs #136 — covers item 1 of the issue's proposed approach (read chart definitions). `cloneChart` and `addChart` are intentionally out of scope here; the next step is straightforward now that `valuesRef` / `categoriesRef` are exposed (a user can already construct a `SheetChart` from a parsed chart).

## Test plan

- [x] Series name from literal `<c:v>`
- [x] Series name from `<c:strCache>` cached `<c:pt><c:v>`
- [x] Series name fallback to bare strRef formula text
- [x] Bar/column wiring via `<c:val>` and `<c:cat>`
- [x] Scatter wiring via `<c:xVal>` and `<c:yVal>`
- [x] `<c:numLit>` series omit `valuesRef` / `categoriesRef`
- [x] Color extraction from `<a:srgbClr>`, including `#`-prefix and case normalization
- [x] Malformed color values dropped (not surfaced as garbage)
- [x] Combo charts index series independently per chart-type element
- [x] `series` is `undefined` when a chart has zero `<c:ser>` children
- [x] Existing roundtrip tests still pass (kinds/title/seriesCount unchanged)
- [x] Full suite: `pnpm test` (lint + typecheck + 27,841 vitest tests)
- [x] `pnpm build`

### DIFF
--- a/README.md
+++ b/README.md
@@ -494,8 +494,9 @@ or any sibling sheet, and accept either `rows` (raw 2-D arrays) or
 
 Charts (`xl/charts/chartN.xml` plus the optional `styleN.xml` /
 `colorsN.xml` companions) are read into a per-sheet `sheet.charts`
-array surfacing the chart kind(s), series count, and plain-text
-title. On `saveXlsx` the chart parts are re-declared in
+array surfacing the chart kind(s), series count, plain-text title,
+and per-series metadata (name, value/category ranges, fill color).
+On `saveXlsx` the chart parts are re-declared in
 `[Content_Types].xml`, the chart-bearing drawing and its rels are
 force-preserved, and the regenerated worksheet body gets a
 `<drawing r:id="..."/>` re-anchor — without these wirings Excel
@@ -511,6 +512,11 @@ for (const sheet of wb.sheets) {
   for (const chart of sheet.charts ?? []) {
     console.log(chart.kinds, chart.seriesCount, chart.title);
     // e.g. ["bar"], 2, "Quarterly Sales"
+
+    for (const s of chart.series ?? []) {
+      console.log(s.kind, s.index, s.name, s.valuesRef, s.categoriesRef, s.color);
+      // e.g. "bar" 0 "Revenue" "Sheet1!$B$2:$B$10" "Sheet1!$A$2:$A$10" "1F77B4"
+    }
   }
 }
 
@@ -520,7 +526,12 @@ const chart = parseChart(xml);
 
 `Chart.kinds` lists every chart-type element present under
 `<c:plotArea>` in declaration order, so combo charts surface as e.g.
-`["bar", "line"]`. Sheets that hucre actively regenerates because they
+`["bar", "line"]`. `Chart.series` mirrors the field shape that
+`ChartSeries` accepts on the write side — a parsed series can be
+fed back into `WriteSheet.charts` to clone or re-bind a chart.
+Bubble/scatter `<c:numLit>` series (literal embedded data, no
+formula) intentionally surface no `valuesRef`/`categoriesRef`.
+Sheets that hucre actively regenerates because they
 also carry hucre-managed images currently keep the chart bodies but
 lose the in-drawing chart anchor — merging hucre's drawing output
 with the original chart graphicFrames is a follow-up.
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1221,11 +1221,38 @@ export type ChartKind =
   | "ofPie";
 
 /**
+ * A single series surfaced from a parsed chart.
+ *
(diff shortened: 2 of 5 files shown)


productdevbook (commented):
Reviewed end-to-end. Solid, focused PR — happy with the shape.

## What I checked

- **API symmetry with the writer.** `ChartSeriesInfo` field names line up cleanly with `ChartSeries` (`name`/`valuesRef`↔`values`/`categoriesRef`↔`categories`/`color`), so a parsed series can drive a fresh `SheetChart` without remapping. Per-kind `index` is exposed but the writer uses positional ordering, which is fine for the cloning use case.
- **XML traversal.** `formulaText` walks `<c:numRef>` → `<c:strRef>` → bare `<c:f>`, matching what the writer emits (`<c:val><c:numRef><c:f>`, `<c:cat><c:strRef><c:f>`, scatter `<c:xVal>/<c:yVal><c:numRef><c:f>`). `<c:numLit>` correctly drops out because there's no `<c:f>`.
- **Color normalization.** `replace(/^#/, "").toUpperCase()` then `^[0-9A-F]{6}$` rejects junk and matches the writer's hex shape — symmetric round-trip.
- **Combo charts.** `localIndex` resets per chart-type element (the comment mentions this), and the existing combo test still passes with the new `series` array. `kinds` continues to dedupe.
- **Tests.** 9 new cases cover literal/cache/formula name fallbacks, scatter wiring, numLit, malformed/`#`-prefixed colors, combo indexing, and the empty-series omission. `pnpm test` (lint + typecheck + 2399 vitest) and `pnpm build` both pass locally.

## Fixes pushed

- Added README documentation for `Chart.series` in the Charts section, with a usage example and a note about `<c:numLit>` series surfacing no `valuesRef` / `categoriesRef`. Commit: `docs(charts): document Chart.series introspection in README`.

## Minor observations (non-blocking)

- Bubble charts have a third axis (`<c:bubbleSize>`) that is not surfaced; mapping bubble's "values" to `<c:yVal>` is a sensible choice for now and consistent with the cloning roadmap. Worth a follow-up if anyone needs the bubble-size ref.
- When two `<c:barChart>` elements appear in one chart (rare), `kinds` dedupes to `["bar"]` while `index` resets per element — a consumer reconstructing series may want a stable (kind, occurrence) pair, but this matches the documented "per chart-type element" semantics.

LGTM — merging once CI is green.

Issue #156, opened 2026-04-03
feat: file encryption/decryption — password-protected XLSX


productdevbook:
## Summary

Read and write password-protected (encrypted) XLSX files. **No open-source JS library supports this.**

## Technical details

Encrypted XLSX = OLE2 container with two streams:
- `\EncryptionInfo` — algorithm, key size, salt, hash
- `\EncryptedPackage` — encrypted ZIP data

**Agile Encryption (Office 2010+):** AES-128/192/256 + CBC + SHA-256/512 + salt + spinCount.
**Standard Encryption (Office 2007):** AES block cipher, no CBC.

## API

```typescript
const wb = await readXlsx(buffer, { password: 'secret' })

const xlsx = await writeXlsx({
  sheets: [...],
  password: 'secret',
})
```

## Scope

- [ ] OLE2 container parsing (reuse from XLS BIFF8 issue)
- [ ] Agile Encryption decryption (AES-256 + SHA-512)
- [ ] Agile Encryption encryption
- [ ] Standard Encryption decryption
- [ ] Use Web Crypto API (edge-compatible, no Node.js crypto)

## References

- MS-OFFCRYPTO: https://learn.microsoft.com/en-us/openspecs/office_file_formats/ms-offcrypto/

PR #191, opened 2026-05-02
feat(xlsx,ods): detect password-protected workbooks and surface EncryptedFileError

productdevbook:
## Summary

Office encrypts password-protected XLSX / XLSM / ODS files inside an OLE2 / Compound File Binary (CFB) container — `\EncryptionInfo` and `\EncryptedPackage` streams per MS-OFFCRYPTO. The container has a fixed 8-byte magic header (`D0 CF 11 E0 A1 B1 1A E1`) that does not overlap with the ZIP / OOXML / ODF signatures, so a quick byte sniff is enough to tell an encrypted workbook apart from a plain one.

Until this PR, every reader fed those files straight into `ZipReader` and callers got a confusing `"Failed to open XLSX file: not a valid ZIP archive"` `ParseError` several layers down. The already-exported `EncryptedFileError` was never actually thrown by any code path.

## What this PR does

- **`src/_input.ts`** — adds two helpers, `isOle2Container(data)` (byte-sniff) and `assertNotEncrypted(data, format)` (throws `EncryptedFileError` with a format hint).
- **`src/errors.ts`** — `EncryptedFileError` now accepts an optional `format: "xlsx" | "ods"` hint plus an optional message override. The zero-arg constructor stays backward-compatible.
- **Reader entry points** — `readXlsx`, `streamXlsxRows`, and `readOds` all sniff the buffered input before opening the ZIP. The unified `read()` / `readObjects()` dispatcher catches encrypted containers before `detectFormat()` rejects them as "not a ZIP archive". The `readXlsxObjects` / `readOdsObjects` shorthands inherit detection through the readers they delegate to.

```ts
import { read, EncryptedFileError } from "hucre";

try {
  await read(buffer);
} catch (err) {
  if (err instanceof EncryptedFileError) {
    // err.format → "xlsx" | "ods" | undefined (undefined from read() since
    // we cannot tell which format the encrypted package wraps without
    // decrypting it first)
    return showPasswordPrompt();
  }
  throw err;
}
```

## What is *not* in this PR

Actual decryption of the OLE2 envelope (the `\EncryptionInfo` parsing, AES-CBC + SHA-256/512 PBKDF, package extraction). That is the bulk of #156 and intentionally out of scope here — this PR only closes the UX gap so callers no longer see a misleading ZIP error and can branch on a typed error today. Full encryption support is tracked in #156.

Refs #156

## Test plan

- [x] New `test/encrypted-file-detection.test.ts` — 21 cases covering:
  - `isOle2Container` byte sniff (magic, padding, ZIP, short buffer, partial match)
  - `readXlsx` (Uint8Array, ArrayBuffer, ReadableStream) — error type, `format`, message, base-class chain
  - `streamXlsxRows` (Uint8Array, ReadableStream)
  - `readOds` (Uint8Array, ReadableStream) — `format="ods"` hint
  - Unified `read()` and `readObjects()` — `format` undefined since the dispatcher cannot disambiguate
  - `readXlsxObjects` / `readOdsObjects`
  - `EncryptedFileError` constructor surface (legacy, format hint, custom message)
- [x] `pnpm exec vitest run` — 980 files / 27924 tests green.
- [x] `pnpm lint` (oxlint + oxfmt --check) — 0 errors.
- [x] `pnpm typecheck` — green.
- [x] `pnpm build` — green.

### DIFF
--- a/src/_input.ts
+++ b/src/_input.ts
@@ -12,7 +12,52 @@
 // ──────────────────────────────────────────────────────────────────────
 
 import type { ReadInput } from "./_types";
-import { ParseError } from "./errors";
+import { EncryptedFileError, ParseError } from "./errors";
+
+// ── OLE2 / Compound File Binary container detection ──────────────────
+//
+// Office password-protected XLSX / XLSM / ODS files are not ZIP archives —
+// they are OLE2 (a.k.a. Compound File Binary, CFB) containers carrying
+// `\EncryptionInfo` and `\EncryptedPackage` streams (MS-OFFCRYPTO
+// §2.3.4.x). They start with a fixed 8-byte magic header that has no
+// overlap with the ZIP / OOXML / ODF signatures, so a quick byte sniff
+// is enough to tell an encrypted container apart from a plain workbook.
+//
+// We don't actually decrypt the package here — full encryption support
+// is tracked in #156. Surfacing the situation early as a typed
+// `EncryptedFileError` saves callers from a confusing
+// `"not a valid ZIP archive"` ParseError several layers down.
+
+/** OLE2 / CFB magic header: `D0 CF 11 E0 A1 B1 1A E1`. */
+const OLE2_MAGIC = Object.freeze([0xd0, 0xcf, 0x11, 0xe0, 0xa1, 0xb1, 0x1a, 0xe1] as const);
+
+/**
+ * Whether `data` starts with the OLE2 / CFB compound-document magic
+ * bytes — the envelope Office uses for password-protected XLSX, XLSM,
+ * and ODS files. Returns `false` for shorter buffers and for any other
+ * leading bytes (plain ZIP archives, XML, etc.).
+ */
+export function isOle2Container(data: Uint8Array): boolean {
+  if (data.length < OLE2_MAGIC.length) return false;
+  for (let i = 0; i < OLE2_MAGIC.length; i++) {
+    if (data[i] !== OLE2_MAGIC[i]) return false;
+  }
+  return true;
+}
+
+/**
+ * Throw {@link EncryptedFileError} when `data` is an OLE2 / CFB
+ * compound-document container — the envelope Office uses for
+ * password-protected XLSX / ODS workbooks. No-op for plain ZIP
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
## Review

Solid scoped fix for a real UX gap. The OLE2 byte sniff is the right approach, and surfacing a typed `EncryptedFileError` ahead of the ZIP reader avoids the misleading "not a valid ZIP archive" `ParseError`. The format hint on the error is a nice touch for callers that need to branch on XLSX vs. ODS encryption paths once decryption (#156) lands.

**Strengths**
- `isOle2Container` is short, allocation-free, and correctly guards against short buffers and partial-magic matches.
- The `EncryptedFileError` constructor extension is properly backward-compatible (zero-arg path preserved, `format` only set when provided).
- Good coverage on the new test file: 21 cases across every public reader entry point, including ArrayBuffer / Uint8Array / ReadableStream variants and the legacy constructor surface.
- Comments are clear about scope: this is detection only, decryption stays under #156.

**Issue I fixed in this branch**
- `streamOdsRows` (`src/ods/stream.ts`) is also a public reader entry point but was missing the encryption sniff. An encrypted ODF passed to it would still bubble up the generic ZIP `ParseError`. I pushed `846924a` adding `assertNotEncrypted(data, "ods")` plus two matching tests in `test/encrypted-file-detection.test.ts` covering the generator and the `format="ods"` hint.

**Other notes (no action needed)**
- `openXlsx` (`src/xlsx/roundtrip.ts`) calls `readXlsx` first, so it inherits detection — confirmed no separate guard required.
- `isOle2Container` and `assertNotEncrypted` are intentionally not re-exported from `src/index.ts`, which keeps them internal and gives us room to evolve the detection signature alongside #156. Fine.
- The unified `read()` deliberately leaves `format` undefined on the error since the OLE2 envelope alone can't disambiguate XLSX vs. ODS — the test suite asserts this explicitly. Good call.

**Verification**
- `pnpm test` (lint + typecheck + vitest): 27937 / 27937 passing, 0 lint errors (existing warnings unrelated).
- `pnpm build`: green.

LGTM after the streamer fix — will merge once CI confirms.

PR #192, opened 2026-05-02
feat(xlsx): cloneChart — bridge parseChart output into SheetChart

productdevbook:
## Summary

- New `cloneChart(source, options)` converts a parsed `Chart` (read-side metadata produced by `parseChart`) into a `SheetChart` ready for `writeXlsx`. Together with `chartKindToWriteKind(kind)` it closes the missing template-composition step from issue #136: load a workbook with one of each chart flavor, pull out the chart, remap its series ranges, and re-emit it (often several times) into a new workbook.
- The bridge is type-narrowed: `Chart.series[].valuesRef` becomes `SheetChart.series[].values`, source colors / category ranges / series names carry through, and any of those fields can be replaced with a per-series override or dropped entirely with `null`. Read-side kinds collapse onto their write counterparts (bar/bar3D → column, doughnut/pie3D → pie, line3D → line, area3D → area, scatter passes through). Non-renderable kinds (bubble, radar, surface, stock, ofPie) require an explicit `options.type` override and otherwise throw with a clear error.
- The clone is then accepted unchanged by `writeChart` and end-to-end by `writeXlsx`, so callers can populate the dashboard sheet with data and immediately attach the cloned chart through the existing `WriteSheet.charts` array.

## What this PR does NOT do

This is a focused slice of #136 (chart composition / data remapping). The full issue also asks for chart drawing-position read support and high-level `getCharts(workbook)` / `addChart(sheet, ...)` shorthands; those remain out of scope here. `Chart.kinds` already drops the `bar` vs `column` orientation distinction at parse time, so cloned bar charts default to vertical (`column`) — callers who need horizontal pass `type: "bar"` explicitly.

## Test plan

- [x] `pnpm test` (lint + typecheck + 27 998 vitest cases, including 20 new `chart-clone` tests)
- [x] `pnpm build`
- [x] Round-trip integration: `parseChart` → `cloneChart` → `writeChart` and `writeXlsx` produce a valid `xl/charts/chart1.xml`
- [x] Three-way template reuse: a single source `Chart` cloned into three independent `SheetChart` instances with distinct titles, data ranges, and colors

Refs #136

### DIFF
--- a/README.md
+++ b/README.md
@@ -581,6 +581,40 @@ toggles `clustered` / `stacked` / `percentStacked`, `legend` accepts
 attributes for screen readers. Doughnut, radar, stock, 3D variants,
 trendlines, and combo charts are out of scope for Phase 1.
 
+#### Cloning a parsed chart with `cloneChart`
+
+`cloneChart(source, options)` bridges the read-side `Chart` produced by
+`parseChart` to the write-side `SheetChart` consumed by `writeXlsx`, so
+a template workbook can supply the visual styling for a new export and
+the caller only needs to swap the data ranges:
+
+```ts
+import { parseChart, cloneChart, writeXlsx } from "hucre";
+
+const template = parseChart(templateChartXml)!;
+
+const chart = cloneChart(template, {
+  anchor: { from: { row: 14, col: 0 }, to: { row: 28, col: 8 } },
+  title: "Q1 Revenue",
+  seriesOverrides: [{ values: "Dashboard!$B$2:$B$13", color: "1070CA" }],
+});
+
+await writeXlsx({
+  sheets: [{ name: "Dashboard", rows: dashboardRows, charts: [chart] }],
+});
+```
+
+`anchor` is required (the read side has no placement metadata).
+Per-series overrides accept `null` to drop an inherited value (e.g.
+`color: null` strips the template tint), can append new series past
+the source length, and fall back to the source's `valuesRef` /
+`categoriesRef` / `name` / `color` when omitted. Source kinds the
+writer can author collapse onto their write counterparts (`bar` /
+`bar3D` → `column`, `doughnut` / `pie3D` → `pie`, `line3D` → `line`,
+`area3D` → `area`); kinds with no analog (`bubble`, `radar`,
+`surface`, `stock`, `ofPie`) require an explicit `options.type`
+override.
+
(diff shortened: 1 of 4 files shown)


productdevbook (commented):
## Review

Reviewed the diff, types, integration with `chart-reader` / `chart-writer`, and ran `pnpm test` (lint + typecheck + 27 998 vitest cases) and `pnpm build` locally — all green. The bridge is well-scoped, the type-narrowing is explicit, and the test suite covers happy path, every override semantics (`null` drop, append-past-length, full replace), every read→write kind mapping, both error guards, and a true `parseChart → cloneChart → writeXlsx` round-trip with `xl/charts/chart1.xml` extraction. Field-by-field merge of `name` / `valuesRef` / `categoriesRef` / `color` lines up cleanly with what `chart-writer` consumes.

## Fixes pushed

- `docs(readme)` (8da6537): added a "Cloning a parsed chart with `cloneChart`" subsection right after the chart-author docs, plus two rows in the XLSX function reference table for `cloneChart` and `chartKindToWriteKind`. The original PR ships the helper but the README still only mentions `parseChart`, so users had no entry point for the new flow.

## Notes / non-blocking

1. **Combo charts collapse silently.** When `source.kinds.length > 1` (e.g. `["bar", "line"]`), `pickWritableKind` picks the first writable kind and `buildSeriesFromSource` flattens every series across every kind under that single `WriteChartKind`. The PR description correctly notes combo is out of scope, and the suggested `options.series` / `options.type` escape hatches work, but a debug-time `console.warn` (or doc nudge in the JSDoc) for "combo source detected, all N series rendered as <kind>" would prevent surprises. Up to you.
2. **`options.series` skips the values guard.** When the caller passes a fully-formed `series` array, the per-element `values` check from `mergeSeries` doesn't run; an empty `values` would slip through to `writeChart`. Low risk because TS already enforces `ChartSeries.values: string`, but a runtime check would match the existing throw-loud style.
3. **`anchor` is shallow-copied.** `out.anchor = options.anchor` keeps the caller's reference. Cloning the same chart into multiple positions inside a loop is the obvious use case; fine as long as callers pass a fresh anchor literal each time (the integration test does).

None of these block merge in my view. Happy to take a follow-up PR for any of them.

CI green, mergeable, README updated — ready to merge once the docs commit lands a passing run.

PR #193, opened 2026-05-02
feat(xlsx): streamXlsxRows honors range and maxRows

productdevbook:
## Summary

`streamXlsxRows` already accepted `sheet` and `dateSystem` from `ReadOptions`, but silently ignored two other widely-used fields: `range` and `maxRows`. The batch reader (`readXlsx`) honors both. This PR plumbs them through the streaming reader so callers can preview/sample/window-read large worksheets without buffering the whole sheet, and the parser actually stops doing work once the cap is hit.

## What changed

- `range` (A1-style ref like `"B2:D1000"`):
  - Rows outside the row span are skipped before they reach the consumer.
  - Cells outside the column span are masked to `null` so column indexes stay aligned (mirrors `parseWorksheet` semantics — the batch reader nulls cells, it doesn't shift them).
  - Parsing short-circuits as soon as the SAX layer emits a row past the end-row (worksheet rows are written in ascending order in valid OOXML).
- `maxRows`:
  - Caps the number of rows yielded.
  - Once the cap fires the underlying ZIP/decompression reader is cancelled, so very large sheets stay cheap when only the first N rows are needed.
- The generator's `finally` block also cancels the upstream reader when the consumer abandons the loop early (`for await … break`), so abandoned generators no longer leak the worksheet stream.

## Scope

This is a focused slice of #77 (memory-cheap streaming reads). The full issue also asks for true streaming without buffering the central directory, which the issue itself rates as high-complexity and beyond v1.0 scope. This PR does not change that constraint — it only lets streaming callers stop pulling and filter, which is the use case the issue sketches as "skip non-needed entries". `headerRow` is intentionally not wired up because in the rest of the codebase it's an `xlsx-objects` concern, not a base-reader concern.

Refs #77

## Test plan

- [x] `pnpm exec oxlint src/xlsx/stream-reader.ts test/streaming.test.ts` — 0 warnings, 0 errors
- [x] `pnpm typecheck`
- [x] `pnpm exec vitest run --exclude '**/.claude/**'` — 2443 / 2443 (86 files)
- [x] `pnpm build`
- [x] 9 new tests under `streamXlsxRows — ReadOptions filters` covering: maxRows cap, maxRows defaults (0/undefined/negative), oversized maxRows, range row+column masking, single-cell range, range early termination, range+maxRows composition, rows entirely outside range, invalid range error, early `break` doesn't throw.

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,6 +255,21 @@ for await (const row of streamXlsxRows(buffer)) {
   console.log(row.index, row.values);
 }
 
+// Cap the number of rows yielded (preview / sampling). The underlying
+// ZIP/SAX stream is cancelled once the cap is reached, so very large
+// sheets stay cheap.
+for await (const row of streamXlsxRows(buffer, { maxRows: 100 })) {
+  console.log(row.index, row.values);
+}
+
+// Filter to an A1 range. Rows outside the row span are skipped; cells
+// outside the column span are masked to `null` (column indexes stay
+// stable). Parsing stops once a row past the end-row is observed.
+for await (const row of streamXlsxRows(buffer, { range: "B2:D1000" })) {
+  // row.values[0] === null (column A is outside)
+  // row.values[1..3] carry B/C/D
+}
+
 // Stream write — add rows incrementally
 const writer = new XlsxStreamWriter({
   name: "BigData",
--- a/src/xlsx/stream-reader.ts
+++ b/src/xlsx/stream-reader.ts
@@ -27,6 +27,34 @@ export interface StreamRow {
   values: CellValue[];
 }
 
+// ── Range filter ────────────────────────────────────────────────────
+
+interface RangeFilter {
+  startRow: number;
+  endRow: number;
+  startCol: number;
+  endCol: number;
+}
+
+/**
+ * Parse a range reference like "A1:D10" into 0-based row/col bounds.
+ * Single-cell refs like "B2" are also accepted (start == end).
+ */
+function parseRangeFilter(ref: string): RangeFilter {
+  const parts = ref.split(":");
+  if (parts.length === 0 || parts.length > 2) {
+    throw new ParseError(`Invalid range reference: "${ref}"`);
+  }
+  const start = parseCellRef(parts[0]!);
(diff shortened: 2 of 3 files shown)


productdevbook (commented):
## Review

Verified locally on this branch:

- `pnpm test` (lint + typecheck + vitest): 2443/2443 passing across 86 files; no new lint or typecheck warnings introduced.
- `pnpm build`: clean.
- Diff scope is tight to `src/xlsx/stream-reader.ts`, the streaming test file, and a README example block.

### Correctness

- `applyRangeFilter` mirrors the batch reader's masking semantics in `parseWorksheet` (cells outside the column window become `null` rather than shifting indexes), so callers can reuse column indexes between batch and streaming reads. The single-cell ref path (`"B2"`) is covered by tests.
- `Math.min`/`Math.max` normalization in `parseRangeFilter` is a nice touch — slightly stricter than the batch reader's `parseRangeRef`, which doesn't normalize reversed refs. Matches user intent.
- Early-termination via `aborted = true` + `cancelSource()` is correctly idempotent: the `pull` controller closes after `aborted`, in-flight SAX callbacks short-circuit, and the consumer-side `finally` cancels again on early `break` without throwing.
- `maxRows ?? 0` with the `> 0` gate matches the batch reader's "0 = unlimited" convention (`src/xlsx/worksheet.ts:189`). Tests cover 0/undefined/negative.

### Notes / minor

- Slight duplication: there are now three essentially-identical range parsers — `parseRange` in `src/cell-utils.ts`, `parseRangeRef` in `src/xlsx/worksheet.ts`, and the new `parseRangeFilter` here. Worth a follow-up consolidation, but not blocking — the new one adds normalization and a typed `ParseError` that the others don't.
- `parsePromise.catch(() => {})` at the tail swallows errors only in the post-cancellation case (the `then` already sets `done = true` on success, so the loop only exits cleanly when the parse resolved). Reasonable, but if the SAX layer ever rejects mid-stream the consumer will hang waiting on `done` — a pre-existing latent issue, not introduced here. Worth tracking separately.
- README example is good and shows both `maxRows` and `range`.

### Conclusion

Looks correct and well-tested. Approving and merging.

PR #194, opened 2026-05-02
feat(xlsx): getCharts / addChart helpers for chart composition

productdevbook:
## Summary

- `getCharts(workbook)` walks every chart anchored on the workbook's sheets and surfaces a flat `ChartLocation[]` with `sheetName`, `sheetIndex`, `chartIndex`, and live references back to the owning `Sheet` and `Chart`. Sheets without charts are skipped; sheet/chart order matches the workbook's read order.
- `addChart(sheet, chart)` appends a `SheetChart` to a `WriteSheet`'s `charts` list, lazily allocating the array on the first call. It validates that the chart carries the writer's three mandatory fields (`type`, non-empty `series`, `anchor.from`) so authoring mistakes are caught at the call site instead of deep inside `writeXlsx`. Returns the chart instance so callers can use it inline.

## Why

This closes the explicit Phase 2 follow-up from the dashboard composition issue: `parseChart` (#190) decodes a chart, `cloneChart` (#192) bridges it onto the writer's shape, and these two helpers wrap the workbook-level walk and the per-sheet attachment so callers don't have to dig into `workbook.sheets[].charts` directly.

## Out of scope

Chart drawing-position read support (surfacing the `xdr:twoCellAnchor` `from`/`to` from the drawing layer back onto the read-side `Chart` record) is still left for a future PR — it threads through `extractSheetDrawing`, the roundtrip carry-over logic, and the `Chart` type, which is more than fits in a single focused change.

Refs #136

## Test plan

- [x] `pnpm test` (lint + typecheck + 30 454 vitest cases, including 13 new `chart-helpers` tests)
- [x] `pnpm build`
- [x] End-to-end: `addChart(sheet, …)` + `writeXlsx` emits `xl/charts/chart1.xml` with the expected `<c:barChart>` body
- [x] `getCharts` preserves identity — returned `chart` and `sheet` references mutate back into the original workbook

### DIFF
--- a/README.md
+++ b/README.md
@@ -615,6 +615,35 @@ writer can author collapse onto their write counterparts (`bar` /
 `surface`, `stock`, `ofPie`) require an explicit `options.type`
 override.
 
+#### Walking and adding charts with `getCharts` / `addChart`
+
+`getCharts(workbook)` flattens every chart anchored on the workbook's
+sheets into a single array, attaching the sheet name and indexes so
+callers don't have to walk `workbook.sheets[].charts` themselves.
+`addChart(sheet, chart)` is the symmetric writer-side helper that
+appends a `SheetChart` to a `WriteSheet`, lazily allocating the
+`charts` array on the first call:
+
+```ts
+import { addChart, getCharts, openXlsx, writeXlsx } from "hucre";
+
+// Read side — find every chart in a template workbook.
+const wb = await openXlsx(templateBytes);
+for (const { sheetName, chart } of getCharts(wb)) {
+  console.log(sheetName, chart.kinds, chart.title);
+}
+
+// Write side — declarative chart attachment.
+const dashboard = { name: "Dashboard", rows: dashboardRows };
+addChart(dashboard, {
+  type: "column",
+  title: "Q1 Revenue",
+  series: [{ name: "Revenue", values: "B2:B13", categories: "A2:A13" }],
+  anchor: { from: { row: 14, col: 0 } },
+});
+await writeXlsx({ sheets: [dashboard] });
+```
+
 ### Unified API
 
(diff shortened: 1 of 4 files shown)


productdevbook (commented):
Reviewed locally. Code, tests, and docs all hold up.

What I checked:
- `getCharts(workbook)` correctly preserves identity (same `sheet` / `chart` references), visits sheets in workbook order, and skips sheets where `charts` is undefined or empty. The 6 unit tests cover empty workbooks, the empty-array case, ordering, identity preservation, and tail-only-sheet — solid edge-case coverage.
- `addChart(sheet, chart)` lazily allocates `sheet.charts` with the canonical `sheet.charts ?? (sheet.charts = [])` pattern, returns the chart for inline use, and validates the writer's three mandatory fields (`type`, non-empty `series`, `anchor.from`). The end-to-end test verifies the chart actually round-trips through `writeXlsx` and lands as `xl/charts/chart1.xml` with the expected `<c:barChart>` body.
- `ChartLocation` carries enough sheet context (`sheet`, `sheetName`, `sheetIndex`, `chartIndex`) for callers to look up siblings without re-walking the workbook.
- Type re-exports in `src/index.ts` are correct and the README has both an example block and a function-table entry.
- Local `pnpm test` is green (1159 files / 32 974 tests) and `pnpm build` produces 183 files at 845 kB. CI is green on the PR.

Minor observations (non-blocking):
- `addChart` throws `TypeError` while `cloneChart` throws plain `Error`. Both are reasonable; `TypeError` is arguably more accurate for argument-shape errors and isn't worth changing.
- Per-series `values` aren't validated here (only that `series` is non-empty), but that's consistent with leaving deep validation to the writer.

No remaining concerns — merging.

PR #186, opened 2026-05-02
feat(xlsx): charts — read & roundtrip preservation (Phase 1)

productdevbook:
## Summary

Adds first-class **read** support and full **roundtrip preservation** for chart parts that hucre previously dropped on save. Implements **Phase 1 (Preservation)** of #32 — chart creation is a future phase.

Before this PR the `xl/charts/chartN.xml` body survived in the raw ZIP but every reference that wired it into the workbook was silently regenerated without the chart declarations, so Excel saw orphan parts and discarded the charts on next open. This PR re-injects them into `[Content_Types].xml`, the chart-bearing drawing, the drawing's rels, and the worksheet body.

## API

```ts
import { readXlsx, openXlsx, saveXlsx, parseChart } from "hucre";

const wb = await readXlsx(buf);

for (const sheet of wb.sheets) {
  for (const chart of sheet.charts ?? []) {
    console.log(chart.kinds, chart.seriesCount, chart.title);
    // e.g. ["bar"], 2, "Quarterly Sales"
  }
}

// Standalone parser when you already have the chart XML.
const chart = parseChart(xml);
```

## Model

```ts
type ChartKind =
  | "bar" | "bar3D" | "line" | "line3D"
  | "pie" | "pie3D" | "doughnut"
  | "area" | "area3D"
  | "scatter" | "bubble" | "radar"
  | "surface" | "surface3D"
  | "stock" | "ofPie";

interface Chart {
  /** Chart-type elements present in <c:plotArea>, in declaration order. */
  kinds: ChartKind[];
  /** Number of <c:ser> series across every chart-type element. */
  seriesCount: number;
  /** Plain-text title pulled from <c:title>, when present. */
  title?: string;
}
```

## Implementation

**Read.** The drawing extractor now walks `<xdr:graphicFrame>` anchors for `<c:chart r:id="...">` references and resolves them against the drawing's rels file. Each sheet exposes `sheet.charts: Chart[]` populated by `parseChart`, which surfaces the chart kind(s), series count, and plain-text title (rich-text run or strRef cache fallback).

**Roundtrip.** The save path scans `_rawEntries` for `xl/charts/chartN.xml`, `xl/charts/styleN.xml`, and `xl/charts/colorsN.xml` and:

- declares `Override` entries for every chart / chartStyle / chartColors part in `[Content_Types].xml` (new `chartIndices` / `chartStyleIndices` / `chartColorsIndices` options on `ContentTypesOptions`),
[...]

PR #195, opened 2026-05-02
feat(xlsx): surface chart cell anchors on parsed charts

productdevbook:
## Summary

`parseChart` (and therefore `sheet.charts[i]`) only surfaced what was inside `xl/charts/chartN.xml`. The cell that pinned the chart to its host sheet — written on the *drawing* layer, one level up — was thrown away. PR #190 noted this gap explicitly: "Chart drawing-position read support … still left for a future PR — it threads through `extractSheetDrawing`, the roundtrip carry-over logic, and the `Chart` type." This PR closes that gap.

## API

```ts
import { readXlsx, type ChartAnchor } from "hucre";

for (const chart of wb.sheets[0].charts ?? []) {
  // chart.anchor is now populated when the drawing pinned the chart
  // to a cell range (the common case).
  console.log(chart.anchor);
  // { from: { row: 1, col: 3 }, to: { row: 16, col: 10 } }
}
```

## Model

```ts
interface ChartAnchor {
  from: { row: number; col: number };
  to?: { row: number; col: number };
}

interface Chart {
  kinds: ChartKind[];
  seriesCount: number;
  title?: string;
  series?: ChartSeriesInfo[];
  anchor?: ChartAnchor; // NEW
}
```

The shape mirrors `SheetChart.anchor` on the writer side, so a parsed `ChartAnchor` slots straight back into a clone target without remapping.

## Implementation

`extractSheetDrawing` already walked every `xdr:twoCellAnchor` / `xdr:oneCellAnchor` / `xdr:absoluteAnchor` to collect chart part paths. It now also reads the `<xdr:from>` / `<xdr:to>` blocks via the existing `parseAnchorPosition` helper and pairs each chart path with its anchor. The reader's chart loop then attaches the anchor to the parsed `Chart` before pushing it onto `sheet.charts`.

Edge cases:
- **`xdr:twoCellAnchor`** — both `from` and `to` surfaced.
- **`xdr:oneCellAnchor`** — only `from` surfaced; `to` is intentionally absent because the size lives in `<xdr:ext>` (EMU width/height), not in a second cell.
- **`xdr:absoluteAnchor`** — `anchor` reported as `undefined`. The chart is positioned in EMU rather than against any cell; fabricating a `(0,0)` anchor would silently misrepresent the placement.
- **Missing `<xdr:from>`** — defensive parsing returns `undefined` rather than `(0,0)`.
- **Multiple charts on one drawing** — each chart gets its own anchor, paired by the drawing's `rId` → chart-part lookup.

## Out of scope

- Pixel-precision offsets (`<xdr:colOff>` / `<xdr:rowOff>`) and intrinsic `<xdr:ext>` size for one-cell anchors. The cell-grid anchor is what `SheetChart.anchor` accepts on the writer side, so the round-trip story is symmetric without those fields.
- Roundtrip preservation already works through #186 (charts pass through verbatim), so the saved drawing keeps the original anchor bytes regardless of whether the new field is consumed.

Refs #136

## Test plan

- [x] `pnpm test` (lint + typecheck + 2 470 vitest cases, including 6 new anchor tests)
- [x] `pnpm build`
- [x] `xdr:twoCellAnchor` surfaces both `from` and `to`
- [x] `xdr:oneCellAnchor` surfaces `from` only (no fabricated `to`)
- [x] `xdr:absoluteAnchor` reports `anchor` undefined
- [x] Drawing missing `<xdr:from>` reports `anchor` undefined (no `(0,0)` fabrication)
- [x] Drawing missing `<xdr:to>` falls back to `from`-only
- [x] Multiple charts on one drawing each get their own anchor
- [x] Roundtrip — re-reading a saved workbook still reports the chart's anchor

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1248,6 +1248,28 @@ export interface ChartSeriesInfo {
 }
 
 /**
+ * Cell-anchored placement for a chart on its host sheet.
+ *
+ * Mirrors the `<xdr:from>` / `<xdr:to>` pair on the drawing-layer
+ * `xdr:twoCellAnchor` (or the `<xdr:from>` alone for a
+ * `xdr:oneCellAnchor`). Coordinates are 0-based row/col indices into
+ * the worksheet — identical to the convention used by
+ * {@link SheetImage.anchor} and {@link SheetChart.anchor}, so a parsed
+ * `ChartAnchor` slots straight back into the writer's shape.
+ *
+ * `to` is optional because Excel also supports `xdr:oneCellAnchor`
+ * (chart pinned to a single cell with intrinsic size).
+ * `xdr:absoluteAnchor` (EMU-positioned) does not surface here — those
+ * charts are reported with `anchor` undefined.
+ */
+export interface ChartAnchor {
+  /** Top-left cell (`<xdr:from>`). */
+  from: { row: number; col: number };
+  /** Bottom-right cell (`<xdr:to>`). Omitted for one-cell anchors. */
+  to?: { row: number; col: number };
+}
+
+/**
  * A chart anchored on a sheet via the sheet's drawing part.
  *
  * Charts come from `xl/charts/chartN.xml`. Hucre exposes the
@@ -1266,6 +1288,13 @@ export interface Chart {
    * declaration order. Empty when the chart has no `<c:ser>` children.
    */
   series?: ChartSeriesInfo[];
+  /**
+   * Cell anchor pulled from the host drawing's `<xdr:twoCellAnchor>` /
+   * `<xdr:oneCellAnchor>`. Undefined when the drawing positions the
+   * chart with `<xdr:absoluteAnchor>` (EMU-positioned, no cell anchor)
+   * or when the drawing's anchor element is missing the `from` block.
+   */
+  anchor?: ChartAnchor;
 }
 
 // ── Workbook ───────────────────────────────────────────────────────
--- a/src/index.ts
+++ b/src/index.ts
@@ -132,7 +132,7 @@ export type {
 export { parseChart } from "./xlsx/chart-reader";
 export { cloneChart, chartKindToWriteKind } from "./xlsx/chart-clone";
 export type { CloneChartOptions, CloneChartSeriesOverride } from "./xlsx/chart-clone";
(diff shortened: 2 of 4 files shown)


productdevbook (commented):
Reviewed the chart cell-anchor read path end to end (`src/xlsx/reader.ts`, `src/_types.ts`, `test/charts.test.ts`).

**Highlights**
- `ChartAnchor` mirrors `SheetChart.anchor`, which makes the round-trip story symmetric — a parsed anchor can flow straight into a writer-side clone target.
- Per-anchor pairing inside `extractSheetDrawing` (rather than collecting anchors and chart paths separately) preserves the anchor sequence even when one drawing carries multiple charts; the new "multiple charts" test covers that explicitly.
- Defensive handling of the three anchor flavors is correctly separated: `twoCellAnchor` surfaces `from + to`, `oneCellAnchor` surfaces `from` only, and `absoluteAnchor` is short-circuited so no fabricated `(0,0)` cell anchor is reported.
- Test coverage is thorough — happy path, every degenerate variant, and a roundtrip that re-reads a saved workbook.

**Fix pushed**
- Merged `main` in to resolve a `src/index.ts` conflict (the recently merged #194 added `addChart`/`getCharts` exports next to the chart export block this PR extends with `ChartAnchor`). Kept both.
- Updated the README chart-read example to document `chart.anchor` and the per-anchor-flavor semantics, since the previous text only described kinds/series/title.
- `pnpm test` (lint + typecheck + 2 483 vitest cases including the 6 new anchor tests + roundtrip) and `pnpm build` both green locally on the merged tip.

**Soft note (non-blocking)**
- `parseAnchorPosition` defaults missing `<xdr:row>` / `<xdr:col>` text to `0`, so an `<xdr:from>` block that is literally present but missing its row/col children would still fabricate a `(0,0)` anchor — slightly weaker than the PR description's "we don't invent (0,0)" promise. Excel always writes both children, so this is a corner case rather than a real-world bug; flagging in case a future hardening pass wants to switch the helper to return `undefined` when neither child is present.
- The cloneChart docstring still says "the read-side `Chart` does not capture placement"; with `Chart.anchor` now populated, that wording is stale. Out of scope for this PR (cloneChart still legitimately requires an explicit `anchor` so callers can relocate the clone), but worth a follow-up touch-up.

Letting CI finish before merge.

PR #196, opened 2026-05-02
feat(xlsx): surface legend position and bar grouping on parsed charts

productdevbook:
## Summary

`parseChart` (and therefore `sheet.charts[i]`) already surfaced kinds, title, and series, but the chart-level layout fields — legend placement and bar grouping — were dropped on the floor. Without them a `cloneChart` round-trip silently flattens a stacked template back to clustered and a bottom-anchored legend back to the default right rail, even when the source XML declared otherwise.

This PR fills both gaps on the read side and threads them through the clone bridge so a "load template, retarget data, re-emit" workflow (the dashboard composition use case from #136) preserves the visual intent of the source chart.

## API

```ts
import { readXlsx, cloneChart } from "hucre";

console.log(source.legend);      // "bottom" | "right" | "top" | "left" | "topRight" | false | undefined
console.log(source.barGrouping); // "stacked" | "percentStacked" | "clustered" | undefined

// Clone now inherits both fields automatically when no override is given.
});
// clone.legend === source.legend
// clone.barGrouping === source.barGrouping
```

## Model

```ts
export type ChartLegendPosition = "top" | "bottom" | "left" | "right" | "topRight";
export type ChartBarGrouping = "clustered" | "stacked" | "percentStacked";

interface Chart {
  // ... existing fields ...
  legend?: false | ChartLegendPosition;   // NEW
  barGrouping?: ChartBarGrouping;         // NEW
}
```

The shapes mirror `SheetChart.legend` and `SheetChart.barGrouping` on the writer side, so a parsed value slots straight back into a clone target without remapping.

## Implementation

- `parseChart` walks the new `<c:legend>` and `<c:grouping>` elements via the existing `findChild` helper.
  - **Legend**: `<c:legendPos val=".."/>` is mapped through the OOXML short codes (`r/l/t/b/tr`) to the writer-side labels. `<c:delete val="1"/>` surfaces as `false` (Excel's "no legend" state). A legend element with no `legendPos` falls back to `"right"` since that is what Excel renders. Unknown `legendPos` values are dropped rather than fabricated.
  - **Bar grouping**: pulled from the first `<c:barChart>` element. `"standard"` is intentionally collapsed to undefined — OOXML treats it as functionally identical to `"clustered"`, and the writer's `barGrouping` field treats both as the unspecified default. Non-bar charts (`<c:lineChart>`, `<c:pieChart>`, ...) never surface a grouping even if the XML carries one.
- `cloneChart` now defaults `legend` and `barGrouping` to the source's values when the caller does not pass an override. `barGrouping` only carries over when the clone target is `bar` / `column` — coercing a stacked bar template into a `line` chart silently drops the meaningless flag rather than emitting a writer-rejected combination.

## Edge cases

- `<c:delete val="1"/>` → `legend: false` (suppressed legend).
- `<c:legend/>` with no children → `legend: "right"` (Excel default for visible legend).
- Unknown `legendPos val` → `legend: undefined` (no fabrication).
- `<c:grouping val="standard"/>` → `barGrouping: undefined` (writer default).
- `<c:grouping>` on a non-bar chart → ignored.
- Combo chart with `<c:barChart>` + `<c:lineChart>` → grouping pulled from the first bar element, in line with Excel's render.
- Override `legend: false` correctly hides a legend the source declared.

Refs #136

## Test plan

- [x] `pnpm test` (lint + typecheck + 2497 vitest cases, including 13 new parse tests + 7 new cloneChart inheritance tests)
- [x] `pnpm build`
- [x] All five `legendPos` values (`r`, `l`, `t`, `b`, `tr`) round-trip through the parser
- [x] `<c:delete val="1"/>` surfaces `legend: false`
- [x] `<c:legend/>` with no `legendPos` falls back to `right`
- [x] Unknown `legendPos` values report `undefined`
- [x] `<c:grouping val="standard"/>` is collapsed to `undefined`
- [x] `<c:grouping>` on a `<c:lineChart>` is ignored
- [x] Combo chart picks the bar grouping from the first bar element
- [x] `cloneChart` carries source legend / barGrouping through when no override is passed
- [x] `cloneChart` drops inherited `barGrouping` when the target type is not `bar`/`column`

### DIFF
--- a/README.md
+++ b/README.md
@@ -533,6 +533,13 @@ for (const sheet of wb.sheets) {
     console.log(chart.anchor);
     // e.g. { from: { row: 1, col: 3 }, to: { row: 16, col: 10 } }
 
+    // chart.legend / chart.barGrouping mirror the writer-side fields
+    // so a parsed chart slots straight back into cloneChart without
+    // remapping. `legend: false` means the source chart explicitly
+    // hid the legend; `barGrouping` only surfaces on bar/column charts.
+    console.log(chart.legend, chart.barGrouping);
+    // e.g. "bottom" "stacked"
+
     for (const s of chart.series ?? []) {
       console.log(s.kind, s.index, s.name, s.valuesRef, s.categoriesRef, s.color);
       // e.g. "bar" 0 "Revenue" "Sheet1!$B$2:$B$10" "Sheet1!$A$2:$A$10" "1F77B4"
@@ -555,7 +562,15 @@ formula) intentionally surface no `valuesRef`/`categoriesRef`.
 `twoCellAnchor` charts surface both `from` and `to`,
 `oneCellAnchor` charts surface `from` only (intrinsic size lives in
 `<xdr:ext>`), and `absoluteAnchor` charts (EMU-positioned, no cell
-anchor) report `anchor` as `undefined`.
+anchor) report `anchor` as `undefined`. `Chart.legend` and
+`Chart.barGrouping` mirror the writer-side fields of the same name:
+`legend` reports `false` when the chart explicitly suppresses the
+legend (`<c:delete val="1"/>`), `right` when `<c:legend>` is present
+without a `legendPos`, and the matching writer label otherwise;
+`barGrouping` is pulled from the first `<c:barChart>` and only
+surfaces the stacked variants (the OOXML `standard` value collapses
+to `undefined` since the writer treats it as the unspecified default,
+and non-bar charts never report a grouping).
 Sheets that hucre actively regenerates because they
 also carry hucre-managed images currently keep the chart bodies but
 lose the in-drawing chart anchor — merging hucre's drawing output
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1311,6 +1311,7 @@ export interface Chart {
    * declaration order. Empty when the chart has no `<c:ser>` children.
    */
   series?: ChartSeriesInfo[];
+  /**
    * Cell anchor pulled from the host drawing's `<xdr:twoCellAnchor>` /
    * `<xdr:oneCellAnchor>`. Undefined when the drawing positions the
    * chart with `<xdr:absoluteAnchor>` (EMU-positioned, no cell anchor)
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1254,6 +1254,29 @@ export interface ChartSeriesInfo {
  * structural metadata needed to recognize, introspect, and clone the
  * chart; the chart body is preserved verbatim through roundtrip.
  */
+/**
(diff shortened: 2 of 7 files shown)


productdevbook (commented):
## Review

Reviewed the chart legend / bar grouping read-side surface and the cloneChart inheritance logic. Overall the design is clean and consistent with the existing parser shape — the new types mirror the writer-side `SheetChart` fields, parsing is gated through the same `findChild` helpers, and the cloneChart bridge correctly drops `barGrouping` when coercing into a non-bar target.

### What I checked

- **Type model**: `ChartLegendPosition` / `ChartBarGrouping` match the writer-side options exactly, so a parsed value slots straight back into a clone target with no remapping. The `false | ChartLegendPosition` shape on `Chart.legend` mirrors `SheetChart.legend`.
- **Legend parsing**: All five OOXML `legendPos` codes (`r/l/t/b/tr`) are mapped, `<c:delete val="1"/>` correctly surfaces `false`, an empty `<c:legend/>` falls back to `right` (Excel's default), and unknown values are dropped rather than fabricated. `<c:delete val=\"0\"/>` is handled correctly via `readBoolVal`.
- **Bar grouping**: Pulled only from the first `<c:barChart>` (or `<c:bar3DChart>` via `kind === \"bar3D\"`), `\"standard\"` collapses to undefined as documented, non-bar charts never report a grouping, and combo charts pick the first bar's grouping consistently with Excel's render.
- **cloneChart inheritance**: Source values carry over only when the caller does not pass an override, override `false` correctly hides an inherited legend, and `barGrouping` is silently dropped when the clone target is not `bar`/`column` — coercing a stacked bar template into a line chart no longer emits a writer-rejected combination.
- **Test coverage**: 13 new parse tests + 7 new cloneChart tests cover every documented edge case, plus the inverse cases (`<c:delete val=\"0\"/>`, missing `legendPos`, missing `grouping`, non-bar `<c:lineChart>` with stray grouping).

### Fixes applied

- Resolved the merge conflict with #195 (anchor surfacing). Both `Chart.anchor` and the new `Chart.legend` / `Chart.barGrouping` fields now coexist on the interface, and the index re-exports cover `ChartAnchor`, `ChartBarGrouping`, and `ChartLegendPosition` together.
- Repaired a broken JSDoc comment that the merge left without its opening `/**`, which was tripping the TypeScript parser as a syntax error.
- Added README coverage for the new read-side fields under the existing chart introspection section, alongside the `Chart.anchor` explanation, so the parsed shape is documented end-to-end.

### Verification

- `pnpm test` — lint + format + typecheck + 2504 vitest cases all pass on the merged tree.
- `pnpm build` — clean (858 kB, 183 files).

CI is in progress as of this comment; I will let it complete before merging.

PR #197, opened 2026-05-02
feat(xlsx): chart axis titles — read, write, and clone-through

productdevbook:
## Summary

`parseChart` and the chart writer surfaced/emitted the chart-level title, but per-axis titles — the labels Excel renders against `<c:catAx>` / `<c:valAx>` — were missing from both sides. A `parseChart` → `cloneChart` → `writeXlsx` round-trip therefore silently dropped axis labels that were present in a template workbook, breaking the dashboard composition use case from #136 for any chart whose axes carried a name like "Quarter" or "Revenue (USD)".

This PR closes that hole at all three layers — read, write, and clone — so a templated chart's axis text survives the full retarget loop without the caller having to thread it manually.

## API

```ts

// ── Read side ──
console.log(source.axes); // { x: { title: "Quarter" }, y: { title: "Revenue (USD)" } }

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      axes: { x: { title: "Quarter" }, y: { title: "Revenue (USD)" } },
    }],
  }],
});

// ── Clone-through ──
  // axes inherit from `source` automatically; pass overrides to remap.
});
```

## Model

```ts
  title?: string;
}

interface Chart {
  // ... existing fields ...
  axes?: { x?: ChartAxisInfo; y?: ChartAxisInfo };
}

interface SheetChart {
  // ... existing fields ...
  axes?: {
    x?: { title?: string };
    y?: { title?: string };
  };
}
```

## Implementation

- **Reader (`chart-reader.ts`)**: walks the plot area's `<c:catAx>` / `<c:valAx>` children. For bar/column/line/area: `x` = `<c:catAx>`, `y` = first `<c:valAx>`. For scatter (and bubble): both axes are `<c:valAx>` — first declared (`axPos="b"`) maps to `x`, second (`axPos="l"`) to `y`. Title text is pulled from rich runs or, when the axis title is a formula, from `<c:strCache>`. Empty / whitespace-only titles are dropped so the consumer never sees a phantom blank label.
- **Writer (`chart-writer.ts`)**: `buildBarAxes` and `buildScatterAxes` now accept `xAxisTitle` / `yAxisTitle` and inject a `<c:title>` element **after** `<c:axPos>` and **before** `<c:crossAx>`. OOXML enforces this child order — putting the title in the wrong position causes Excel to ignore the label entirely. A new `buildAxisTitle` helper mirrors the chart-level `buildTitle` but with a 10pt default font (matching Excel's axis-title style vs the 14pt chart title). Pie has no axes, so the writer silently skips the field there.
- **Clone bridge (`chart-clone.ts`)**: a new `resolveAxes` helper merges the source's `axes` with per-axis overrides via the same `applyOverride` semantics already used for series fields (`undefined` = inherit, `null` = drop, string = replace). When the resolved chart type is `pie`, the axes block is dropped entirely so a doughnut/pie template that happened to carry stray axis titles doesn't poison the clone.

## Edge cases

- `<c:title>` with whitespace-only text → not surfaced (read), not emitted (write).
- Multi-run rich titles like `<a:r>Region </a:r><a:r>(2024)</a:r>` → joined into a single string.
- Cloning into `pie` strips inherited axes; cloning into `column` from a `pie` source with no axes leaves `axes` undefined.
- One axis titled, the other not → only the populated side appears in `Chart.axes` / `SheetChart.axes`.
- Override `axes: { y: { title: null } }` drops the inherited Y label while keeping X.
- `<c:title>` placement in the writer is verified to land between `<c:axPos>` and `<c:crossAx>` (the OOXML-enforced position).

Refs #136

## Test plan

- [x] `pnpm test` (lint + typecheck + 2507 vitest cases, including 7 new parser tests, 9 new writer tests, 8 new clone tests, and 1 end-to-end round-trip via `writeXlsx`)
- [x] `pnpm build`
- [x] Round-trip: `parseChart` → `cloneChart` → `writeXlsx` → `parseChart` preserves both axis titles verbatim
- [x] Scatter chart axis titles map to the X (`axPos="b"`) and Y (`axPos="l"`) value axes correctly
- [x] Pie charts silently drop `axes` on both write and clone paths
- [x] Empty / whitespace-only titles are dropped on read and write
- [x] XML-special characters (`&`, `<`, `>`, `"`) in axis titles are escaped on write
- [x] OOXML child order enforced: `<c:title>` appears after `<c:axPos>` and before `<c:crossAx>` inside both `<c:catAx>` and `<c:valAx>`

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -611,6 +611,23 @@ export interface SheetChart {
   altText?: string;
   /** Caption for the chart frame (lands in xdr:cNvPr/@title). */
   frameTitle?: string;
+  /**
+   * Per-axis labels rendered alongside the plot area. The `x` axis is
+   * the category axis for bar/column/line/area (or the bottom value
+   * axis for scatter); the `y` axis is the value axis. Ignored for
+   * `pie` charts because pie has no axes in OOXML.
+   *
+   * Each entry maps to a `<c:title>` element nested inside the
+   * matching `<c:catAx>` / `<c:valAx>`. Pass an empty string or omit
+   * the entry to skip the title — Excel renders no axis label by
+   * default.
+   */
+  axes?: {
+    /** Category axis (bar/column/line/area) or X value axis (scatter). */
+    x?: { title?: string };
+    /** Value axis. */
+    y?: { title?: string };
+  };
 }
 
 // ── Accessibility ──────────────────────────────────────────────────
@@ -1270,6 +1287,20 @@ export interface ChartAnchor {
 }
 
 /**
+ * Per-axis metadata pulled from the chart's `<c:catAx>` / `<c:valAx>`
+ * elements.
+ *
+ * Currently only the axis title is surfaced — Excel stores it as a
+ * `<c:title>` child element inside each axis, and it's the field that
+ * dashboard cloning needs to preserve through a `parseChart` →
+ * {@link cloneChart} → `writeXlsx` round-trip.
+ */
+export interface ChartAxisInfo {
+  /** Plain-text title from the axis's `<c:title>`. Omitted when absent. */
+  title?: string;
+}
+
+/**
  * A chart anchored on a sheet via the sheet's drawing part.
  *
  * Charts come from `xl/charts/chartN.xml`. Hucre exposes the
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
Reviewed the chart axis titles patch end-to-end (parser, writer, clone bridge, type exports, and tests).

What I checked
- Reader correctly walks `<c:catAx>` / `<c:valAx>` and pulls `<c:title>` text from either rich runs or a `<c:strCache>` fallback. The scatter mapping (first valAx → x, second → y) lines up with the writer side.
- Writer injects `<c:title>` strictly between `<c:axPos>` and `<c:crossAx>`, which is the OOXML-enforced child order — there's a dedicated test asserting that ordering, plus tests for XML escaping, whitespace dropping, and pie-chart skipping.
- `cloneChart` carries `axes` through the same `applyOverride` (`undefined` = inherit, `null` = drop, string = replace) used for series fields and silently drops the entire axes block when the resolved type is `pie`.
- All 2528 vitest cases pass locally, lint and typecheck are clean, build succeeds.

Fixes pushed
- Resolved the merge conflict against main (#196 added `Chart.legend` and `Chart.barGrouping`): kept all three new fields, merged the imports/exports/test sections, and re-ran `pnpm test` + `pnpm build`.
- Extended the README chart section to document the new `axes` shape on the parser, writer, and `cloneChart` paths, mirroring how `legend` / `barGrouping` are described.

CI is green on the merged branch, no remaining concerns from my side.

PR #198, opened 2026-05-02
feat(xlsx): chart data labels — read, write, and clone-through

productdevbook:
## Summary

Adds support for the small text annotations Excel paints next to each data point ("data labels"). Hucre previously had no model for `<c:dLbls>` on either the read or write side, which meant a parsed chart that carried percent labels in a pie template silently dropped them on `cloneChart` — and authors had no way to enable value labels on a freshly-built chart at all. This closes both gaps with a single typed field.

## API

```ts
import { writeXlsx, parseChart, cloneChart } from "hucre";

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      dataLabels: { showValue: true, position: "outEnd" },
    }, {
      type: "pie",
      series: [{ values: "C2:C5", categories: "A2:A5" }],
      dataLabels: { showPercent: true, position: "bestFit", separator: " — " },
    }],
  }],
});

// ── Per-series override (suppresses one series even when chart-level is on) ──
{
  type: "column",
  dataLabels: { showValue: true },
  series: [
    { values: "B2:B5" },                     // gets labels
    { values: "C2:C5", dataLabels: false },  // labels suppressed for this one
  ],
}

// ── Read + clone-through ──
console.log(source.dataLabels);
// { showValue: true, position: "outEnd" }

  // dataLabels inherit from `source`; pass an override to remap.
});
```

## Model

```ts
export type ChartDataLabelPosition =
  | "t" | "b" | "l" | "r" | "ctr"
  | "inEnd" | "inBase" | "outEnd" | "bestFit";

export interface ChartDataLabels {
  showValue?: boolean;
  showCategoryName?: boolean;
  showSeriesName?: boolean;
  showPercent?: boolean;
  position?: ChartDataLabelPosition;
  separator?: string;
}

interface SheetChart  { dataLabels?: ChartDataLabels;          /* …existing fields… */ }
interface ChartSeries { dataLabels?: ChartDataLabels | false;  /* …existing fields… */ }

// Read-side mirror
export interface ChartDataLabelsInfo { /* same fields */ }
interface Chart           { dataLabels?: ChartDataLabelsInfo; /* …existing fields… */ }
interface ChartSeriesInfo { dataLabels?: ChartDataLabelsInfo; /* …existing fields… */ }
```

## Implementation

- **Reader (`chart-reader.ts`)**: walks each chart-type element for a sibling `<c:dLbls>` and each `<c:ser>` for a child `<c:dLbls>`. The first chart-type-level block surfaces as `Chart.dataLabels` (combo charts that carry one `<c:dLbls>` per kind are reported via the dominant first kind, matching how `kinds[0]` is treated elsewhere). `<c:delete val="1">` at the root suppresses labels — this returns `undefined` rather than a "labels off" record so callers see the absence cleanly. Empty/no-toggle blocks also collapse to `undefined`.
- **Writer (`chart-writer.ts`)**: every chart-type builder (`buildBarChart`, `buildLineChart`, `buildAreaChart`, `buildPieChart`, `buildScatterChart`) now threads the chart-level `dataLabels` through to `buildSeries` (per-series override semantics) and emits the chart-type-level `<c:dLbls>` after the last `<c:ser>` and before `<c:gapWidth>` / `<c:axId>` / `<c:marker>` (the OOXML-enforced position — Excel ignores the block if it lands anywhere else). At the series level, `<c:dLbls>` sits between `<c:spPr>` and `<c:cat>`/`<c:val>` per CT_BarSer / CT_LineSer / etc. The toggle order inside `<c:dLbls>` follows the CT_DLbls schema (`dLblPos`, `showLegendKey`, `showVal`, `showCatName`, `showSerName`, `showPercent`, `showBubbleSize`, `separator`) — this is enforced or the file fails Excel's strict-validation pass.
- **`dataLabels: false` semantics**: Excel's idiom for "this chart has labels on, but suppress them for series N" is a per-series `<c:dLbls><c:dLbl><c:idx val="0"/><c:delete val="1"/></c:dLbl><c:delete val="1"/></c:dLbls>` block. The writer emits exactly that shape.
- **Clone bridge (`chart-clone.ts`)**: `resolveChartDataLabels` and `resolveSeriesDataLabels` carry the same `undefined`/`null`/value override grammar used everywhere else in `cloneChart` — `undefined` inherits, `null` drops, an object replaces, and at the series level `false` suppresses. The read-side `ChartDataLabelsInfo` is a structural superset of the write-side `ChartDataLabels`, so the resolver simply spreads when inheriting.

## Edge cases

- `<c:dLbls>` containing only `<c:delete val="1">` → not surfaced (read), so it does not poison a clone.
- All `<c:show*>` toggles set to `0` and no `dLblPos` → reader collapses to `undefined` (no point surfacing an effectively-empty record).
- `<c:dLblPos val="moonbase">` → reader silently ignores (Excel does the same — it falls back to its built-in default for the kind).
- `<c:showVal val="true">` and `<c:showVal val="1">` both parse as `true`; the OOXML spec accepts both shapes.
- Override `seriesOverrides[i].dataLabels = null` drops the inherited series block but lets the chart-level default apply.
- Override `seriesOverrides[i].dataLabels = false` suppresses for that series alone, even when the chart-level default has labels on — written as `<c:dLbls><c:delete val="1">` per Excel's convention.
- Combo charts (multiple chart-type elements): the chart-level `dataLabels` reports the first chart-type element's block, matching the existing `kinds[0]` precedent.
- Pie / doughnut: `showPercent` and `position: "bestFit"` are the meaningful combination — round-trips cleanly. Multi-toggle labels with a custom `separator` (e.g. `"; "`, `" — "`) round-trip with XML-special characters escaped.

Refs #136

## Test plan

- [x] `pnpm test` (lint + typecheck + 2531 vitest cases — adds 11 writer tests, 8 reader tests, 7 clone tests, and 1 end-to-end `parseChart → cloneChart → writeXlsx → parseChart` round-trip)
- [x] `pnpm build`


productdevbook (commented):
Reviewed the chart data labels feature end-to-end.

**Correctness / design**
- `<c:dLbls>` placement in the writer follows the OOXML CT_BarChart / CT_LineChart / CT_AreaChart / CT_PieChart / CT_ScatterChart schemas: after the last `<c:ser>` and before `gapWidth` / `overlap` / `marker` / `axId`. The toggle order inside the block (`dLblPos` → `showLegendKey` → `showVal` → `showCatName` → `showSerName` → `showPercent` → `showBubbleSize` → `separator`) matches CT_DLbls so Excel doesn't ignore the block on strict validation.
- The series-level `<c:dLbls>` lands between `<c:spPr>` and `<c:cat>` / `<c:val>` per CT_BarSer / CT_LineSer.
- `dataLabels: false` correctly emits the `<c:dLbl idx=0 delete=1/>` + root `<c:delete val=1/>` shape Excel uses for "labels on globally, suppressed for this series".
- Clone bridge resolvers (`resolveChartDataLabels`, `resolveSeriesDataLabels`) preserve the project's `undefined` (inherit) / `null` (drop) / object (replace) grammar consistently.
- Reader collapses delete-only and all-zeroes blocks to `undefined`, ignores invalid `dLblPos` values, and accepts `1`/`0`/`true`/`false` on the show-toggle attributes — matches the OOXML xsd:boolean grammar.

**Edge cases covered**
- Combo charts (`kinds[0]` precedent); multi-toggle pie + `bestFit` + custom separator round-trip; XML escaping of separator characters; series-level override independent of chart-level; `<c:dLbls>` as `<c:delete val=1>` not surfacing.

**Fixes applied in this push**
- Resolved merge conflicts with #197 (chart axis titles): `dataLabels` and `axes` now coexist on `Chart`, `SheetChart`, `CloneChartOptions`, and the test files keep the two describe blocks separate.
- Updated `README.md` to document `Chart.dataLabels`, `SheetChart.dataLabels`, the per-series override (including `false`), and the `cloneChart` override grammar.

**Test results**
- `pnpm test`: 2555 tests pass (lint + typecheck + vitest).
- `pnpm build`: clean.

No remaining concerns. Merging once CI is green.

PR #199, opened 2026-05-02
feat(xlsx): chart axis gridlines — read, write, and clone-through

productdevbook:
## Summary

`parseChart` and the chart writer already carried the axis title across the read / write / clone bridge (#197), but the gridlines Excel paints across the plot area (`<c:majorGridlines>` / `<c:minorGridlines>`) were ignored on both sides. A `parseChart` → `cloneChart` → `writeXlsx` round-trip therefore stripped the reference lines from any template chart that had them on, breaking the dashboard composition flow from #136 for the common "value-axis with major gridlines" look.

This PR closes that hole at all three layers — read, write, and clone — so a templated chart's gridline configuration survives the full retarget loop without the caller having to thread it manually.

## API

```ts

// ── Read side ──
console.log(source.axes);
// { y: { gridlines: { major: true } } }

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      axes: {
        y: { title: "Revenue (USD)", gridlines: { major: true, minor: true } },
        x: { gridlines: { major: true } },
      },
    }],
  }],
});

// ── Clone-through ──
  // gridlines inherit from `source` automatically; pass overrides to remap.
});

// Drop the inherited gridlines on a single axis:
const clean = cloneChart(source, {
  axes: { y: { gridlines: null } },
});
```

## Model

```ts
export interface ChartAxisGridlines {
  major?: boolean;
  minor?: boolean;
}

interface ChartAxisInfo {
  title?: string;
  gridlines?: ChartAxisGridlines;
}

interface SheetChart {
  // ... existing fields ...
  axes?: {
    x?: { title?: string; gridlines?: ChartAxisGridlines };
    y?: { title?: string; gridlines?: ChartAxisGridlines };
  };
}
```

## Implementation

- **Reader (`chart-reader.ts`)**: walks each `<c:catAx>` / `<c:valAx>` for a `<c:majorGridlines>` / `<c:minorGridlines>` child and surfaces the flags. Nested `<c:spPr>` styling is tolerated — the presence of the outer element is what flips the toggle, matching how Excel reads the flag back. `parseAxisInfo` now considers both title and gridlines when deciding whether the axis surfaces; an axis with only gridlines (no title) still yields a `ChartAxisInfo`.
- **Writer (`chart-writer.ts`)**: `buildBarAxes` and `buildScatterAxes` accept `xGridlines` / `yGridlines` and slot the gridline elements between `<c:axPos>` and the optional `<c:title>`. OOXML enforces this child order — putting gridlines anywhere else causes Excel's strict validator to reject the file. Major-before-minor is also schema-enforced. Pie charts have no axes, so the writer silently skips the field there.
- **Clone bridge (`chart-clone.ts`)**: a new `applyGridlinesOverride` helper carries the same `undefined` (inherit) / `null` (drop) / object (replace) grammar already used for axis titles. `resolveAxes` now folds title and gridlines into the same `axes.{x,y}` record, dropping it entirely when both sides resolve to `undefined`. When the resolved chart type is `pie` the whole axes block is dropped, so a doughnut/pie template that happened to carry stray gridlines does not poison the clone.

## Edge cases

- `<c:majorGridlines>` containing nested `<c:spPr>` styling → the toggle still flips on (read), and the writer emits an empty element on round-trip (the styling is not yet round-tripped, but the gridline itself is preserved).
- Both flags omitted / set to `false` → reader returns `undefined`, writer emits no gridline element, clone returns `axes: undefined` if no other axis field is set.
- Override `axes: { y: { gridlines: null } }` drops the inherited gridlines while keeping the inherited title. Override `axes: { y: { gridlines: { major: true, minor: true } } }` replaces the inherited block wholesale.
- Combo charts that declare both `<c:catAx>` and one or more `<c:valAx>`: `parseAxes` already maps `x = catAx`, `y = first valAx` for bar/column/line/area, and `x = first valAx (axPos="b")`, `y = second valAx (axPos="l")` for scatter — the gridline detection inherits that mapping with no extra logic.
- Cloning into `pie` strips inherited gridlines along with axis titles; cloning into `column` from a `pie` source with no axes leaves `axes` undefined.

Refs #136

## Test plan

- [x] `pnpm test` (lint + typecheck + 2579 vitest cases — adds 7 parser tests, 9 writer tests, 8 clone tests)
- [x] `pnpm build`

### DIFF
--- a/README.md
+++ b/README.md
@@ -540,10 +540,11 @@ for (const sheet of wb.sheets) {
     console.log(chart.legend, chart.barGrouping);
     // e.g. "bottom" "stacked"
 
-    // chart.axes carries per-axis labels pulled from <c:catAx>/<c:valAx>
-    // <c:title>. Only populated axes show up — pie/doughnut never do.
+    // chart.axes carries per-axis labels and gridline visibility pulled
+    // from <c:catAx>/<c:valAx>. Only populated axes show up — pie/doughnut
+    // never do.
     console.log(chart.axes);
-    // e.g. { x: { title: "Quarter" }, y: { title: "Revenue (USD)" } }
+    // e.g. { x: { title: "Quarter" }, y: { title: "Revenue (USD)", gridlines: { major: true } } }
 
     // chart.dataLabels surfaces the chart-type-level <c:dLbls> block.
     // showValue / showCategoryName / showSeriesName / showPercent and
@@ -584,11 +585,14 @@ without a `legendPos`, and the matching writer label otherwise;
 surfaces the stacked variants (the OOXML `standard` value collapses
 to `undefined` since the writer treats it as the unspecified default,
 and non-bar charts never report a grouping). `Chart.axes` mirrors
-the writer-side `SheetChart.axes` and surfaces per-axis labels: `x`
-is the category axis (or, for scatter, the first value axis) and
-`y` is the value axis. Empty / whitespace-only `<c:title>` text is
-dropped, charts without any axis label leave `axes` undefined, and
-pie/doughnut charts (which have no axes in OOXML) never report one.
+the writer-side `SheetChart.axes` and surfaces per-axis labels and
+gridline visibility: `x` is the category axis (or, for scatter, the
+first value axis) and `y` is the value axis. Empty / whitespace-only
+`<c:title>` text is dropped, `gridlines: { major, minor }` flips on
+when the matching `<c:majorGridlines>` / `<c:minorGridlines>` element
+is present (any nested styling is tolerated), charts without any
+axis label or gridline leave `axes` undefined, and pie/doughnut
+charts (which have no axes in OOXML) never report one.
 `Chart.dataLabels` mirrors the writer-side `SheetChart.dataLabels`
 and surfaces the toggles Excel carries inside `<c:dLbls>`
 (`showValue`, `showCategoryName`, `showSeriesName`, `showPercent`,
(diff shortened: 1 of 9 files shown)


productdevbook (commented):
## Review

Read through the diff and walked the three layers (`chart-reader.ts`, `chart-writer.ts`, `chart-clone.ts`) plus the new test blocks. The implementation cleanly mirrors the axis-title infrastructure landed in #197 and reuses the same `undefined` (inherit) / `null` (drop) / object (replace) override grammar that was established for titles and chart-level data labels — consistency is good.

### Correctness checks

- **Reader**: `parseAxisGridlines` flips on with the mere presence of `<c:majorGridlines>` / `<c:minorGridlines>`, tolerating nested `<c:spPr>` styling — matches Excel's read semantics. Returning `undefined` when neither child is present keeps the parsed shape lean and avoids redundant write-back of `{ major: false, minor: false }`.
- **Writer**: `buildAxisGridlines` emits the elements in the OOXML-required order (after `<c:axPos>`, before `<c:title>`, major before minor), confirmed by the new positional test. Pie charts silently skip gridlines because the pie branch never calls `buildBarAxes` / `buildScatterAxes`.
- **Clone**: `applyGridlinesOverride` correctly handles all four grammar branches; `{ major: false, minor: false }` collapses to `undefined` (drop), and an override with only one flag set replaces the inherited block wholesale (consistent with how a single-axis title override replaces wholesale).
- **Type safety**: `ChartAxisGridlines` is exported from `index.ts`; readers and writers narrow on it everywhere.

### Edge cases I checked

- Combo charts (`<c:catAx>` + multiple `<c:valAx>`): mapping inherits the existing `parseAxes` logic, no extra branching needed.
- Scatter X (`axPos="b"`) and Y (`axPos="l"`) gridlines map to the correct value axis — covered by the new scatter-specific tests.
- Pie source carrying stray `axes.gridlines` is dropped on clone.
- An axis with only gridlines (no title) still surfaces a `ChartAxisInfo`, which is the new behavior the diff intentionally introduced.

### Fixes I added

Pushed one follow-up commit (`cb4e7f8`):

- **README**: extended the `Chart.axes` example to show `gridlines`, updated the writer paragraph to describe the gridline element placement and major-before-minor ordering, and updated the `cloneChart` paragraph to document the `axes.{x,y}.gridlines` override grammar (`null` to drop, object to replace).
- **Integration test**: added `round-trips axis gridlines through parseChart → cloneChart → writeXlsx → parseChart` mirroring the axis-title round-trip from #197. Confirms both major and minor flags survive the full retarget loop on `<c:catAx>` and `<c:valAx>`.

`pnpm test` (lint + typecheck + 2580 vitest cases — one more than the PR claimed because of the new round-trip) and `pnpm build` both pass on the updated branch.

No remaining concerns. Ready to merge once CI is green.

PR #200, opened 2026-05-02
feat(xlsx): doughnut as a writable chart kind — read, write, clone-through

productdevbook:
## Summary

`doughnut` is one of the chart families issue #136 explicitly calls out for the dashboard composition flow: a template workbook stores a donut chart, the caller pulls it via `parseChart`, retargets the data ranges through `cloneChart`, and re-emits the result. Previously the writer had no doughnut case — `cloneChart` silently flattened doughnut sources into plain pies, dropping the hole the dashboard relied on. This PR closes that gap at all three layers (read, write, clone-through) so a templated doughnut survives the full retarget loop with its hole intact.

## API

```ts
import { readXlsx, writeXlsx, parseChart, cloneChart } from "hucre";

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Distribution",
    rows: [["Category", "Share"], ["Cloud", 42], ["On-prem", 28], ["Hybrid", 30]],
    charts: [{
      type: "doughnut",
      title: "Workload Mix",
      holeSize: 60,                 // 10–90 (Excel's UI band; default 50)
      series: [{ values: "B2:B4", categories: "A2:A4" }],
    }],
  }],
});

// ── Read side ──
console.log(source.kinds, source.holeSize);
// ["doughnut"] 65

// ── Clone-through ──
// Default keeps the doughnut shape and inherits holeSize from the template.
});
// → { type: "doughnut", holeSize: 65, ... }

// Override the hole or flatten to a plain pie when the dashboard wants one:
const flat = cloneChart(source, {
  type: "pie",                       // drops holeSize
});
```

## Implementation

- **`WriteChartKind`** gains `"doughnut"`. `SheetChart` gains a numeric `holeSize?` field documented as the Excel 10–90 band with a 50 default.
- **`chart-writer`**: a new `buildDoughnutChart` mirrors the pie case (`varyColors=1`, every declared series, no axes) plus the schema-required `<c:firstSliceAng val="0">` and `<c:holeSize val=".."/>` block. `clampHoleSize` rounds, replaces non-finite values with the default, and clamps to 10–90 — matching Excel's UI behavior. `chartKindElement` learns the `doughnut → c:doughnutChart` mapping.
- **`chart-reader`**: `parseHoleSize` reads `<c:holeSize>` off the first `<c:doughnutChart>` and surfaces it on `Chart.holeSize`. The 1–99 OOXML schema range is honored on read; clamping happens on write.
- **`chart-clone`**: `chartKindToWriteKind("doughnut")` now returns `"doughnut"` (was `"pie"`). `cloneChart` carries `source.holeSize` onto the clone when the resolved type is `doughnut`, accepts an `options.holeSize` override, and drops the field whenever the type resolves to anything else (so flattening to pie or column does not leak the hole). The pie-guard around `axes` is widened to `pie` + `doughnut` since neither has axes.

## Edge cases

- `holeSize: NaN` / `Infinity` → fall back to Excel's 50 default (writer never emits an invalid integer).
- `holeSize: 5` / `120` → clamped to 10 / 90 respectively.
- Float `holeSize` (e.g. 42.7) → rounded to 43.
- `holeSize` set on a non-doughnut chart → silently ignored by the writer.
- Doughnut clone with `axes` overrides → axes block dropped (doughnut has no axes).
- Multi-ring templates (multiple `<c:ser>` under `<c:doughnutChart>`) → every series carried through, matching how Excel renders concentric rings.

## Test plan

- [x] `pnpm test` (oxlint + oxfmt + tsgo + vitest): all 2576 tests pass, including the new write / read / clone / round-trip / integration suites added across `test/charts-write.test.ts`, `test/charts.test.ts`, and `test/chart-clone.test.ts`.
- [x] `pnpm build` succeeds.

Refs #136

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -540,15 +540,14 @@ export interface SheetImage {
 
 /**
  * Chart kinds supported by {@link writeXlsx} when authoring charts via
- * {@link WriteSheet.charts}. Phase 1 covers the four most common chart
- * families — bar/column, line, pie, and scatter — plus area as a
- * thin variant of `line`.
+ * {@link WriteSheet.charts}. Covers the most common chart families —
+ * bar/column, line, pie, doughnut, scatter, and area.
  *
  * Distinct from the read-side {@link ChartKind} (which mirrors the
  * full set of OOXML chart-type element local names) — the write side
  * exposes only the kinds the chart author can emit today.
  */
-export type WriteChartKind = "bar" | "column" | "line" | "pie" | "scatter" | "area";
+export type WriteChartKind = "bar" | "column" | "line" | "pie" | "doughnut" | "scatter" | "area";
 
 /**
  * Where a data label is placed relative to its data point.
@@ -667,9 +666,16 @@ export interface SheetChart {
    */
   barGrouping?: "clustered" | "stacked" | "percentStacked";
   /**
+   * Doughnut hole size as a percentage of the outer radius. Accepted
+   * range: 10 – 90 (Excel's UI clamps values outside this band).
+   * Default: `50` — the Excel default. Ignored for non-doughnut chart
+   * kinds.
+   */
+  holeSize?: number;
+  /**
    * Whether the legend is shown and where. Default: `"right"` for
-   * pie/bar/line/area, `"bottom"` for scatter. Pass `false` to hide
-   * the legend.
+   * pie/doughnut/bar/line/area, `"bottom"` for scatter. Pass `false`
+   * to hide the legend.
    */
   legend?: false | "top" | "bottom" | "left" | "right" | "topRight";
   /** Show the chart-level title element. Default: `true` when `title` is set. */
@@ -689,7 +695,7 @@ export interface SheetChart {
    * Per-axis labels rendered alongside the plot area. The `x` axis is
    * the category axis for bar/column/line/area (or the bottom value
    * axis for scatter); the `y` axis is the value axis. Ignored for
-   * `pie` charts because pie has no axes in OOXML.
+   * `pie` and `doughnut` charts because they have no axes in OOXML.
    *
    * Each entry maps to a `<c:title>` element nested inside the
    * matching `<c:catAx>` / `<c:valAx>`. Pass an empty string or omit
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed end to end (read, write, clone-through, integration tests, README) and the implementation matches the contract the PR description promises.

**Strengths**
- The OOXML element order in `buildDoughnutChart` is correct: `varyColors → ser… → dLbls? → firstSliceAng → holeSize`. Excel's strict-mode validator accepts this. `firstSliceAng val="0"` is harmless but useful for round-trip stability against templates that already declare it.
- `clampHoleSize` covers the full numeric edge surface — `undefined`, `NaN`, `±Infinity` all collapse to the Excel default (50), floats round, and the 10–90 UI band is enforced. The reader keeps the wider 1–99 OOXML schema range so external templates don't get rejected on read but are re-clamped on write — that asymmetry is intentional and matches Excel's own behavior.
- `chartKindToWriteKind("doughnut") → "doughnut"` (was `"pie"`) is the load-bearing fix for the dashboard composition flow described in #136. The pie-axes guard is correctly widened to `pie` and `doughnut` since neither has axes in OOXML.
- `cloneChart` only carries `holeSize` when the resolved type is `doughnut`, so flattening to pie via `options.type = "pie"` cleanly drops the hole hint without leaking it into the writer (covered by the dedicated test).
- Test coverage is thorough: writer XML shape, clamping bounds, multi-series rings, parser schema-range handling, end-to-end through `writeXlsx` → `parseChart`, and the cloneChart round-trip.

**Fixes I applied in `b87b356`**
- Resolved a 3-file merge conflict against `main` (gridlines feature merged in `2c5289e` after this PR was opened):
  - `src/_types.ts`: kept the gridlines documentation on `axes` while widening the no-axes carve-out to cover `doughnut` alongside `pie`.
  - `test/chart-clone.test.ts`: kept both the gridlines round-trip integration test (from main) and the new doughnut round-trip test (from this branch) as separate cases.
  - `test/charts.test.ts`: kept the `parseChart — axis gridlines` and `parseChart — doughnut hole size` describe blocks side by side.
- Updated `README.md` to reflect the new surface: the writer now lists seven chart families (bar/column/line/pie/doughnut/scatter/area), both `pie` and `doughnut` are noted as ignoring the `axes` block, the `cloneChart` kind table now maps `doughnut → doughnut` (was `→ pie`), the `holeSize` knob is documented on the write side, and `Chart.holeSize` is documented on the read side.

**Verification**
- `pnpm test` (oxlint + oxfmt + tsgo + vitest): 2601 tests pass on the merge commit.
- `pnpm build`: succeeds.

No remaining concerns. Will merge once CI completes.

PR #201, opened 2026-05-02
feat(xlsx): line and area chart stacking — read, write, clone-through

productdevbook:
## Summary

Bar / column charts already accept a `barGrouping` option for stacked / percentStacked layouts (PRs #186, #187, #196). The matching `<c:grouping>` element on `<c:lineChart>` and `<c:areaChart>` was hardcoded to `"standard"` on the writer side and ignored on the parser side. As a result, stacked line / area templates flattened to plain (non-stacked) ones on every `parseChart` → `cloneChart` → `writeXlsx` round-trip, and stacked line / stacked area charts could not be authored from scratch through `writeXlsx`. Issue #136 explicitly calls out a stacked-bar dashboard variant — the same pattern applies to line / area, and this PR closes that gap at all three layers.

## API

```ts
import { writeXlsx, parseChart, cloneChart } from "hucre";

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Trend",
    rows: [
      ["Quarter", "Cloud", "On-prem"],
      ["Q1", 10, 20],
      ["Q2", 15, 18],
      ["Q3", 20, 16],
      ["Q4", 25, 14],
    ],
    charts: [{
      type: "line",
      lineGrouping: "stacked",          // standard | stacked | percentStacked
      title: "Workload trend",
      series: [
        { name: "Cloud",   values: "B2:B5", categories: "A2:A5" },
        { name: "On-prem", values: "C2:C5", categories: "A2:A5" },
      ],
      anchor: { from: { row: 6, col: 0 } },
    }, {
      type: "area",
      areaGrouping: "percentStacked",   // standard | stacked | percentStacked
      series: [
        { name: "Cloud",   values: "B2:B5", categories: "A2:A5" },
        { name: "On-prem", values: "C2:C5", categories: "A2:A5" },
      ],
      anchor: { from: { row: 25, col: 0 } },
    }],
  }],
});

// ── Read side ──
const wb = await readXlsx(templateBytes);
const source = wb.sheets[0].charts![0];
console.log(source.lineGrouping); // "stacked"  (from <c:lineChart><c:grouping val="stacked"/></c:lineChart>)
console.log(source.areaGrouping); // "percentStacked" (when the source is an areaChart)

// ── Clone-through ──
const clone = cloneChart(source, {
  anchor: { from: { row: 14, col: 0 } },
  seriesOverrides: [{ values: "Dashboard!$B$2:$B$13" }],
});
// clone.lineGrouping === "stacked"

// Override per-family:
[...]

PR #202, opened 2026-05-02
feat(xlsx): chart axis scale and number format — read, write, and clone-through

productdevbook:
## Summary

`parseChart` and the chart writer carried axis title (#197), gridlines (#199), and data labels (#198) through the read / write / clone bridge, but the value-axis numeric scaling (`<c:min>` / `<c:max>` / `<c:majorUnit>` / `<c:minorUnit>` / `<c:logBase>`) and the tick-label number format (`<c:numFmt>`) were ignored on both sides. A `parseChart` → `cloneChart` → `writeXlsx` round-trip therefore stripped pinned axis bounds and currency / percent / thousands formatting from any template chart that had them, breaking the dashboard-composition flow from #136 for the very common "explicit min/max + currency tick labels" look.

This PR closes that gap at all three layers — read, write, and clone — so a templated chart's scale and tick-label format survive the full retarget loop without the caller having to thread them manually.

## API

```ts

// ── Read side ──
console.log(source.axes);
// {
//   y: {
//     scale: { min: 0, max: 100, majorUnit: 25 },
//     numberFormat: { formatCode: "$#,##0" },
//   },
// }

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      axes: {
        y: {
          title: "Revenue (USD)",
          gridlines: { major: true },
          scale: { min: 0, max: 100, majorUnit: 25 },
          numberFormat: { formatCode: "$#,##0" },
        },
      },
    }],
  }],
});

// ── Clone-through ──
  // scale and numberFormat inherit from `source` automatically; pass
  // overrides to remap.
});

// Drop the inherited scale on a single axis:
const auto = cloneChart(source, {
  axes: { y: { scale: null } },
});
```

## Model

```ts
export interface ChartAxisScale {
  min?: number;
  max?: number;
  majorUnit?: number;   // tick spacing, must be > 0
  minorUnit?: number;   // tick spacing, must be > 0
  logBase?: number;     // 2..1000
}

export interface ChartAxisNumberFormat {
  formatCode: string;       // OOXML number-format code
  sourceLinked?: boolean;   // true → ignore formatCode, inherit from data range
}

interface ChartAxisInfo {
  title?: string;
  gridlines?: ChartAxisGridlines;
  scale?: ChartAxisScale;
  numberFormat?: ChartAxisNumberFormat;
}

interface SheetChart {
  axes?: {
    x?: { title?: string; gridlines?: ChartAxisGridlines; scale?: ChartAxisScale; numberFormat?: ChartAxisNumberFormat };
    y?: { title?: string; gridlines?: ChartAxisGridlines; scale?: ChartAxisScale; numberFormat?: ChartAxisNumberFormat };
  };
}
```

## Implementation

- **Reader (`chart-reader.ts`)**: `parseAxisScale` walks `<c:scaling>` for `<c:min>` / `<c:max>` / `<c:logBase>` and the axis itself for `<c:majorUnit>` / `<c:minorUnit>` (tick spacings live directly under `<c:catAx>` / `<c:valAx>` per CT_CatAx / CT_ValAx, not inside `<c:scaling>`). `parseAxisNumberFormat` reads `<c:numFmt formatCode="…" sourceLinked="…"/>` and normalizes `sourceLinked` from the boolean-attribute set Excel emits (`0` / `1` / `"true"` / `"false"`). Empty `formatCode` strings collapse the whole record so the consumer never sees `<c:numFmt formatCode=""/>`. `parseAxisInfo` now folds title / gridlines / scale / numberFormat into the same axis record.
- **Writer (`chart-writer.ts`)**: a single `AxisRenderOptions` struct now carries the four per-axis fields through `buildBarAxes` / `buildScatterAxes`. `buildAxisScaling` slots `<c:logBase>` before `<c:orientation>` and `<c:max>` / `<c:min>` after it (the order CT_Scaling enforces). `buildAxisNumFmt` lands `<c:numFmt>` after the optional `<c:title>` and before `<c:crossAx>`; `buildAxisTickUnits` lands `<c:majorUnit>` / `<c:minorUnit>` after `<c:crossBetween>`, the position CT_ValAx / CT_CatAx require. Pie and doughnut have no axes, so the writer silently skips the field there. `normalizeAxisScale` drops non-finite numbers, zero/negative tick spacings, log bases outside 2..1000, and degenerate `min >= max` ranges (preserving min, dropping max) so the writer can never emit a `<c:scaling>` Excel rejects.
- **Clone bridge (`chart-clone.ts`)**: `applyScaleOverride` and `applyNumberFormatOverride` carry the same `undefined` (inherit) / `null` (drop) / object (replace) grammar already used for axis titles and gridlines. `resolveAxes` now folds title / gridlines / scale / numberFormat into the same `axes.{x,y}` record, dropping it entirely when all four resolve to `undefined`. When the resolved chart type is `pie` or `doughnut` the whole axes block is dropped, so a doughnut template that happened to carry stray scale/numFmt values does not poison the clone.

## Edge cases

- `<c:scaling>` with only `<c:orientation>` (Excel's autoscale baseline) → reader returns `undefined`, writer emits the same minimal `<c:scaling>`.
- `<c:majorUnit val="0"/>` / negative tick spacing → reader filters it out, writer never emits it. Same for `Number.NaN`.
- `min >= max` on the write side → max is dropped, min is preserved.
- `logBase` outside 2..1000 → writer drops it (Excel rejects values outside that band).
- `numFmt formatCode=""` → reader returns `undefined`, writer skips emission.
- Override `axes: { y: { scale: null } }` drops the inherited scale while keeping title / gridlines / numFmt.
- Override `axes: { y: { scale: { max: 50 } } }` replaces the inherited scale wholesale (does **not** merge into a `{ min: 0, max: 50, majorUnit: 25 }` — wholesale replace is the contract).
- Cloning into `pie` / `doughnut` strips inherited scale and numberFormat along with axis titles / gridlines.
- XML-special characters in `formatCode` (e.g. `'"<x>"&"y"'`) are escaped on write.
- Spec-enforced child order on the writer side: `<c:logBase>` precedes `<c:orientation>` inside `<c:scaling>`; `<c:numFmt>` lands between `<c:title>` and `<c:crossAx>`; `<c:majorUnit>` / `<c:minorUnit>` land after `<c:crossBetween>`.

Refs #136

## Test plan

- [x] `pnpm test` (lint + typecheck + 2642 vitest cases — adds 5 parser tests for scale, 4 parser tests for number format, 8 writer tests for scale, 7 writer tests for number format, 7 clone tests for scale, 7 clone tests for number format, plus 1 end-to-end round-trip via `parseChart` → `cloneChart` → `writeXlsx` → `parseChart`)
- [x] `pnpm build`
- [x] Round-trip: pinned `min` / `max` / `majorUnit` and `formatCode: "\$#,##0"` survive the full retarget loop verbatim
- [x] Scatter scale lands on the X (`axPos="b"`) value axis when `axes.x.scale` is set
- [x] Pie / doughnut charts silently drop scale / numberFormat on both write and clone paths
- [x] Spec-enforced child order verified by index-based assertions (orientation before max/min; logBase before orientation; majorUnit/minorUnit after crossBetween; numFmt between title and crossAx)

### DIFF
--- a/README.md
+++ b/README.md
@@ -540,11 +540,19 @@ for (const sheet of wb.sheets) {
     console.log(chart.legend, chart.barGrouping);
     // e.g. "bottom" "stacked"
 
-    // chart.axes carries per-axis labels and gridline visibility pulled
-    // from <c:catAx>/<c:valAx>. Only populated axes show up — pie/doughnut
-    // never do.
+    // chart.axes carries per-axis labels, gridline visibility, numeric
+    // scaling, and tick-label number format pulled from <c:catAx>/<c:valAx>.
+    // Only populated axes show up — pie/doughnut never do.
     console.log(chart.axes);
-    // e.g. { x: { title: "Quarter" }, y: { title: "Revenue (USD)", gridlines: { major: true } } }
+    // e.g. {
+    //   x: { title: "Quarter" },
+    //   y: {
+    //     title: "Revenue (USD)",
+    //     gridlines: { major: true },
+    //     scale: { min: 0, max: 100, majorUnit: 25 },
+    //     numberFormat: { formatCode: "$#,##0" },
+    //   },
+    // }
 
     // chart.dataLabels surfaces the chart-type-level <c:dLbls> block.
     // showValue / showCategoryName / showSeriesName / showPercent and
@@ -585,14 +593,22 @@ without a `legendPos`, and the matching writer label otherwise;
 surfaces the stacked variants (the OOXML `standard` value collapses
 to `undefined` since the writer treats it as the unspecified default,
 and non-bar charts never report a grouping). `Chart.axes` mirrors
-the writer-side `SheetChart.axes` and surfaces per-axis labels and
-gridline visibility: `x` is the category axis (or, for scatter, the
-first value axis) and `y` is the value axis. Empty / whitespace-only
-`<c:title>` text is dropped, `gridlines: { major, minor }` flips on
-when the matching `<c:majorGridlines>` / `<c:minorGridlines>` element
-is present (any nested styling is tolerated), charts without any
-axis label or gridline leave `axes` undefined, and pie/doughnut
-charts (which have no axes in OOXML) never report one.
+the writer-side `SheetChart.axes` and surfaces per-axis labels,
+gridline visibility, numeric scaling and tick-label number format:
+`x` is the category axis (or, for scatter, the first value axis)
+and `y` is the value axis. Empty / whitespace-only `<c:title>` text
+is dropped, `gridlines: { major, minor }` flips on when the matching
+`<c:majorGridlines>` / `<c:minorGridlines>` element is present (any
+nested styling is tolerated), `scale: { min, max, majorUnit, minorUnit, logBase }`
+captures the explicit `<c:min>` / `<c:max>` / `<c:logBase>` (under
+`<c:scaling>`) and `<c:majorUnit>` / `<c:minorUnit>` (direct axis
+children) — fields Excel auto-computes are left off so the round
+trip never accidentally pins a value, and zero or negative tick
+spacings are filtered out — and `numberFormat: { formatCode, sourceLinked }`
+mirrors `<c:numFmt>` (an empty `formatCode` collapses the record).
+Charts without any axis label, gridline, scale, or number format
+leave `axes` undefined, and pie/doughnut charts (which have no axes
+in OOXML) never report one.
 `Chart.dataLabels` mirrors the writer-side `SheetChart.dataLabels`
 and surfaces the toggles Excel carries inside `<c:dLbls>`
 (`showValue`, `showCategoryName`, `showSeriesName`, `showPercent`,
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
## Review summary

Reviewed the full diff and traced the read / write / clone bridge end-to-end. The implementation is solid: per-axis spec child order is correctly enforced (`<c:logBase>` before `<c:orientation>`, `<c:max>`/`<c:min>` after; `<c:numFmt>` between `<c:title>` and `<c:crossAx>`; `<c:majorUnit>` / `<c:minorUnit>` after `<c:crossBetween>` per CT_ValAx), the normalization is defensive on both sides (non-finite numbers, zero/negative tick spacings, log bases outside `2..1000`, `min >= max`, and empty `formatCode` all collapse cleanly), and the wholesale-replace contract in `applyScaleOverride` / `applyNumberFormatOverride` matches the existing grammar for axis titles and gridlines.

The `pie` / `doughnut` short-circuit in `resolveAxes` (chart-clone.ts:225) and the `buildPlotArea` pie/doughnut branches (which never call the axis builders) ensure inherited scale / numberFormat cannot poison those families — verified by the tests.

Test coverage looks thorough: 5 + 4 reader tests, 8 + 7 writer tests, 7 + 7 clone tests, plus the round-trip integration test that closes the loop through `parseChart -> cloneChart -> writeXlsx -> parseChart`. Spec-order assertions use index-based comparisons rather than fragile substring sequences, which is the right call.

## Local verification

- `pnpm test` (lint + typecheck + 2642 vitest cases) passes locally
- `pnpm build` succeeds

## Follow-up I added

Pushed a small docs-only commit (`9407a23`) updating the README so the parser example, read-side narrative, writer-side authoring section, and the `cloneChart` override grammar all mention the new `scale` / `numberFormat` fields. No code changes.

## Remaining concerns

None blocking. Two minor observations for future work, not for this PR:

1. On `<c:catAx>` the writer still threads `xScale` through `buildAxisScaling`, which means a caller setting `axes.x.scale.{min,max,logBase}` on a bar/column chart would emit those children inside a category-axis `<c:scaling>`. CT_CatAx allows `<c:scaling>` and Excel tolerates the values, but they're meaningless for textual categories. Not a regression (the previous code also emitted `<c:scaling>` on catAx with just `<c:orientation>`), and harmless in practice — happy to leave as-is unless you'd rather drop scale silently for the category axis on bar/column.
2. `parseAxisScale` does not enforce the `2..1000` band on `<c:logBase>` (only the writer does). That's a deliberate "surface what Excel emitted" stance and matches the documentation, just flagging for awareness.

LGTM — ready to merge once CI is green on the docs commit.

productdevbook:
Quick follow-up: resolved a fresh README conflict that landed when #201 (line/area grouping) merged into main between my review push and this comment. The conflict was purely in two adjacent paragraphs of the chart docs — kept both: the new `lineGrouping` / `areaGrouping` text from main and the new scale / numberFormat narrative from this branch, plus their corresponding mentions in the `cloneChart` override section. Pushed as a merge commit (235ac18).

Re-ran the suite after the merge: `pnpm test` passes 2669 tests (up from 2642 pre-merge, the delta is the line/area grouping tests #201 added) and `pnpm build` succeeds. Awaiting CI on the merge commit before merging.

PR #203, opened 2026-05-02
feat(xlsx): chart first slice angle — read, write, clone-through

productdevbook:
## Summary

Pie / doughnut charts use `<c:firstSliceAng>` to rotate the starting wedge clockwise from 12 o'clock — a common ask when composing dashboards whose pie / doughnut charts need to align visually (e.g. start at 3 o'clock = 90 degrees, paired pies that mirror each other at 0 / 180). The reader ignored the element entirely and the doughnut writer hardcoded `0`, so a template's rotation flattened on every `parseChart` → `cloneChart` → `writeXlsx` round-trip and could not be authored from scratch.

This PR closes that gap at all three layers — read, write, and clone — so a template's pie / doughnut rotation survives the full retarget loop and can be set explicitly on a fresh chart, completing one more piece of the dashboard-composition flow from #136.

## API

```ts

// ── Read side ──
console.log(source.firstSliceAng); // 90  (template starts at 3 o'clock)

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "doughnut",
      series: [{ values: "B2:B6", categories: "A2:A6" }],
      firstSliceAng: 90,   // start at 3 o'clock
      holeSize: 60,
    }],
  }],
});

// ── Clone-through ──
  // firstSliceAng inherits from `source` automatically; pass an
  // override (or coerce to a non-pie/doughnut type) to drop it.
});

// Flatten doughnut → pie keeps the rotation:
const flat = cloneChart(source, {
  type: "pie",
});
console.log(flat.firstSliceAng); // 90  (rotation survives the flatten)
```

## Model

```ts
interface SheetChart {
  /** Pie / doughnut starting angle in degrees, 0..360. Default 0. */
  firstSliceAng?: number;
}

interface Chart {
  /** Surfaced from <c:pieChart>/<c:doughnutChart><c:firstSliceAng val=".."/>. */
  firstSliceAng?: number;
}

interface CloneChartOptions {
  /** Override SheetChart.firstSliceAng. */
  firstSliceAng?: number;
}
```

## Implementation

- **Reader (`chart-reader.ts`)**: `parseFirstSliceAng` walks `<c:pieChart>` / `<c:pie3DChart>` / `<c:doughnutChart>` / `<c:ofPieChart>` for `<c:firstSliceAng val=".."/>` and parses the integer attribute. Values outside CT_FirstSliceAng's `0..360` band are dropped (rather than clamped) so a corrupt template does not silently rewrite as a different angle. `0` and the schema-equivalent `360` collapse to `undefined` so absence and the default round-trip identically — symmetric with how `holeSize` / `barGrouping` collapse OOXML defaults to `undefined`.
- **Writer (`chart-writer.ts`)**: a new `clampFirstSliceAng` helper rounds non-integer input, wraps out-of-band values modulo 360 (`380` → `20`, `-90` → `270` — what Excel's chart-formatting pane does when a user types an out-of-band value), and returns `undefined` for the default `0` so the pie writer can elide the element. The doughnut writer always emits `<c:firstSliceAng>` (matching Excel's reference serialization) and substitutes `0` when the helper returns `undefined`. `<c:firstSliceAng>` lands after `<c:dLbls>` and before `<c:holeSize>` — the order CT_PieChart / CT_DoughnutChart enforce.
- **Clone bridge (`chart-clone.ts`)**: the rotation carries onto the clone when the resolved target is `pie` or `doughnut` (including a doughnut → pie flatten — the element lives on both kinds, so the rotation survives). Coercion into a non-pie family drops the inherited value silently so it never leaks into a chart kind that has no rotation knob. Override grammar mirrors `holeSize`: an explicit `options.firstSliceAng` wins over the source's parsed value.

## Edge cases

- `<c:firstSliceAng val="0"/>` and `<c:firstSliceAng val="360"/>` both collapse to `undefined` on read.
- Negative or `>360` values are dropped on read (out of CT_FirstSliceAng band).
- Out-of-band values on write wrap modulo 360 (`380` → `20`, `-90` → `270`, `360` → `0` → omitted on pie / `0` on doughnut).
- `NaN` / `Infinity` collapse to the default (omitted on pie, `0` on doughnut).
- Non-integer input rounds (`47.6` → `48`).
- Pie with no rotation set elides `<c:firstSliceAng>`; doughnut always emits it (`val="0"` is the Excel reference baseline).
- Cloning a doughnut template with rotation onto a pie target keeps the angle (`type: "pie"` on a `firstSliceAng: 90` source → `firstSliceAng: 90`).
- Cloning into a non-pie / non-doughnut family (column, line, scatter, area) drops the inherited rotation.
- Setting `firstSliceAng` on a non-pie / non-doughnut `SheetChart` is silently ignored by the writer.
- Spec-enforced child order: `<c:firstSliceAng>` lands after `<c:dLbls>` and before `<c:holeSize>` on doughnut.

Refs #136

## Test plan

- [x] `pnpm typecheck`
- [x] `pnpm vitest run` (chart suite — all 2276 tests pass; adds 8 parser tests, 11 writer tests, 6 clone tests, plus 1 end-to-end round-trip via `parseChart` → `cloneChart` → `writeXlsx` → `parseChart`)
- [x] `pnpm build`
- [x] Round-trip: pinned `firstSliceAng: 135` survives the full retarget loop verbatim
- [x] Doughnut → pie flatten preserves the inherited rotation
- [x] Pie / doughnut silently drop the field for non-pie / non-doughnut clones

### DIFF
--- a/README.md
+++ b/README.md
@@ -610,6 +610,13 @@ plus `position` and `separator`). Series-level overrides land on
 on doughnut charts so a parsed template can round-trip its hole back
 through `cloneChart`; non-doughnut charts (and doughnut charts that
 omit the element) never report it.
+`Chart.firstSliceAng` surfaces the
+`<c:pieChart><c:firstSliceAng val=".."/>` /
+`<c:doughnutChart><c:firstSliceAng val=".."/>` rotation (degrees
+clockwise from 12 o'clock) on pie and doughnut charts. The OOXML
+default `0` (and the schema-equivalent `360`) collapses to
+`undefined` so absence and the default round-trip identically;
+non-pie / non-doughnut charts never report it.
 Sheets that hucre actively regenerates because they
 also carry hucre-managed images currently keep the chart bodies but
 lose the in-drawing chart anchor — merging hucre's drawing output
@@ -678,6 +685,11 @@ even when the chart-level default has them on). For doughnut charts,
 `holeSize` (10 – 90, Excel's UI band; default 50) controls the
 diameter of the inner hole — values outside the band are clamped to
 the closest end and non-doughnut kinds silently ignore the field.
+For pie and doughnut charts, `firstSliceAng` (0 – 360 degrees, default
+0 = 12 o'clock) rotates the first wedge clockwise — useful for
+aligning paired charts in a dashboard. Out-of-band values wrap modulo
+360 (380 → 20, -90 → 270) the same way Excel's chart-formatting pane
+does, and non-pie / non-doughnut kinds silently ignore the field.
 Radar, stock, 3D variants, trendlines, and combo charts are out of
 scope today.
 
@@ -727,7 +739,11 @@ into a column drops the inherited grouping rather than silently
 emitting a value the writer would ignore. Doughnut clones also inherit
 the parsed `holeSize` from the template; pass `holeSize: 60` to
 override or `type: "pie"` to flatten into a plain pie (the hole hint
-is dropped silently in that case). Data labels inherit too: omit `dataLabels`
+is dropped silently in that case). Pie and doughnut clones inherit the
+parsed `firstSliceAng` from the template; the rotation also survives a
+`type: "pie"` flattening of a doughnut source (the element lives on
+both pie and doughnut), but is dropped when the resolved clone target
+is anything else. Data labels inherit too: omit `dataLabels`
 to carry the source's chart-level labels through, pass an object to
 replace, or `null` to drop them; per-series overrides accept the same
 `undefined`/`null`/object grammar plus `false` to suppress labels on
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -687,6 +687,18 @@ export interface SheetChart {
    */
   holeSize?: number;
   /**
+   * Pie / doughnut starting angle in degrees, measured clockwise from
+   * the 12 o'clock position. Accepted range: 0 – 360 (the OOXML schema
+   * range). Default: `0` — the Excel default (first slice begins at
+   * 12 o'clock). Maps to `<c:firstSliceAng val=".."/>`. Ignored for
+   * non-pie / non-doughnut chart kinds.
+   *
+   * Useful for rotating the first wedge into a specific quadrant when
+   * composing a dashboard whose pie / doughnut charts should align
+   * visually (e.g. `90` to start at 3 o'clock).
+   */
+  firstSliceAng?: number;
+  /**
    * Whether the legend is shown and where. Default: `"right"` for
    * pie/doughnut/bar/line/area, `"bottom"` for scatter. Pass `false`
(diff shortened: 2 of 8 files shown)


productdevbook (commented):
Reviewed the read/write/clone-through implementation end-to-end.

## What I checked
- **Reader** (`parseFirstSliceAng`): correctly walks pie/pie3D/doughnut/ofPie kinds, integer-parses `val`, drops out-of-band (negative or >360), and collapses both `0` and `360` to `undefined` so absence and the OOXML default round-trip identically. Mirrors how `parseHoleSize` behaves — good consistency.
- **Writer** (`clampFirstSliceAng`): rounds non-integer input, wraps modulo 360 (`-90` -> `270`, `380` -> `20`), and returns `undefined` for `0` so the pie writer can elide the element while the doughnut writer falls back to `0` (matching Excel's reference serialization). The pie/doughnut child order (`varyColors` -> `ser` -> `dLbls` -> `firstSliceAng` -> `holeSize`) matches CT_PieChart / CT_DoughnutChart.
- **Clone bridge**: rotation carries when the resolved type is `pie` or `doughnut` (including doughnut -> pie flatten, since the element lives on both), and is silently dropped on coercion to other families. Override grammar mirrors `holeSize`.
- **Tests**: 8 reader cases + 11 writer cases + 6 clone cases + 1 end-to-end round-trip — covers defaults, malformed input, negative/>360, NaN/Infinity, non-integer rounding, the schema-equivalent `360`, non-pie/doughnut suppression, and the doughnut OOXML child-order constraint. Looks thorough.

## What I did
- Merged `main` into the branch to resolve the conflict introduced by #202 (chart axis scale and number format). Conflicts were in:
  - `README.md` — combined both feature paragraphs in the cloneChart docs.
  - `test/chart-clone.test.ts` — split the shared integration test into two distinct round-trip tests, one for axis scale/numberFormat and one for `firstSliceAng`, so both features keep their end-to-end coverage.
  - Auto-merged `_types.ts`, `chart-clone.ts`, `chart-reader.ts`, `chart-writer.ts`, `charts-write.test.ts`, `charts.test.ts`.
- Re-ran `pnpm test` (lint + typecheck + vitest) and `pnpm build` after the merge — all 2695 tests pass, build succeeds.

## Minor nit (non-blocking)
The comment in `clampFirstSliceAng` says "The OOXML schema actually allows 360 as a value, so we keep it distinct from 0", but the implementation collapses `360 % 360 === 0` to `undefined` — same as `0`. This actually matches the documented intent and the reader's symmetric collapse of `360 -> undefined`, so the behavior is correct; only the inline comment is slightly misleading. Not worth blocking the PR over — feel free to tighten the comment in a follow-up if you like.

LGTM once CI passes.

PR #204, opened 2026-05-02
feat(xlsx): bar/column gap width and overlap — read, write, clone-through

productdevbook:
## Summary

Bar / column charts use `<c:gapWidth>` (0 – 500 % of the bar width) to control the empty space between category groups and `<c:overlap>` (-100..100 %) to control how series within a group separate or stack — both are common asks when a dashboard's bar charts need to align visually (e.g. tighter `gapWidth: 50` for compact group spacing, slight negative `overlap: -10` to add breathing room between series). The reader ignored both elements entirely and the writer hardcoded `gapWidth=150` for clustered and `overlap=100` for stacked, so a template's spacing flattened on every `parseChart` → `cloneChart` → `writeXlsx` round-trip and could not be authored from scratch.

This PR closes that gap at all three layers — read, write, and clone — so a template's bar spacing survives the full retarget loop and can be set explicitly on a fresh chart, completing one more piece of the dashboard-composition flow from #136.

## API

```ts

// ── Read side ──
console.log(source.gapWidth); // 50  (template's tight group spacing)
console.log(source.overlap);  // -25 (slight separation between series)

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      series: [
        { values: "B2:B6", categories: "A2:A6" },
        { values: "C2:C6", categories: "A2:A6" },
      ],
      gapWidth: 50,   // half-width gaps between groups
      overlap: -25,   // 25% separation between series within a group
    }],
  }],
});

// ── Clone-through ──
  // gapWidth / overlap inherit from `source` automatically; pass an
  // override (or coerce to a non-bar/column type) to drop them.
});
```

## Model

```ts
interface SheetChart {
  /** Bar/column gap width as % of bar width, 0..500. Default 150 (clustered), omitted (stacked). */
  gapWidth?: number;
  /** Bar/column series overlap as % of bar width, -100..100. Default 0 (clustered), 100 (stacked). */
  overlap?: number;
}

interface Chart {
  /** Surfaced from <c:barChart>/<c:bar3DChart><c:gapWidth val="..."/>. */
  gapWidth?: number;
  /** Surfaced from <c:barChart>/<c:bar3DChart><c:overlap val="..."/>. */
  overlap?: number;
}

interface CloneChartOptions {
  /** Override SheetChart.gapWidth (bar/column only). */
  gapWidth?: number;
  /** Override SheetChart.overlap (bar/column only). */
  overlap?: number;
}
```

## Implementation

- **Reader (`chart-reader.ts`)**: `parseGapWidth` and `parseOverlap` walk `<c:barChart>` / `<c:bar3DChart>` for the matching `val=".."` attribute. Out-of-band values (`<0` / `>500` for `gapWidth`, `<-100` / `>100` for `overlap`) are dropped (rather than clamped) so a corrupt template does not silently rewrite as different spacing. The OOXML defaults (`gapWidth=150`, `overlap=0`) collapse to `undefined` so absence and the default round-trip identically — symmetric with how `holeSize` / `barGrouping` collapse OOXML defaults to `undefined`.
- **Writer (`chart-writer.ts`)**: new `clampGapWidth` / `clampOverlap` helpers round non-integer input and clamp out-of-band values to the schema bounds (`gapWidth: 999` → `500`, `overlap: -250` → `-100`). Existing per-grouping defaults are preserved exactly — clustered still emits `gapWidth=150` only, stacked still emits `overlap=100` only, so untouched charts stay byte-identical with what Excel writes. An explicit `chart.gapWidth` / `chart.overlap` always emits the matching element (even on a stacked chart, or with a value that happens to equal the per-grouping default), so callers can pin both knobs together. `<c:gapWidth>` lands before `<c:overlap>` per the OOXML CT_BarChart child sequence (`barDir → grouping → varyColors → ser* → dLbls? → gapWidth? → overlap? → serLines* → axId+`).
- **Clone bridge (`chart-clone.ts`)**: both fields carry onto the clone when the resolved target is `bar` or `column`. Coercion into a non-bar family (line, pie, area, scatter, doughnut) drops the inherited values silently so they never leak into a chart kind that has no `<c:barChart>` element. Override grammar mirrors `holeSize` / `firstSliceAng`: an explicit `options.gapWidth` / `options.overlap` wins over the source's parsed value.

## Edge cases

- `<c:gapWidth val="150"/>` and `<c:overlap val="0"/>` both collapse to `undefined` on read (OOXML defaults).
- Out-of-band values dropped on read (`gapWidth: -1` / `501`, `overlap: -101` / `101` → `undefined`).
- Out-of-band values clamp on write (`gapWidth: 999` → `500`, `overlap: -250` → `-100`).
- `NaN` / `Infinity` collapse to the per-grouping default (clustered: `gapWidth=150`, no overlap; stacked: no gapWidth, `overlap=100`).
- Non-integer input rounds (`175.6` → `176`, `-33.4` → `-33`).
- Setting `gapWidth` on a stacked chart forces emission alongside the default `overlap=100`.
- Setting `overlap` on a clustered chart forces emission alongside the default `gapWidth=150`.
- Cloning a column template with spacing onto a line target drops both fields (no `<c:barChart>` host).
- Setting `gapWidth` / `overlap` on a non-bar `SheetChart` is silently ignored by the writer.
- Spec-enforced child order: `<c:gapWidth>` lands before `<c:overlap>` inside `<c:barChart>`, both before `<c:axId>`.
- `<c:bar3DChart>` is also recognized on the read side (writer never emits 3D variants).

Refs #136

## Test plan

- [x] `pnpm typecheck`
- [x] `pnpm vitest run` (chart suite — adds 39 tests across reader / writer / clone, full suite passes 2705/2705)
- [x] `pnpm build`
- [x] Round-trip: pinned `gapWidth: 50, overlap: -25` survives the full retarget loop verbatim
- [x] Bar / column silently drop both fields for non-bar / non-column clones
- [x] Existing default behavior preserved (clustered → only `gapWidth=150`, stacked → only `overlap=100`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -626,6 +626,13 @@ plus `position` and `separator`). Series-level overrides land on
 on doughnut charts so a parsed template can round-trip its hole back
 through `cloneChart`; non-doughnut charts (and doughnut charts that
 omit the element) never report it.
+`Chart.gapWidth` and `Chart.overlap` surface the bar / column
+spacing knobs (`<c:barChart><c:gapWidth val=".."/>` and
+`<c:overlap val=".."/>`). `gapWidth` (0 – 500 % of the bar width)
+controls the gap between category groups and `overlap` (-100..100 %)
+controls how much series within a group separate or stack. The OOXML
+defaults (`gapWidth=150`, `overlap=0`) collapse to `undefined`;
+non-bar / non-column charts never report either.
 Sheets that hucre actively regenerates because they
 also carry hucre-managed images currently keep the chart bodies but
 lose the in-drawing chart anchor — merging hucre's drawing output
@@ -706,6 +713,12 @@ even when the chart-level default has them on). For doughnut charts,
 `holeSize` (10 – 90, Excel's UI band; default 50) controls the
 diameter of the inner hole — values outside the band are clamped to
 the closest end and non-doughnut kinds silently ignore the field.
+For bar and column charts, `gapWidth` (0 – 500 % of the bar width;
+default `150` for unstacked, omitted for stacked) controls the empty
+space between adjacent category groups and `overlap` (-100..100 %;
+default `0` for clustered, `100` for stacked) controls how series
+within a group separate or stack. Out-of-band values clamp to the
+schema bounds, and non-bar / non-column kinds silently ignore both.
 Radar, stock, 3D variants, trendlines, and combo charts are out of
 scope today.
 
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -666,6 +666,27 @@ export interface SheetChart {
    */
   barGrouping?: "clustered" | "stacked" | "percentStacked";
   /**
+   * Bar/column gap width as a percentage of the bar width — the empty
+   * space between adjacent category groups. Accepted range: `0` – `500`
+   * (the OOXML `ST_GapAmount` schema). Excel's default is `150` (each
+   * group's gap equals 1.5× the bar width). Smaller values pack groups
+   * tighter; `0` removes the gap entirely. Maps to
+   * `<c:barChart><c:gapWidth val=".."/></c:barChart>`. Ignored for
+   * non-bar / non-column chart kinds.
+   */
+  gapWidth?: number;
+  /**
+   * Bar/column series overlap as a percentage of the bar width.
+   * Accepted range: `-100` – `100` (the OOXML `ST_Overlap` schema).
+   * Negative values open a gap between series within a group, positive
+   * values stack them on top of each other. Excel's default is `0` for
+   * `clustered` (side-by-side) and `100` for `stacked` /
+   * `percentStacked` (fully overlapped). Maps to
+   * `<c:barChart><c:overlap val=".."/></c:barChart>`. Ignored for
+   * non-bar / non-column chart kinds.
+   */
+  overlap?: number;
+  /**
    * Line subtype. Default: `"standard"`. `"stacked"` accumulates
    * series end-to-end, `"percentStacked"` normalizes each category to
    * 100%. Ignored for non-line chart kinds. Maps to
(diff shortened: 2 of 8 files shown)


productdevbook (commented):
## Review

Looked at the diff end-to-end (reader, writer, clone, types, README, and the new tests across `test/charts.test.ts`, `test/charts-write.test.ts`, `test/chart-clone.test.ts`).

**Correctness & design**

- Reader: `parseGapWidth` / `parseOverlap` correctly walk `<c:barChart>` / `<c:bar3DChart>` and drop out-of-band values rather than clamp them. The `150 → undefined` and `0 → undefined` collapses match how `holeSize` / `barGrouping` already behave, so absence and the schema default round-trip identically.
- Writer: `clampGapWidth` / `clampOverlap` round non-finite or non-integer input and clamp to schema bounds (no wrap-around — these are physical bar-width percentages, agreed). The OOXML CT_BarChart child order is preserved (`barDir → grouping → varyColors → ser* → dLbls? → gapWidth? → overlap? → axId+`), and the per-grouping defaults stay byte-identical with what Excel writes (clustered: only `gapWidth=150`; stacked / percentStacked: only `overlap=100`). Forcing emission when the value happens to equal the per-grouping default is the right call so callers can pin both knobs on a stacked chart.
- Clone bridge: gates `gapWidth` / `overlap` on `type === "bar" || "column"`, so flattening a column template into a line / pie clone drops the inherited values cleanly. Override grammar mirrors `holeSize` and `firstSliceAng`.

**Test coverage**

39 added tests cover: read happy-paths and OOXML defaults collapsing to undefined, schema bounds (0..500 / -100..100, inclusive), malformed values, `<c:bar3DChart>` symmetry, non-bar kinds being ignored on read; write defaults preserved, explicit emission, clamping, rounding, NaN / Infinity falling through to per-grouping defaults, child order; clone inheritance, override precedence, type-coercion drops. The end-to-end `parseChart → cloneChart → writeXlsx → parseChart` round-trip with `gapWidth=50, overlap=-25` is the right shape.

**Fixes I applied**

The branch was conflicting against `main` after #203 (firstSliceAng) landed. I merged `main` in and resolved overlapping touch-points in `src/xlsx/chart-reader.ts` (locals + result-attach order), `src/_types.ts` (`Chart` interface), `README.md` (two adjacent paragraphs), and the test files (kept both sets of describe blocks + both round-trip integration tests as separate `it`s). The two features are independent — no shared helpers — so the resolution was purely additive. Then I ran `pnpm test` (lint + typecheck + vitest: 88 files / 2731 tests pass) and `pnpm build` (clean) and pushed `6224b4f`.

**Remaining concerns**

None. CI was still running when I posted; happy to merge once it goes green.

PR #205, opened 2026-05-02
feat(xlsx): chart series smooth flag — read, write, clone-through

productdevbook:
## Summary

Line and scatter charts use `<c:smooth>` to render the curved Catmull-Rom-style spline Excel exposes under "Format Data Series → Line → Smoothed line" — a common ask when a dashboard's line charts need to flow through their data points instead of drawing straight segments. The reader ignored the element entirely and the line writer hardcoded `<c:smooth val="0">` for every series, so a template's curved-line state flattened on every `parseChart` → `cloneChart` → `writeXlsx` round-trip and could not be authored from scratch.

This PR closes that gap at all three layers — read, write, and clone — so a template's smoothed lines survive the full retarget loop and can be set explicitly per series on a fresh chart, completing one more piece of the dashboard-composition flow from #136.

## API

```ts

// ── Read side ──
console.log(source.series?.[0].smooth); // true (template's curved trend line)

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "line",
      series: [
        { name: "Curved",   values: "B2:B6", categories: "A2:A6", smooth: true },
        { name: "Straight", values: "C2:C6", categories: "A2:A6" },
      ],
    }],
  }],
});

// ── Clone-through ──
  seriesOverrides: [
    { values: "Dashboard!$B$2:$B$13" }, // smooth inherits from source
    { values: "Dashboard!$C$2:$C$13", smooth: true },  // upgrade to curved
    { values: "Dashboard!$D$2:$D$13", smooth: null },  // drop inherited curve
  ],
});
```

## Model

```ts
interface ChartSeries {
  /** Smooth the line connecting data points (line/scatter only). */
  smooth?: boolean;
}

interface ChartSeriesInfo {
  /** Surfaced from <c:ser><c:smooth val=".."/>; only on line/scatter. */
  smooth?: boolean;
}

interface CloneChartSeriesOverride {
  /** undefined = inherit, null = drop, boolean = replace. */
  smooth?: boolean | null;
}
```

## Implementation

- **Reader (`chart-reader.ts`)**: `parseSmooth` walks `<c:ser>` for the `<c:smooth val=".."/>` child and returns `true` when it resolves to the OOXML truthy spelling (`"1"` / `"true"`). The OOXML default `false` collapses to `undefined` so absence and `<c:smooth val="0"/>` round-trip identically — symmetric with how `holeSize` / `firstSliceAng` collapse OOXML defaults to `undefined`. Surfaced only for `line` / `line3D` / `scatter` kinds because the OOXML schema places `<c:smooth>` exclusively on `CT_LineSer` and `CT_ScatterSer`. A corrupt template carrying `<c:smooth>` on a bar / pie / area series is dropped silently rather than surfacing a flag the writer would never emit anyway.
- **Writer (`chart-writer.ts`)**: the line writer (which must emit `<c:smooth>` per `CT_LineSer`) now threads each series' `smooth` flag through to the per-series builder — straight by default (`val="0"`), curved when the caller pinned `smooth: true`. The scatter writer only emits the optional `<c:smooth>` element when the flag is explicitly `true`, falling back to the omit-by-default shape Excel writes for straight scatter series. All other chart families ignore the flag silently — bar / column / pie / doughnut / area never touched the smooth path.
- **Clone bridge (`chart-clone.ts`)**: `resolveSmooth` mirrors the existing `dataLabels` override grammar — `undefined` inherits the source series' smooth, `null` drops the inherited flag, a boolean replaces it. `false` collapses to `undefined` for symmetry with the writer's elision logic. After per-series merging, a post-build sweep strips every series' `smooth` field when the resolved chart type is anything other than `line` or `scatter` so a doughnut → column flatten (or any other coercion) cannot leak the element into a chart kind whose schema rejects it.

## Edge cases

- `<c:smooth val="0"/>` and absence both collapse to `undefined` on read.
- `<c:smooth val="true"/>` reads as `true` (OOXML allows the boolean spelling).
- Invalid `val` attributes drop to `undefined` rather than fabricating a flag.
- Non-line / non-scatter chart families ignore the flag at all three layers — read drops it, write skips it, clone strips it.
- Multi-series line charts emit `<c:smooth>` per series in declaration order so each series can carry an independent smooth state.
- Setting `smooth: false` on a `ChartSeries` collapses to the same shape as omitting it (writer emits `val="0"` for line, omits the element for scatter).
- Cloning a curved line template into a column / bar / pie / doughnut / area target drops the inherited flag (no `<c:smooth>` host).
- Spec-enforced child order: `<c:smooth>` lands at the tail of `<c:ser>`, after `<c:val>` / `<c:yVal>`.
- `seriesOverrides[i].smooth = false` (vs `null`) collapses to `undefined` so absence / `false` / `null` round-trip identically.

Refs #136

## Test plan

- [x] `pnpm typecheck`
- [x] `pnpm vitest run` (chart suite — adds 27 tests across reader / writer / clone, full suite passes 2755/2755)
- [x] `pnpm build`
- [x] Round-trip: pinned `smooth: true` survives the full `parseChart` → `cloneChart` → `writeXlsx` → `parseChart` retarget loop verbatim
- [x] Multi-series line charts emit smooth independently per series
- [x] Non-line / non-scatter clones silently drop the inherited flag

### DIFF
--- a/README.md
+++ b/README.md
@@ -640,6 +640,13 @@ clockwise from 12 o'clock) on pie and doughnut charts. The OOXML
 default `0` (and the schema-equivalent `360`) collapses to
 `undefined` so absence and the default round-trip identically;
 non-pie / non-doughnut charts never report it.
+`ChartSeriesInfo.smooth` surfaces the per-series
+`<c:ser><c:smooth val=".."/>` flag — Excel's "Format Data Series →
+Line → Smoothed line" toggle — only on `line` / `line3D` / `scatter`
+series (the OOXML schema places `<c:smooth>` exclusively on
+`CT_LineSer` and `CT_ScatterSer`). Absence and the OOXML default
+`val="0"` both collapse to `undefined`, so only an explicit
+`<c:smooth val="1"/>` round-trips as `smooth: true`.
 Sheets that hucre actively regenerates because they
 also carry hucre-managed images currently keep the chart bodies but
 lose the in-drawing chart anchor — merging hucre's drawing output
@@ -731,6 +738,13 @@ For pie and doughnut charts, `firstSliceAng` (0 – 360 degrees, default
 aligning paired charts in a dashboard. Out-of-band values wrap modulo
 360 (380 → 20, -90 → 270) the same way Excel's chart-formatting pane
 does, and non-pie / non-doughnut kinds silently ignore the field.
+For line and scatter charts, each `series[i].smooth` flag toggles
+Excel's curved-line variant (`<c:smooth val="..">` inside `<c:ser>`).
+Line series always emit the element — `smooth: true` writes `val="1"`,
+absence or `false` writes the OOXML default `val="0"`; scatter series
+only emit `<c:smooth val="1"/>` when `smooth` is explicitly `true` so
+untouched scatter charts stay byte-clean. Bar / column / pie /
+doughnut / area kinds silently ignore the flag.
 Radar, stock, 3D variants, trendlines, and combo charts are out of
 scope today.
 
@@ -795,7 +809,11 @@ too: omit `dataLabels`
 to carry the source's chart-level labels through, pass an object to
 replace, or `null` to drop them; per-series overrides accept the same
 `undefined`/`null`/object grammar plus `false` to suppress labels on
-a single series.
+a single series. Per-series smooth-line state inherits from the
+template by default; `seriesOverrides[i].smooth` accepts the same
+`undefined` (inherit) / `null` (drop) / `boolean` (replace) grammar,
+and the inherited flag is dropped automatically when the resolved
+clone target is anything other than `line` or `scatter`.
 
 #### Walking and adding charts with `getCharts` / `addChart`
 
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -633,6 +633,18 @@ export interface ChartSeries {
    * {@link SheetChart.dataLabels} enables them.
    */
   dataLabels?: ChartDataLabels | false;
+  /**
+   * Smooth the line connecting data points using a Catmull-Rom-style
+   * spline. Maps to `<c:smooth val="..">` inside the `<c:ser>` element.
+   * Only meaningful for `line` and `scatter` charts — ignored for every
+   * other chart kind (the OOXML schema does not allow `<c:smooth>` on
+   * bar / column / pie / doughnut / area series).
+   *
+   * Default: `false` (straight-line segments). Set `true` to render the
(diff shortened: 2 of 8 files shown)


productdevbook (commented):
Reviewed at the read / write / clone layers — the implementation is clean and consistent with the conventions established by the adjacent `firstSliceAng`, `gapWidth`, and `dataLabels` features.

What I verified:
- **Reader** (`parseSmooth`) correctly limits surfacing to `line` / `line3D` / `scatter` per the OOXML schema (`<c:smooth>` is only valid on `CT_LineSer` / `CT_ScatterSer`). Re-using the existing `readBoolAttr` helper means `"1"` / `"true"` both parse, and `"0"` / `"false"` / absence collapse to `undefined` — symmetric with how `holeSize` and `firstSliceAng` collapse OOXML defaults.
- **Writer** matches the schema's optionality split: `<c:smooth>` is required on `CT_LineSer` so the line writer always emits `val="0"` / `val="1"`, while it's optional on `CT_ScatterSer` and only gets emitted when `smooth: true`. The element is appended after `<c:val>` / `<c:yVal>`, which respects the CT_LineSer / CT_ScatterSer child order. Other chart writers (bar, column, pie, doughnut, area) never thread `smooth` into `buildSeries`, so the flag cannot leak into chart kinds whose schema rejects it.
- **Clone bridge** mirrors the `dataLabels` per-series override grammar (`undefined` inherits, `null` drops, boolean replaces, `false` collapses to `undefined` for symmetry). The post-build sweep that strips `smooth` when the resolved type is not `line` / `scatter` is good defense-in-depth — it guards both the `seriesOverrides` path *and* the explicit `options.series` path (covered by the dedicated test). `line3D` source kinds correctly flow through `chartKindToWriteKind` to a `line` clone, so a `line3D` template's smoothed state survives the round-trip.
- **Tests** (27 new) cover the OOXML default collapse, the boolean spelling, the non-line/scatter ignore-at-each-layer, multi-series independence, the OOXML child order, and a full `parseChart` → `cloneChart` → `writeXlsx` → `parseChart` round-trip with mixed series.

CI is green (build passing), `pnpm lint` / `pnpm typecheck` / `pnpm vitest run` (2755/2755) and `pnpm build` all pass locally on the branch.

Pushed one follow-up commit to the same branch: a README update documenting the new flag at the read, write, and clone layers in the same style as the adjacent `firstSliceAng` / `gapWidth` / `dataLabels` paragraphs. No code changes needed.

Will merge once the CI run on the doc commit reports green.

PR #206, opened 2026-05-02
feat(xlsx): chart series markers — read, write, clone-through

productdevbook:
## Summary

Surfaces the per-series `<c:marker>` block at the read, write, and clone layers so a template's marker styling (symbol shape, point size, fill / outline colors) survives `parseChart` → `cloneChart` → `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:marker>` is the OOXML element that controls the per-data-point glyph rendered by line and scatter charts — Excel exposes the same knobs under "Format Data Series → Marker Options". Until now hucre's writer silently emitted the chart-type-level `<c:marker val="1"/>` toggle on line charts (so markers showed by default) but had no way to pin a specific shape, size, or color per series, and `parseChart` did not surface the parsed configuration. This bridges the last per-series styling gap for the dashboard composition flow tracked in #136.

## API

```ts

// ── Read side ──
console.log(source.series?.[0].marker);
// { symbol: "diamond", size: 10, fill: "1F77B4", line: "0F3F60" }

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "line",
      series: [
        { name: "Revenue", values: "B2:B6", categories: "A2:A6",
          marker: { symbol: "circle", size: 8, fill: "1F77B4" } },
        { name: "Cost",    values: "C2:C6", categories: "A2:A6",
          marker: { symbol: "square", size: 6, fill: "FF7F0E" } },
      ],
    }],
  }],
});

// ── Clone-through ──
  seriesOverrides: [
    { values: "Dashboard!$B$2:$B$13" },                       // marker inherits
    { values: "Dashboard!$C$2:$C$13", marker: { symbol: "x" } },  // replace wholesale
    { values: "Dashboard!$D$2:$D$13", marker: null },         // drop inherited block
  ],
});
```

## Model

```ts
type ChartMarkerSymbol =
  | "none" | "auto"
  | "circle" | "square" | "diamond" | "triangle"
  | "x" | "star" | "dot" | "dash" | "plus";

interface ChartMarker {
  /** Marker glyph shape. */
  symbol?: ChartMarkerSymbol;
  /** Marker glyph size in points (clamped to OOXML 2..72). */
  size?: number;
  /** Marker fill color (6-digit RGB hex, e.g. "1F77B4"). */
  fill?: string;
  /** Marker outline color (6-digit RGB hex). */
  line?: string;
}

interface ChartSeries {
  /** ...existing fields... */
  marker?: ChartMarker;
}
```

The read-side `ChartSeriesInfo.marker` mirrors the same shape so a parsed marker slots straight back into `cloneChart` without transformation.

## Behavior

- **Read** — `parseChart` pulls `<c:symbol>`, `<c:size>`, and the `<c:spPr>` fill / outline colors off `<c:marker>` only on `line` / `line3D` / `scatter` series. Empty `<c:marker/>` blocks collapse to `undefined`, unknown symbol values are dropped, sizes outside `2..72` are clamped, and malformed hex values are filtered out.
- **Write** — `<c:marker>` is emitted between `<c:spPr>` and `<c:dLbls>` per the OOXML series schema (CT_LineSer / CT_ScatterSer). The chart-type-level `<c:marker val="1"/>` toggle on line charts is preserved (Excel's per-series default), and the writer silently drops the field on every non-line/non-scatter family.
- **Clone** — `seriesOverrides[i].marker` accepts the standard `undefined` (inherit) / `null` (drop) / object (replace wholesale, no per-field merge) grammar that mirrors the smooth, dataLabels, and axis-scale resolvers shipped earlier. The inherited block is also dropped automatically when the resolved clone target is anything other than `line` or `scatter`.

Closes the marker piece of #136.

Refs #136

## Test plan

- [x] `pnpm test` (lint + typecheck + 67703 vitest tests) — all green
- [x] `pnpm build` — clean obuild
- [x] New tests cover: parse, parse-collapse-empty, clamp / drop invalid, multi-series independence, writer per-series emission, OOXML element ordering, kind-restriction, clone inherit / replace / null / drop on type-flatten, end-to-end `parseChart → cloneChart → writeXlsx → parseChart` round-trip preserving the full marker block

### DIFF
--- a/README.md
+++ b/README.md
@@ -647,6 +647,15 @@ series (the OOXML schema places `<c:smooth>` exclusively on
 `CT_LineSer` and `CT_ScatterSer`). Absence and the OOXML default
 `val="0"` both collapse to `undefined`, so only an explicit
 `<c:smooth val="1"/>` round-trips as `smooth: true`.
+`ChartSeriesInfo.marker` surfaces the per-series `<c:ser><c:marker>`
+glyph configuration on `line` / `line3D` / `scatter` series (the
+schema places `<c:marker>` only on `CT_LineSer` and `CT_ScatterSer`).
+The reader pulls `symbol` (`circle` / `square` / `diamond` /
+`triangle` / `x` / `star` / `dot` / `dash` / `plus` / `auto` /
+`none`), `size` (clamped to the OOXML 2..72 band), and the marker's
+fill / outline colors out of `<c:spPr><a:solidFill>` and
+`<c:spPr><a:ln><a:solidFill>` — empty `<c:marker/>` elements collapse
+to `undefined` so absence and a bare element round-trip identically.
 Sheets that hucre actively regenerates because they
 also carry hucre-managed images currently keep the chart bodies but
 lose the in-drawing chart anchor — merging hucre's drawing output
@@ -745,6 +754,16 @@ absence or `false` writes the OOXML default `val="0"`; scatter series
 only emit `<c:smooth val="1"/>` when `smooth` is explicitly `true` so
 untouched scatter charts stay byte-clean. Bar / column / pie /
 doughnut / area kinds silently ignore the flag.
+For line and scatter charts, each `series[i].marker` block also
+controls the per-point glyph: `symbol`
+(`circle` / `square` / `diamond` / `triangle` / `x` / `star` / `dot` /
+`dash` / `plus` / `auto` / `none`) picks the shape, `size` (2 – 72)
+sets the diameter, and `fill` / `line` (6-digit RGB hex) tint the
+glyph fill and outline. Out-of-range sizes clamp to the schema band,
+unknown symbols and malformed hex values are dropped so Excel never
+receives an attribute it would reject, and an empty marker (`{}`)
+collapses to no `<c:marker>` at all. Bar / column / pie / doughnut /
+area kinds silently ignore the field.
 Radar, stock, 3D variants, trendlines, and combo charts are out of
 scope today.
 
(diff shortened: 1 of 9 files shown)


productdevbook (commented):
Reviewed the diff end-to-end and the implementation is clean and consistent with the existing smooth/dataLabels patterns:

- Type model (`ChartMarker` / `ChartMarkerSymbol`) is well-documented and exported alongside other public chart types in `src/index.ts`.
- Reader correctly limits `<c:marker>` parsing to `line` / `line3D` / `scatter`, clamps size to the OOXML 2..72 band, drops unknown symbols, normalizes hex to uppercase, and collapses empty marker blocks to `undefined`.
- Writer slots `<c:marker>` between `<c:spPr>` and `<c:dLbls>` per the CT_LineSer / CT_ScatterSer schema, mirrors the same clamp / normalize / drop-malformed semantics, preserves the chart-type-level `<c:marker val="1"/>` toggle on line charts, and silently ignores the field on bar / column / pie / doughnut / area families.
- Clone resolver implements the `undefined` (inherit) / `null` (drop) / object (replace wholesale) grammar that matches the smooth and dataLabels resolvers, returns a fresh shallow copy so callers cannot mutate the parsed source, and post-strips marker on type-flatten coercions.
- Tests cover parse, parse-collapse-empty, clamp, drop-invalid, multi-series independence, writer per-series emission, OOXML element ordering, kind restriction across all non-line/non-scatter families, every clone-grammar branch, and a full `parseChart → cloneChart → writeXlsx → parseChart` round-trip.
- README updates accurately describe the new field across read, write, and clone sections.

CI is green. Merging.

PR #207, opened 2026-05-02
feat(xlsx): chart series line stroke — read, write, clone-through

productdevbook:
## Summary

Surfaces the per-series `<a:ln>` stroke (preset dash pattern + width in points) at the read, write, and clone layers so a template's line styling survives `parseChart` → `cloneChart` → `writeXlsx` and can be authored from scratch on a fresh chart.

`<a:ln>` is the OOXML element that controls how Excel paints the connecting line on `line` and `scatter` series — Excel exposes the same knobs under "Format Data Series → Line → Dash type / Width". Until now hucre had no way to pin a dash pattern or stroke width per series, and `parseChart` did not surface either field. This bridges another per-series styling gap for the dashboard composition flow tracked in #136.

## API

```ts

// ── Read side ──
console.log(source.series?.[0].stroke);
// { dash: "dashDot", width: 2.5 }

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "line",
      series: [
        { name: "Forecast", values: "B2:B6", categories: "A2:A6",
          stroke: { dash: "dash", width: 1.5 } },
        { name: "Actual",   values: "C2:C6", categories: "A2:A6",
          stroke: { dash: "solid", width: 2.5 } },
      ],
    }],
  }],
});

// ── Clone-through ──
  seriesOverrides: [
    { values: "Dashboard!$B$2:$B$13" },                            // stroke inherits
    { values: "Dashboard!$C$2:$C$13", stroke: { dash: "dot" } },  // replace wholesale
    { values: "Dashboard!$D$2:$D$13", stroke: null },             // drop inherited block
  ],
});
```

## Model

```ts
type ChartLineDashStyle =
  | "solid" | "dot" | "dash" | "lgDash"
  | "dashDot" | "lgDashDot" | "lgDashDotDot"
  | "sysDash" | "sysDot" | "sysDashDot" | "sysDashDotDot";

interface ChartLineStroke {
  /** Preset dash pattern. */
  dash?: ChartLineDashStyle;
  /** Stroke width in points (clamped to Excel's 0.25..13.5 pt band). */
  width?: number;
}

interface ChartSeries {
  /** ...existing fields... */
  stroke?: ChartLineStroke;
}
```

The read-side `ChartSeriesInfo.stroke` mirrors the same shape so a parsed stroke slots straight back into `cloneChart` without transformation.

## Behavior

- **Read** — `parseChart` pulls `<a:prstDash>` and the `w` attribute off `<c:spPr><a:ln>` only on `line` / `line3D` / `scatter` series. Empty `<a:ln/>` blocks collapse to `undefined`, unknown dash values are dropped, widths outside `0.25..13.5 pt` are clamped, and EMU is converted back to points (snapped to the 0.25 pt grid Excel exposes).
- **Write** — Stroke metadata layers onto the existing `<c:spPr>` that already carries `series.color`. A `color + stroke` combo behaves like Excel's UI: the line picks up the fill color, and the dash / width override visibility-only attributes. A colorless series with stroke now emits `<c:spPr><a:ln>` carrying just the dash and width. The writer silently drops the field on every non-line / non-scatter family.
- **Clone** — `seriesOverrides[i].stroke` accepts the standard `undefined` (inherit) / `null` (drop) / object (replace wholesale, no per-field merge) grammar that mirrors the smooth, dataLabels, and axis-scale resolvers shipped earlier. The inherited block is also dropped automatically when the resolved clone target is anything other than `line` or `scatter`.

## Edge cases

- `<a:ln w="31750"/>` reads as `{ width: 2.5 }` (1 pt = 12 700 EMU).
- Width inputs are snapped to the 0.25 pt grid before EMU conversion so a parsed-then-written stroke is byte-stable.
- Width above 13.5 pt or below 0.25 pt clamps on both read and write so Excel's UI band is preserved.
- Non-finite widths (NaN, Infinity) drop to `undefined` rather than emit an invalid `w` attribute.
- An unknown `<a:prstDash val="...">` reads as `undefined` rather than surface a token Excel rejects.
- Empty stroke `{}` collapses to no `<c:spPr>` at write time when the series has no fill color either, so untouched chart XML stays byte-clean.
- Bar / column / pie / doughnut / area templates carrying a stray `<a:ln>` on a series do not surface the stroke field — those families do not paint a connecting line.
- Clone flatten (line → column) post-build sweep strips both `smooth` and `stroke` so a coerced family does not carry leak-through styling the OOXML schema rejects.

Refs #136

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -611,6 +611,59 @@ export interface ChartDataLabels {
 }
 
 /**
+ * Preset dash pattern for a chart series line stroke.
+ *
+ * Mirrors the OOXML `ST_PresetLineDashVal` enum exactly. Each value
+ * names a stock pattern Excel paints without needing a custom dash
+ * array. The Excel "Format Data Series → Line → Dash type" UI exposes
+ * these stock patterns; Excel ignores any unrecognized value.
+ */
+export type ChartLineDashStyle =
+  | "solid"
+  | "dot"
+  | "dash"
+  | "lgDash"
+  | "dashDot"
+  | "lgDashDot"
+  | "lgDashDotDot"
+  | "sysDash"
+  | "sysDot"
+  | "sysDashDot"
+  | "sysDashDotDot";
+
+/**
+ * Per-series line stroke styling for line / scatter charts.
+ *
+ * Maps to the `<a:ln>` element nested inside `<c:ser><c:spPr>` — the
+ * same wrapper that already carries the series fill color. Only
+ * meaningful on `line` and `scatter` series; the field is silently
+ * dropped on every other chart family at all three layers (read,
+ * write, clone), since dashing and stroke width have no visible effect
+ * on bar / pie / doughnut / area renderings.
+ *
+ * Every field is optional — a bare `{}` collapses to no stroke
+ * configuration and leaves Excel's per-series default in place. Set
+ * `dash: "solid"` to explicitly reset a template's dashed stroke back
+ * to a continuous line.
+ */
+export interface ChartLineStroke {
+  /**
+   * Preset dash pattern. See {@link ChartLineDashStyle} for the
+   * accepted set.
+   */
+  dash?: ChartLineDashStyle;
+  /**
+   * Stroke width in points. Excel's UI exposes the 0.25 – 13.5 pt band;
+   * the writer clamps anything outside that range and rounds to the
+   * nearest quarter-point so a round-trip cannot drift. The OOXML
+   * attribute is in EMU (1 pt = 12 700 EMU); the writer performs the
+   * conversion and the reader inverts it. Non-finite values are
+   * dropped so the writer can elide the attribute entirely.
+   */
+  width?: number;
+}
+
+/**
  * A single data series inside a chart.
  *
  * `values` and `categories` are A1-style cell range references.
@@ -645,6 +698,14 @@ export interface ChartSeries {
    * Smoothed line".
    */
   smooth?: boolean;
+  /**
+   * Per-series line stroke (dash pattern + width) for line / scatter
+   * charts. Maps to `<a:ln>` inside `<c:ser><c:spPr>`. Ignored on every
+   * other chart family — bar / column / pie / doughnut / area never
+   * render a connecting line, so dashing and stroke width have no
+   * visible effect there. See {@link ChartLineStroke}.
+   */
+  stroke?: ChartLineStroke;
 }
 
 /**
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
Reviewed the per-series line stroke implementation end to end.

## Strengths
- The `ChartLineStroke` shape (preset `dash` + `width` in points) maps cleanly to OOXML `<a:ln>` and slots into the same `<c:spPr>` wrapper that already carries `series.color`, so `color + stroke` layers correctly without doubling the fill.
- Symmetric clamping on read and write keeps the 0.25–13.5 pt UI band stable, and snapping to the 0.25 pt grid before EMU conversion means a parsed-then-written stroke is byte-stable (the dedicated 9 525-EMU idempotency test pins this).
- `parseSeries` and the writer both gate the field on `line` / `line3D` / `scatter` only, and the post-build sweep in `cloneChart` strips `stroke` along with `smooth` when flattening to a non-line family — the negative-path tests on bar/column/pie/doughnut/area are exhaustive.
- Override grammar (`undefined` inherit / `null` drop / object replace wholesale) mirrors the prior smooth, dataLabels, and axis-scale resolvers exactly, and `cloneStroke` returning `undefined` for an empty `{}` keeps round-trips minimal.

## Conflict resolution
The branch was conflicting against `main` because PR #206 (chart series markers) shipped on the same `<c:ser>` line/scatter scope and touched almost every file this PR modifies. Merged `main` in and resolved by combining both fields side-by-side in:
- `src/_types.ts` — both `ChartLineStroke` + `ChartLineDashStyle` and `ChartMarker` + `ChartMarkerSymbol` exported, `ChartSeries.stroke` and `ChartSeries.marker` (and the matching `ChartSeriesInfo.*`) coexist.
- `src/index.ts` — both barrel re-exports merged into the alphabetized list.
- `src/xlsx/chart-clone.ts` — `CloneChartSeriesOverride` carries both `stroke` and `marker`, the post-build line/scatter sweep strips both alongside `smooth`, and `mergeSeries` calls both `resolveStroke` and `resolveMarker`.
- `src/xlsx/chart-reader.ts` — `parseSeries` calls `parseSeriesStroke` and `parseMarker` inside the same `line / line3D / scatter` guard.
- `src/xlsx/chart-writer.ts` — `buildLineChart` / `buildScatterChart` thread both `options.stroke` and `options.marker` into `buildSeries`, which threads stroke into the shared `buildSeriesSpPr` and emits the marker block in OOXML order (between `<c:spPr>` and `<c:dLbls>`).
- `test/chart-clone.test.ts`, `test/charts.test.ts`, `test/charts-write.test.ts` — the two feature describe blocks were tangled into one another; resolved by separating into two clean back-to-back blocks per file. All 2 825 tests pass after the merge.
- `README.md` — added a stroke paragraph mirroring the marker write-up in the read, write, and clone-through sections.

## Verification
- `pnpm test` (lint + typecheck + 2 825 vitest cases) passes.
- `pnpm build` produces 944 kB across 183 dist files with no errors.

LGTM, merging once CI confirms.

PR #208, opened 2026-05-02
feat(xlsx): chart dispBlanksAs — read, write, clone-through

productdevbook:
## Summary

Surfaces the chart-level `<c:dispBlanksAs>` element at the read, write, and clone layers so a template's "Hidden and Empty Cells" configuration survives `parseChart` → `cloneChart` → `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:dispBlanksAs>` is the OOXML element that controls how Excel paints series across cells whose value is missing or blank — Excel exposes the same knob under "Select Data Source → Hidden and Empty Cells". Until now hucre's writer always emitted a hardcoded `<c:dispBlanksAs val="gap"/>` and `parseChart` did not surface the parsed value, so a template using `span` to bridge gaps could not round-trip through clone. This bridges another chart-level configuration gap for the dashboard composition flow tracked in #136.

## API

```ts

// ── Read side ──
console.log(source.dispBlanksAs); // "span"

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "line",
      series: [
        { name: "Revenue", values: "B2:B6", categories: "A2:A6" },
      ],
      dispBlanksAs: "span", // bridge gaps in the source data
    }],
  }],
});

// ── Clone-through ──
  dispBlanksAs: "zero", // override the source: drop to zero instead
});
```

## Model

```ts
type ChartDisplayBlanksAs = "gap" | "zero" | "span";

interface SheetChart {
  /** ...existing fields... */
  dispBlanksAs?: ChartDisplayBlanksAs;
}

interface Chart {
  /** ...existing fields... */
  dispBlanksAs?: ChartDisplayBlanksAs;
}
```

The read-side `Chart.dispBlanksAs` mirrors the same shape so a parsed value slots straight back into `cloneChart` without transformation.

## Behavior

- **Read** — `parseChart` pulls `<c:chart><c:dispBlanksAs val="..">` and surfaces the literal value. The OOXML default `"gap"` collapses to `undefined` so absence and the default round-trip identically; unknown / malformed values drop to `undefined` rather than fabricate a token Excel rejects.
- **Write** — The writer threads `chart.dispBlanksAs` through to the rendered `<c:dispBlanksAs val="..">` (defaulting to `"gap"` when absent). Unknown values fall back to `"gap"` so a malformed input cannot produce invalid OOXML. The element is always emitted — Excel itself includes it in every reference serialization, so writing it explicitly keeps the file's intent unambiguous.
- **Clone** — `options.dispBlanksAs` accepts the standard `undefined` (inherit) / `null` (drop the inherited value, falling back to the writer's default) / value (replace) grammar that mirrors the dataLabels / smooth / axis-scale resolvers shipped earlier. Unlike smooth and marker, `<c:dispBlanksAs>` lives on `<c:chart>` and is valid on every chart family, so a coercion (line → column) preserves the inherited value.

## Edge cases

- A chart whose `<c:dispBlanksAs val="gap"/>` is emitted explicitly round-trips identically to a chart that omits the element entirely — both surface as `undefined` on the read side and re-emit as `gap` on the write side.
- An unknown `<c:dispBlanksAs val="bogus"/>` reads as `undefined` rather than surface a token Excel rejects.
- A missing `val` attribute on `<c:dispBlanksAs/>` reads as `undefined`.
- `dispBlanksAs: null` on `cloneChart` drops the inherited value so the writer falls back to the OOXML `"gap"` default — useful when a template uses `"span"` but the cloned dashboard chart should render gaps explicitly.
- Only one `<c:dispBlanksAs>` element is emitted per chart, even when the field is overridden — guards against any regression that would double-emit.

Refs #136

## Test plan

- [x] `pnpm test` (lint + typecheck + 2810 vitest tests, all passing)
- [x] `pnpm build`
- [x] Reader, writer, and clone tests cover: defaults, all three values, unknown/malformed inputs, missing `val`, and round-trip through `writeXlsx` → `parseChart`

### DIFF
--- a/README.md
+++ b/README.md
@@ -640,6 +640,13 @@ clockwise from 12 o'clock) on pie and doughnut charts. The OOXML
 default `0` (and the schema-equivalent `360`) collapses to
 `undefined` so absence and the default round-trip identically;
 non-pie / non-doughnut charts never report it.
+`Chart.dispBlanksAs` surfaces the chart-level
+`<c:chart><c:dispBlanksAs val=".."/>` element — Excel's "Select Data
+Source → Hidden and Empty Cells" knob — and reports the literal
+`"zero"` or `"span"` token. The OOXML default `"gap"` collapses to
+`undefined` so absence and the default round-trip identically;
+unknown / malformed values (and a missing `val` attribute) drop to
+`undefined` rather than fabricate a token Excel rejects.
 `ChartSeriesInfo.smooth` surfaces the per-series
 `<c:ser><c:smooth val=".."/>` flag — Excel's "Format Data Series →
 Line → Smoothed line" toggle — only on `line` / `line3D` / `scatter`
@@ -756,6 +763,14 @@ For pie and doughnut charts, `firstSliceAng` (0 – 360 degrees, default
 aligning paired charts in a dashboard. Out-of-band values wrap modulo
 360 (380 → 20, -90 → 270) the same way Excel's chart-formatting pane
 does, and non-pie / non-doughnut kinds silently ignore the field.
+The chart-level `dispBlanksAs` field maps to
+`<c:chart><c:dispBlanksAs val=".."/>` — Excel's "Select Data Source →
+Hidden and Empty Cells" toggle — and accepts `"gap"` (leave a break),
+`"zero"` (drop missing points to the X axis) or `"span"` (connect
+across the gap; line / scatter only). The writer always emits the
+element, defaulting to `"gap"` (the OOXML default Excel itself emits)
+and clamping unknown tokens back to `"gap"` so a malformed input
+cannot produce invalid OOXML.
 For line and scatter charts, each `series[i].smooth` flag toggles
 Excel's curved-line variant (`<c:smooth val="..">` inside `<c:ser>`).
 Line series always emit the element — `smooth: true` writes `val="1"`,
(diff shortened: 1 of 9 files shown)


productdevbook (commented):
## Review

Reviewed the chart `dispBlanksAs` round-trip change end-to-end.

### Correctness
- `parseChart` reads `<c:chart><c:dispBlanksAs val=".."/>` and surfaces only the three OOXML-defined tokens. The default `"gap"` collapses to `undefined`, and unknown values / missing `val` attributes drop to `undefined` rather than fabricate a token Excel rejects — exactly the pattern used by `parseFirstSliceAng` and `parseSmooth`.
- `writeChart` always emits `<c:dispBlanksAs>` (matches Excel's reference serialization) and clamps unknown tokens back to `"gap"` via `DISP_BLANKS_AS_VALUES`. Element ordering (after `c:plotVisOnly` inside `c:chart`) is correct per the OOXML CT_Chart sequence.
- `resolveDispBlanksAs` follows the same `undefined` (inherit) / `null` (drop) / value (replace) grammar as the dataLabels / smooth / marker resolvers. Critically, since `<c:dispBlanksAs>` is valid on every chart family, the clone correctly preserves the inherited value across a kind coercion (line → column), unlike smooth and marker.

### Test coverage
- 18 new tests across `parseChart`, `writeChart`, and `cloneChart` cover defaults, all three values, unknown / malformed inputs, missing `val`, write-side ordering, single-emit guard, every chart family, and a full `writeXlsx → parseChart` round-trip. No gaps spotted.

### Type safety / consistency
- `ChartDisplayBlanksAs` is exported from `index.ts` alongside `ChartDataLabelPosition` and the other public chart enums, and the field name / docs / override grammar are symmetric with the recently merged smooth, marker, and stroke features.

### Fixes applied on this branch
- Resolved a merge conflict in `src/xlsx/chart-writer.ts` against `main` (import-list conflict with the `ChartLineDashStyle` / `ChartLineStroke` imports introduced by the line-stroke PR — kept all three in alphabetical order).
- Added a README paragraph for `dispBlanksAs` in each of the three chart sections (read, write, clone), mirroring the surrounding feature docs (smooth / marker / stroke).
- Re-ran `pnpm test` (lint + typecheck + 2844 vitest tests, all passing) and `pnpm build` against the merged tree before pushing.

No remaining concerns. Merging once CI is green on the rebased head.

PR #209, opened 2026-05-02
feat(xlsx): chart series invertIfNegative flag — read, write, clone-through

productdevbook:
## Summary

Surfaces the per-series `<c:invertIfNegative>` flag at the read, write, and clone layers so a template's invert-if-negative state survives `parseChart` → `cloneChart` → `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:invertIfNegative>` is the OOXML element behind Excel's "Format Data Series → Fill → Invert if negative" toggle on bar / column charts — it tells Excel to paint negative bars in the inverted fill color (white by default) so a series with mixed-sign values reads at a glance. Until now hucre had no way to pin the flag per series, and `parseChart` did not surface it. This bridges another per-series styling gap for the dashboard composition flow tracked in #136.

## API

```ts

// ── Read side ──
console.log(source.series?.[0].invertIfNegative); // true

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      series: [
        { name: "Net P&L", values: "B2:B6", categories: "A2:A6",
          invertIfNegative: true },
        { name: "Baseline", values: "C2:C6", categories: "A2:A6" },
      ],
    }],
  }],
});

// ── Clone-through ──
  seriesOverrides: [
    { values: "Dashboard!$B$2:$B$13" },                          // inherits flag
    { values: "Dashboard!$C$2:$C$13", invertIfNegative: true },  // replace
    { values: "Dashboard!$D$2:$D$13", invertIfNegative: null },  // drop inherited
  ],
});
```

## Model

```ts
interface ChartSeries {
  /** ...existing fields... */
  invertIfNegative?: boolean;
}

interface ChartSeriesInfo {
  /** ...existing fields... */
  invertIfNegative?: boolean;
}
```

The read-side `ChartSeriesInfo.invertIfNegative` mirrors the same shape so a parsed value slots straight back into `cloneChart` without transformation.

## Behavior

- **Read** — `parseChart` pulls `<c:ser><c:invertIfNegative val=".."/>` only on `bar` / `bar3D` series. The OOXML default `false` collapses to `undefined` so absence and the default round-trip identically; missing `val` attributes drop to `undefined`; line / pie / doughnut / area / scatter templates carrying a stray `<c:invertIfNegative>` do not surface a setting that the writer would never emit anyway.
- **Write** — The writer emits `<c:invertIfNegative val="1"/>` only when the series sets the flag to `true`; `false` and absence both produce no element, matching Excel's reference serialization. The element lands between `<c:spPr>` and `<c:dLbls>` per the CT_BarSer schema (`idx → order → tx? → spPr? → invertIfNegative? → dLbls? → cat? → val?`). Non-bar / non-column families silently drop the field.
- **Clone** — `seriesOverrides[i].invertIfNegative` accepts the standard `undefined` (inherit) / `null` (drop the inherited flag) / `boolean` (replace) grammar that mirrors the smooth and stroke resolvers shipped earlier. The inherited flag is dropped automatically when the resolved clone target is anything other than `bar` or `column`, so a column → line flatten cannot leak the element into a chart kind whose schema rejects it.

## Edge cases

- A bar series with `invertIfNegative: false` round-trips identically to a series that omits the flag entirely — the writer omits `<c:invertIfNegative>` and the reader collapses both shapes to `undefined`.
- A line / pie / doughnut / area / scatter series carrying `invertIfNegative: true` is silently ignored at write time so the OOXML schema is never violated.
- An `<c:invertIfNegative/>` element with no `val` attribute reads as `undefined` rather than fabricate a flag Excel would not emit.
- The element is placed in the spec-required slot (after `<c:spPr>`, before `<c:cat>` / `<c:val>`) so Excel's strict validator accepts the file.

Refs #136

## Test plan

- [x] `pnpm test` (lint + typecheck + 2851 vitest tests, all passing)
- [x] `pnpm build`
- [x] Reader, writer, and clone tests cover: defaults, true / false / absent, missing `val`, multi-series independence, ignored on non-bar families, OOXML element ordering, and a full `parseChart → cloneChart → writeXlsx → parseChart` round-trip.

### DIFF
--- a/README.md
+++ b/README.md
@@ -665,6 +665,13 @@ and `width` is reported in points after converting from EMU and
 clamping to Excel's 0.25 – 13.5 pt UI band. Empty `<a:ln/>` blocks,
 unknown dash tokens, and out-of-band widths collapse to `undefined`
 so the parsed shape stays minimal.
+`ChartSeriesInfo.invertIfNegative` surfaces the per-series
+`<c:ser><c:invertIfNegative val=".."/>` flag — Excel's "Format Data
+Series → Fill → Invert if negative" toggle — only on `bar` / `bar3D`
+series (the OOXML schema places `<c:invertIfNegative>` exclusively on
+`CT_BarSer` / `CT_Bar3DSer`). Absence and the OOXML default
+`val="0"` both collapse to `undefined`, so only an explicit
+`<c:invertIfNegative val="1"/>` round-trips as `invertIfNegative: true`.
 Sheets that hucre actively regenerates because they
 also carry hucre-managed images currently keep the chart bodies but
 lose the in-drawing chart anchor — merging hucre's drawing output
@@ -785,6 +792,13 @@ so a `color + stroke` combo behaves like Excel's UI: the line picks
 up the fill color, and dash / width override visibility-only
 attributes. Bar / column / pie / doughnut / area kinds silently
 ignore the field.
+For bar and column charts, each `series[i].invertIfNegative` flag
+mirrors Excel's "Format Data Series → Fill → Invert if negative"
+toggle (`<c:invertIfNegative val=".."/>` inside `<c:ser>`). The
+writer only emits the element when `invertIfNegative` is explicitly
+`true` — `false` and absence both round-trip as the OOXML default
+(omission), so untouched bar charts stay byte-clean. Line / pie /
+doughnut / area / scatter kinds silently ignore the flag.
 Radar, stock, 3D variants, trendlines, and combo charts are out of
 scope today.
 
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
## Review

**Scope.** Adds the per-series `invertIfNegative` flag at the read, write, and clone layers (`<c:invertIfNegative>` on `CT_BarSer` / `CT_Bar3DSer`). Surface area is small and consistent with the existing per-series feature shape (smooth, marker, stroke).

**Correctness.**
- Reader gates surfacing on `kind === "bar" || kind === "bar3D"` and collapses `false` / missing `val` to `undefined`, matching the writer's elision so absence and the OOXML default round-trip identically.
- Writer slot is correct per `CT_BarSer`: `idx → order → tx? → spPr? → invertIfNegative? → dLbls? → cat? → val?`. The marker slot is occupied only on line/scatter, so there's no risk of double-population in bar callers.
- Clone resolver follows the established `undefined` (inherit) / `null` (drop) / boolean (replace) grammar, with `false` collapsing to `undefined` for symmetry. The post-build strip drops the field on any non-bar/non-column resolved type, even when the caller passes `options.series` wholesale.

**Tests.** Comprehensive — 25+ new cases across reader, writer, clone, and a full `parseChart → cloneChart → writeXlsx → parseChart` round-trip. Edge cases covered: explicit `false`, missing `val`, multi-series independence, non-bar families ignored at all three layers, and OOXML element ordering enforced.

**Fixes I applied on this branch.**
- Resolved a 3-way merge conflict with `main` (`dispBlanksAs` PR #205 landed in the same regions): kept both new doc paragraphs in `README.md`, and both new `describe` blocks in `test/chart-clone.test.ts` and `test/charts-write.test.ts`.
- Re-ran `pnpm test` (lint + typecheck + 2870 vitest tests, all passing) and `pnpm build` (954 kB total, 183 files) on the merged tree before pushing.

No remaining concerns from my side. Holding for CI before merging.

PR #210, opened 2026-05-02
feat(xlsx): chart varyColors flag — read, write, clone-through

productdevbook:
## Summary

Surfaces the chart-level `<c:varyColors>` element at the read, write, and clone layers so a template's "Vary colors by point" toggle survives `parseChart` → `cloneChart` → `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:varyColors>` is the OOXML element behind Excel's "Format Data Series → Fill → Vary colors by point" toggle. The OOXML schema places the element on every chart-type element except `surfaceChart` / `surface3DChart` / `stockChart`, with per-family defaults that flip its meaning:

- `pie` / `doughnut` / `pie3D` / `ofPie` default to `true` (each slice paints in a unique color)
- bar / column / line / area / scatter default to `false` (every point on a series shares one color)

Until now hucre's writer hardcoded those defaults and `parseChart` did not surface the value, so a template using the non-default value (a column chart with rainbow bars, or a doughnut collapsed to a single color) could not round-trip through clone. This bridges another chart-level configuration gap for the dashboard composition flow tracked in #136.

## API

```ts

// ── Read side ──
console.log(source.varyColors); // true on a column chart with rainbow bars,
                                // false on a doughnut collapsed to one color,
                                // undefined when the template uses the family default

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [
      // Single-series column chart with one color per bar.
      {
        type: "column",
        varyColors: true,
      },
      // Doughnut collapsed to a single color (Excel's "single color" preset).
      {
        type: "doughnut",
        series: [{ name: "Share", values: "B2:B6", categories: "A2:A6" }],
        anchor: { from: { row: 6, col: 8 } },
        varyColors: false,
      },
    ],
  }],
});

// ── Clone-through ──
  varyColors: false,            // replace
  // varyColors: null,          // drop the inherited flag (writer falls back to per-family default)
  // varyColors: undefined,     // inherit the source's parsed value
});
```

## Model

```ts
interface SheetChart {
  /** ...existing fields... */
  varyColors?: boolean;
}

interface Chart {
  /** ...existing fields... */
  varyColors?: boolean;
}

interface CloneChartOptions {
  /** ...existing fields... */
  varyColors?: boolean | null;
}
```

The read-side `Chart.varyColors` mirrors the same shape so a parsed value slots straight back into `cloneChart` without transformation.

## Behavior

- **Read** — `parseChart` pulls `<c:varyColors val=".."/>` off the first chart-type element that carries one. Values matching the per-family OOXML default collapse to `undefined` so absence and the default round-trip identically; only non-default values surface. Pie / doughnut / pie3D / ofPie default to `true`; every other family defaults to `false`. The reader accepts the OOXML truthy / falsy spellings (`"1"` / `"true"` / `"0"` / `"false"`); unknown values and missing `val` attributes drop to `undefined`.
- **Write** — The writer always emits `<c:varyColors>` on the chart-type element (the OOXML schema lists it as required on every authored family). Absent `varyColors`, the writer falls back to the per-family default (`val="1"` for pie / doughnut, `val="0"` for bar / column / line / area / scatter), matching Excel's reference serialization. An explicit `chart.varyColors` always wins, so a column chart can flip to `val="1"` and a doughnut to `val="0"`.
- **Clone** — `options.varyColors` accepts the standard `undefined` (inherit) / `null` (drop the inherited flag, falling back to the writer's per-family default) / `boolean` (replace) grammar that mirrors `dispBlanksAs` resolved earlier. The inherited value carries through every coercion (column → pie, doughnut → line, etc.) since the element is valid on every authored family.

## Edge cases

- A column chart with `varyColors: false` round-trips identically to a column chart that omits the field — the writer emits the per-family default `val="0"` and the reader collapses both shapes to `undefined`.
- A pie chart with `varyColors: true` round-trips identically to a pie chart that omits the field (default `val="1"` collapses to `undefined`); pinning `varyColors: false` flips to `val="0"` and re-parses as `false` (the non-default).
- A `<c:varyColors/>` element with no `val` attribute reads as `undefined` rather than fabricate a flag Excel would not emit.
- The element is placed in the spec-required slot inside each chart-type element — after `<c:barDir>`/`<c:grouping>` on `<c:barChart>`, after `<c:scatterStyle>` on `<c:scatterChart>`, etc. — so Excel's strict validator accepts the file.

Refs #136

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm vitest run --exclude="**/.claude/**"` (all 2874 tests passing, including 32 new `varyColors` tests across reader, writer, and clone)
- [x] Reader, writer, and clone tests cover: per-family defaults, true / false / absent, the `"true"` / `"false"` spellings, missing `val`, unknown tokens, multi-element selection, OOXML element ordering inside `<c:barChart>` / `<c:scatterChart>`, and a full `parseChart → cloneChart → writeXlsx → parseChart` round-trip.

### DIFF
--- a/README.md
+++ b/README.md
@@ -647,6 +647,17 @@ Source → Hidden and Empty Cells" knob — and reports the literal
 `undefined` so absence and the default round-trip identically;
 unknown / malformed values (and a missing `val` attribute) drop to
 `undefined` rather than fabricate a token Excel rejects.
+`Chart.varyColors` surfaces the chart-type element's
+`<c:varyColors val=".."/>` flag — Excel's "Format Data Series → Fill →
+Vary colors by point" toggle — and reports the literal `true` or
+`false` value when it differs from the per-family OOXML default. Pie /
+doughnut / pie3D / ofPie default to `true` (every slice paints in a
+unique color), so absence and `<c:varyColors val="1"/>` both collapse
+to `undefined`; only an explicit `val="0"` surfaces `false` (the
+single-color override). Every other chart family defaults to `false`,
+so absence and `<c:varyColors val="0"/>` both collapse to `undefined`
+and only an explicit `val="1"` surfaces `true`. Unknown / malformed
+values and a missing `val` attribute drop to `undefined`.
 `ChartSeriesInfo.smooth` surfaces the per-series
 `<c:ser><c:smooth val=".."/>` flag — Excel's "Format Data Series →
 Line → Smoothed line" toggle — only on `line` / `line3D` / `scatter`
@@ -771,6 +782,15 @@ across the gap; line / scatter only). The writer always emits the
 element, defaulting to `"gap"` (the OOXML default Excel itself emits)
 and clamping unknown tokens back to `"gap"` so a malformed input
 cannot produce invalid OOXML.
+The chart-level `varyColors` field maps to `<c:varyColors val=".."/>`
+on the chart-type element — Excel's "Format Data Series → Fill → Vary
+colors by point" toggle. Absent it, the writer falls back to Excel's
+per-family defaults (`true` on pie / doughnut, `false` on bar / column
+/ line / area / scatter), so a fresh chart matches Excel's reference
+serialization. Pin `varyColors: true` on a single-series column or
+bar chart to paint each bar a different color, or pin `false` on a
+doughnut to collapse every wedge to one color (Excel's "single color"
+preset).
 For line and scatter charts, each `series[i].smooth` flag toggles
 Excel's curved-line variant (`<c:smooth val="..">` inside `<c:ser>`).
 Line series always emit the element — `smooth: true` writes `val="1"`,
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
Reviewed the diff end-to-end. The implementation cleanly mirrors the existing `dispBlanksAs` pattern across read / write / clone, which is a nice consistency win at the call site.

Correctness checks:
- Reader's `VARY_COLORS_DEFAULT_TRUE` set (`pie`, `pie3D`, `doughnut`, `ofPie`) matches the OOXML per-family defaults; every other family falls through to the false-default branch via the `Set.has` check, including `bar3D` / `line3D` / `area3D` / `bubble` / `radar`.
- Writer's `VARY_COLORS_DEFAULT_TRUE_TYPES` correctly lists only `pie` / `doughnut` because `WriteChartKind` doesn't include `pie3D` / `ofPie`.
- The collapse-to-undefined logic (`parsed === familyDefaultsTrue`) round-trips absence and the per-family default identically, so a fresh column chart (val="0") and a defaulted pie chart (val="1") both surface `undefined`.
- `findChild` walks direct children, which is correct for `<c:varyColors>` as a direct child of the chart-type element.
- Element ordering inside `<c:barChart>` (after `barDir` / `grouping`) and `<c:scatterChart>` (after `scatterStyle`) is asserted by tests and matches the OOXML CT_*Chart sequences.
- Clone resolver follows the same `undefined` (inherit) / `null` (drop) / `boolean` (replace) grammar as `dispBlanksAs`, and the field carries through every coercion since the element is valid on every authored family.

Test coverage is thorough: per-family defaults, true/false/absent, "true"/"false" spellings, missing `val`, unknown tokens, multi-element selection on combo charts, OOXML element ordering, single-emission guard, and full parse → clone → write → re-parse round-trip on both column and doughnut. README is updated in the read / write / clone sections to match the new field.

CI is green on the head commit. Merging.

PR #211, opened 2026-05-03
feat(xlsx): chart scatterStyle preset — read, write, clone-through

productdevbook:
## Summary

Surfaces the chart-level `<c:scatterStyle>` element at the read, write, and clone layers so a scatter template's chart-picker preset (`"none"` / `"line"` / `"lineMarker"` / `"marker"` / `"smooth"` / `"smoothMarker"`) survives `parseChart` → `cloneChart` → `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:scatterStyle>` is the OOXML element behind Excel's "Change Chart Type → XY (Scatter)" picker. The OOXML schema places the element exclusively on `<c:scatterChart>` and lists it as required there. Until now hucre's writer hardcoded `val="lineMarker"` and `parseChart` did not surface the value, so a template authored with smoothed lines or markers-only could not round-trip through clone. This bridges another chart-level configuration gap for the dashboard composition flow tracked in #136.

## API

```ts

// ── Read side ──
console.log(source.scatterStyle); // "smooth" on a smooth-line scatter,
                                  // "marker" on a markers-only scatter,
                                  // undefined when the chart is not scatter
                                  // or omits the element

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [
      // Scatter with smoothed curves and markers.
      {
        type: "scatter",
        series: [{ name: "Trend", values: "B2:B6", categories: "A2:A6" }],
        scatterStyle: "smoothMarker",
      },
      // Markers only (no connecting line).
      {
        type: "scatter",
        series: [{ name: "Sample", values: "B2:B6", categories: "A2:A6" }],
        anchor: { from: { row: 6, col: 8 } },
        scatterStyle: "marker",
      },
    ],
  }],
});

// ── Clone-through ──
  scatterStyle: "smooth",          // replace
  // scatterStyle: null,           // drop the inherited preset (writer falls back to "lineMarker")
  // scatterStyle: undefined,      // inherit the source's parsed value
});
```

## Model

```ts
type ChartScatterStyle =
  | "none"
  | "line"
  | "lineMarker"
  | "marker"
  | "smooth"
  | "smoothMarker";

interface SheetChart {
  /** ...existing fields... */
  scatterStyle?: ChartScatterStyle;
}

interface Chart {
  /** ...existing fields... */
  scatterStyle?: ChartScatterStyle;
}

interface CloneChartOptions {
  /** ...existing fields... */
  scatterStyle?: ChartScatterStyle | null;
}
```

The read-side `Chart.scatterStyle` mirrors the same shape so a parsed value slots straight back into `cloneChart` without transformation.

## Behavior

- **Read** — `parseChart` pulls `<c:scatterStyle val=".."/>` off the first `<c:scatterChart>` element. Every literal value in the OOXML `ST_ScatterStyle` enum surfaces as-is so a clone preserves the exact preset; missing elements, missing `val` attributes, and tokens outside the enum drop to `undefined`. The reader does not collapse the writer's `lineMarker` default — keeping every recognized token literal makes the round-trip exact.
- **Write** — The writer always emits `<c:scatterStyle>` on `<c:scatterChart>` (the OOXML schema lists it as required there). Absent `scatterStyle`, the writer falls back to `"lineMarker"` — Excel's chart-picker default and what every fresh hucre scatter chart already used. Values outside the enum fall back to the same default rather than emit a token Excel's strict validator would reject. Non-scatter chart families silently ignore the field.
- **Clone** — `options.scatterStyle` accepts the standard `undefined` (inherit) / `null` (drop, fall back to the writer's `"lineMarker"` default) / value (replace) override grammar that mirrors `dispBlanksAs` / `varyColors`. The inherited value is silently dropped when the resolved clone target is anything other than `scatter` since the schema rejects `<c:scatterStyle>` on every other chart family.

## Edge cases

- A scatter chart authored with `scatterStyle: "lineMarker"` round-trips identically to one that omits the field — the writer emits the same default token either way and the reader surfaces it literally.
- A `<c:scatterStyle/>` element with no `val` attribute reads as `undefined` rather than fabricate a preset.
- An out-of-enum token on the read side (e.g. `val="bogus"`) drops to `undefined`; the writer never emits an unknown token even when one is set on `SheetChart` directly.
- A scatter template flattened to a column / line clone drops the inherited preset so it does not leak into a chart kind whose schema rejects the element.
- The element is placed in the spec-required slot inside `<c:scatterChart>` (before `<c:varyColors>` and `<c:ser>`) so Excel's strict validator accepts the file.

Refs #136

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm test` (lint + typecheck + vitest) passes — all 2924 tests, including 24 new `scatterStyle` tests across reader, writer, and clone
- [x] Reader, writer, and clone tests cover: every enum value, the writer's `lineMarker` default, missing `val`, unknown tokens, omitted element, gating on the `<c:scatterChart>` parent, OOXML element ordering inside `<c:scatterChart>`, drop-on-non-scatter coercion, and a full `parseChart → cloneChart → writeXlsx → parseChart` round-trip

### DIFF
--- a/README.md
+++ b/README.md
@@ -658,6 +658,15 @@ single-color override). Every other chart family defaults to `false`,
 so absence and `<c:varyColors val="0"/>` both collapse to `undefined`
 and only an explicit `val="1"` surfaces `true`. Unknown / malformed
 values and a missing `val` attribute drop to `undefined`.
+`Chart.scatterStyle` surfaces `<c:scatterChart><c:scatterStyle
+val=".."/></c:scatterChart>` — the chart-level XY-scatter preset
+Excel selects in the chart-type picker (`"none"`, `"line"`,
+`"lineMarker"`, `"marker"`, `"smooth"`, or `"smoothMarker"`). Every
+recognized token surfaces literally so a clone preserves the exact
+preset; missing elements, missing `val` attributes, and tokens
+outside the OOXML enum drop to `undefined`. Only `scatter` charts
+report the field — the schema places the element exclusively on
+`<c:scatterChart>`.
 `ChartSeriesInfo.smooth` surfaces the per-series
 `<c:ser><c:smooth val=".."/>` flag — Excel's "Format Data Series →
 Line → Smoothed line" toggle — only on `line` / `line3D` / `scatter`
@@ -798,6 +807,18 @@ serialization. Pin `varyColors: true` on a single-series column or
 bar chart to paint each bar a different color, or pin `false` on a
 doughnut to collapse every wedge to one color (Excel's "single color"
 preset).
+The chart-level `scatterStyle` field maps to `<c:scatterStyle val=".."/>`
+on `<c:scatterChart>` and picks one of Excel's six XY-scatter presets:
+`"none"` / `"marker"` (markers only), `"line"` (straight lines, no
+markers), `"lineMarker"` (Excel's chart-picker default — straight
+lines with markers; the writer's fallback when the field is absent),
+`"smooth"` (smoothed curves, no markers), or `"smoothMarker"`
+(smoothed curves with markers). The element is required by the OOXML
+schema on `<c:scatterChart>` and the writer always emits it; values
+outside the enum fall back to `"lineMarker"` so a malformed input
+cannot produce invalid OOXML. Other chart kinds silently ignore the
+field — the schema places `<c:scatterStyle>` exclusively on
+`<c:scatterChart>`.
 For line and scatter charts, each `series[i].smooth` flag toggles
 Excel's curved-line variant (`<c:smooth val="..">` inside `<c:ser>`).
 Line series always emit the element — `smooth: true` writes `val="1"`,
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
Reviewed the diff end-to-end. The implementation cleanly mirrors the established pattern used by the recently merged `varyColors` (#210) and `dispBlanksAs` (#208) features:

- **Reader** (`chart-reader.ts`): `parseScatterStyle` is properly gated on `kind === "scatter"` so the element is only surfaced when it appears on its schema-permitted parent. The `VALID_SCATTER_STYLES` set guards against unknown tokens and missing `val` attributes both collapse to `undefined`. Good defensive parsing.
- **Writer** (`chart-writer.ts`): `resolveScatterStyle` validates against the same enum and falls back to `"lineMarker"` (Excel's chart-picker default), preserving prior behavior for charts that omit the field. The element is emitted only on `buildScatterChart`, so non-scatter families never see it. Element ordering inside `<c:scatterChart>` (scatterStyle -> varyColors -> ser*) matches the OOXML CT_ScatterChart sequence.
- **Clone** (`chart-clone.ts`): `resolveScatterStyle` follows the same `undefined`/`null`/value override grammar as `resolveVaryColors`, and the resolution is correctly gated behind `type === "scatter"` so a flattened clone (scatter -> line/column) does not leak the field into a chart kind that rejects it.
- **Types** (`_types.ts`): The `ChartScatterStyle` union covers the full ST_ScatterStyle enum with thorough JSDoc explaining each preset's render and the asymmetry between the schema default (`marker`) and the writer default (`lineMarker`).
- **Tests**: 24 new tests across reader/writer/clone covering every enum value, missing element, missing val, unknown token, parent gating, element ordering, drop-on-non-scatter coercion, and full round-trip via `writeXlsx` + `parseChart`. The combo-chart test (first `<c:scatterChart>` wins) matches how other chart-type fields behave.
- **Docs**: README updates land in the right sections (read-side `Chart` table, write-side `SheetChart`, and `cloneChart` override grammar).

CI is SUCCESS on commit c64e58b. Merging.

PR #212, opened 2026-05-03
feat(xlsx): chart plotVisOnly toggle — read, write, clone-through

productdevbook:
## Summary

Surfaces the chart-level `<c:plotVisOnly>` element at the read, write, and clone layers so a template's "Show data in hidden rows and columns" toggle survives `parseChart` → `cloneChart` → `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:plotVisOnly>` is the OOXML element behind Excel's "Hidden and Empty Cells → Show data in hidden rows and columns" checkbox. The two are inverses: when the checkbox is checked, hidden cells stay in the chart and `plotVisOnly` is `false`; when unchecked (Excel's UI default), hidden cells drop out and `plotVisOnly` is `true`. Until now hucre's writer hardcoded `val="1"` and `parseChart` did not surface the value, so a template that overrode the default (commonly used when the source range carries hidden helper rows the chart should still plot) could not round-trip through clone. This bridges another chart-level configuration gap for the dashboard composition flow tracked in #136.

## API

```ts

// ── Read side ──
console.log(source.plotVisOnly); // false on a chart that keeps hidden cells,
                                 // undefined when the template uses the OOXML default

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [
      // Keep hidden helper rows in the rendered chart.
      {
        type: "column",
        plotVisOnly: false,
      },
    ],
  }],
});

// ── Clone-through ──
  plotVisOnly: false,            // replace
  // plotVisOnly: null,          // drop the inherited flag (writer falls back to OOXML default true)
  // plotVisOnly: undefined,     // inherit the source's parsed value
});
```

## Model

```ts
interface SheetChart {
  /** ...existing fields... */
  plotVisOnly?: boolean;
}

interface Chart {
  /** ...existing fields... */
  plotVisOnly?: boolean;
}

interface CloneChartOptions {
  /** ...existing fields... */
  plotVisOnly?: boolean | null;
}
```

The read-side `Chart.plotVisOnly` mirrors the same shape so a parsed value slots straight back into `cloneChart` without transformation.

## Behavior

- **Read** — `parseChart` pulls `<c:plotVisOnly val=".."/>` off `<c:chart>`. The OOXML default `true` collapses to `undefined` so absence and the default round-trip identically; only an explicit `val="0"` surfaces `false`. The reader accepts the OOXML truthy / falsy spellings (`"1"` / `"true"` / `"0"` / `"false"`); unknown values and missing `val` attributes drop to `undefined`.
- **Write** — The writer always emits `<c:plotVisOnly>` on `<c:chart>` (matching Excel's reference serialization, which always carries the element). Absent `plotVisOnly`, the writer emits `val="1"` (the OOXML default). An explicit `chart.plotVisOnly === false` flips it to `val="0"`.
- **Clone** — `options.plotVisOnly` accepts the standard `undefined` (inherit) / `null` (drop the inherited flag, falling back to the writer's default `true`) / `boolean` (replace) grammar that mirrors `dispBlanksAs` and `varyColors`. The inherited value carries through every coercion (column → pie, doughnut → line, etc.) since the element lives on `<c:chart>` and is valid on every chart family.

## Edge cases

- A chart with `plotVisOnly: true` round-trips identically to a chart that omits the field — the writer emits `val="1"` and the reader collapses both shapes to `undefined`.
- A chart with `plotVisOnly: false` round-trips as `false` because that is the non-default value.
- A `<c:plotVisOnly/>` element with no `val` attribute reads as `undefined` rather than fabricate a flag Excel would not emit.
- The element is placed in the spec-required slot inside `<c:chart>` — after `<c:legend>` and before `<c:dispBlanksAs>` — so Excel's strict validator accepts the file.

Refs #136

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm test` (lint + typecheck + vitest, all 2948 tests passing including 25 new `plotVisOnly` tests across reader, writer, and clone)
- [x] Reader, writer, and clone tests cover: defaults, true / false / absent, the `"true"` / `"false"` spellings, missing `val`, unknown tokens, OOXML element ordering inside `<c:chart>`, single-emission guard, every chart family, and a full `parseChart -> cloneChart -> writeXlsx -> parseChart` round-trip.

### DIFF
--- a/README.md
+++ b/README.md
@@ -667,6 +667,16 @@ preset; missing elements, missing `val` attributes, and tokens
 outside the OOXML enum drop to `undefined`. Only `scatter` charts
 report the field — the schema places the element exclusively on
 `<c:scatterChart>`.
+`Chart.plotVisOnly` surfaces the chart-level
+`<c:chart><c:plotVisOnly val=".."/>` flag — the inverse of Excel's
+"Hidden and Empty Cells → Show data in hidden rows and columns"
+checkbox. The OOXML default `true` (hidden cells drop out) collapses
+to `undefined` so absence and `<c:plotVisOnly val="1"/>` round-trip
+identically; only an explicit `val="0"` surfaces `false` (the
+non-default that keeps hidden cells in the chart). The reader accepts
+the OOXML truthy / falsy spellings (`"1"` / `"true"` / `"0"` /
+`"false"`); unknown values and missing `val` attributes drop to
+`undefined`.
 `ChartSeriesInfo.smooth` surfaces the per-series
 `<c:ser><c:smooth val=".."/>` flag — Excel's "Format Data Series →
 Line → Smoothed line" toggle — only on `line` / `line3D` / `scatter`
@@ -819,6 +829,15 @@ outside the enum fall back to `"lineMarker"` so a malformed input
 cannot produce invalid OOXML. Other chart kinds silently ignore the
 field — the schema places `<c:scatterStyle>` exclusively on
 `<c:scatterChart>`.
+The chart-level `plotVisOnly` field maps to `<c:plotVisOnly val=".."/>`
+on `<c:chart>` — the inverse of Excel's "Hidden and Empty Cells →
+Show data in hidden rows and columns" checkbox. Absent it, the writer
+emits the OOXML default `val="1"` (hidden rows / columns drop out of
+the chart), matching Excel's reference serialization. Pin
+`plotVisOnly: false` to keep hidden helper cells in the rendered
+chart (`val="0"`). The writer always emits the element so the
+rendered intent is explicit on roundtrip — no chart family is special-
+cased.
 For line and scatter charts, each `series[i].smooth` flag toggles
 Excel's curved-line variant (`<c:smooth val="..">` inside `<c:ser>`).
 Line series always emit the element — `smooth: true` writes `val="1"`,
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
Reviewed the PR end-to-end. The changes look correct and complete.

What I checked:
- Reader (`parsePlotVisOnly`) collapses the OOXML default `true` and the `"true"` spelling to `undefined`, surfaces only `false` for an explicit `"0"` / `"false"`, and drops missing-`val` / unknown tokens. Tests cover all of these.
- Writer always emits `<c:plotVisOnly>` (matching Excel's reference serialization), correctly placed between `<c:legend>` and `<c:dispBlanksAs>` in the CT_Chart sequence. The hardcoded `val=1` was replaced with `resolvePlotVisOnly(chart) ? 1 : 0`, no double-emission.
- Clone resolver mirrors the `dispBlanksAs` / `varyColors` `undefined`/`null`/`boolean` grammar and is invoked before the scatter-style coercion so the chart-family-agnostic flag carries through every type flatten.
- Type docs on `SheetChart.plotVisOnly`, `Chart.plotVisOnly`, and `CloneChartOptions.plotVisOnly` are precise and consistent with the existing OOXML-default-collapses-to-undefined pattern.
- README has matching reader / writer / clone-options entries; no separate web docs page exists for these chart toggles, so README is the right surface.
- 25 new tests across reader, writer, and clone — including a full `cloneChart -> writeXlsx -> parseChart` roundtrip and a per-chart-family sweep. CI is green.

No remaining concerns. Merging.

PR #213, opened 2026-05-03
feat(xlsx): chart axis tick marks and tick label position — read, write, clone-through

productdevbook:
## Summary

Surfaces the per-axis `<c:majorTickMark>` / `<c:minorTickMark>` / `<c:tickLblPos>` elements at the read, write, and clone layers so a template's tick rendering survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The three elements live alongside `<c:numFmt>` on every CT_CatAx / CT_ValAx / CT_DateAx / CT_SerAx and control where Excel paints the short tick marks (`none` / `in` / `out` / `cross`) and where the numeric or category tick labels sit relative to the axis line (`nextTo` / `low` / `high` / `none`). Until now hucre's writer never emitted any of the three, so a template that anchored axis labels to the chart edge with `tickLblPos="low"` or styled the value-axis ticks with `majorTickMark="cross"` flattened to Excel's defaults on every parse -> clone -> write loop. This bridges another axis-configuration gap for the dashboard composition flow tracked in #136.

## API

```ts

// ── Read side ──
console.log(source.axes?.y?.majorTickMark); // "cross" when the template overrode the OOXML default,
                                            // undefined when the template used "out" (the default)
console.log(source.axes?.y?.tickLblPos);    // "low" when labels are pinned to the chart edge

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      axes: {
        y: {
          majorTickMark: "cross", // tick marks straddle the axis line
          minorTickMark: "in",    // minor ticks point inward
          tickLblPos: "low",      // labels pinned to the bottom edge
        },
      },
    }],
  }],
});

// ── Clone-through ──
  axes: {
    y: {
      majorTickMark: "out",   // replace
      minorTickMark: null,    // drop the inherited flag (writer falls back to "none")
      // tickLblPos: undefined  // inherit the source's parsed value
    },
  },
});
```

## Model

```ts
export type ChartAxisTickMark = "none" | "in" | "out" | "cross";
export type ChartAxisTickLabelPosition = "nextTo" | "low" | "high" | "none";

interface ChartAxisInfo {
  /* ...existing fields... */
  majorTickMark?: ChartAxisTickMark;
  minorTickMark?: ChartAxisTickMark;
  tickLblPos?: ChartAxisTickLabelPosition;
}

interface SheetChart {
  axes?: {
    x?: { /* ... */; majorTickMark?: ChartAxisTickMark; minorTickMark?: ChartAxisTickMark; tickLblPos?: ChartAxisTickLabelPosition };
    y?: { /* ... */; majorTickMark?: ChartAxisTickMark; minorTickMark?: ChartAxisTickMark; tickLblPos?: ChartAxisTickLabelPosition };
  };
}

interface CloneChartOptions {
  axes?: {
    x?: { /* ... */; majorTickMark?: ChartAxisTickMark | null; minorTickMark?: ChartAxisTickMark | null; tickLblPos?: ChartAxisTickLabelPosition | null };
    y?: { /* ... */; majorTickMark?: ChartAxisTickMark | null; minorTickMark?: ChartAxisTickMark | null; tickLblPos?: ChartAxisTickLabelPosition | null };
  };
}
```

The read-side `ChartAxisInfo.{majorTickMark,minorTickMark,tickLblPos}` mirrors the same shape so a parsed value slots straight back into `cloneChart` without transformation.

## Behavior

- **Read** — `parseChart` pulls the three elements off `<c:catAx>` / `<c:valAx>`. The OOXML defaults (`majorTickMark="out"`, `minorTickMark="none"`, `tickLblPos="nextTo"`) collapse to `undefined` so absence and the default round-trip identically; only non-default values surface. Unknown enum tokens drop to `undefined` rather than fabricate a value Excel would reject.
- **Write** — Each element is emitted only when the matching field is pinned (no defaults injected), keeping fresh charts visually identical to Excel's reference serialization. The OOXML schema's strict child order is preserved: `axId` -> `scaling` -> `delete` -> `axPos` -> `majorGridlines` -> `minorGridlines` -> `title` -> `numFmt` -> `majorTickMark` -> `minorTickMark` -> `tickLblPos` -> `crossAx` -> ... Invalid enum tokens are filtered at write time so a typo'd input cannot produce invalid OOXML.
- **Clone** — Each axis override accepts the standard `undefined` (inherit) / `null` (drop the inherited flag, falling back to the writer's per-element default) / value (replace) grammar that mirrors `gridlines` / `scale` / `numberFormat`. When the resolved clone target is `pie` or `doughnut` the entire `axes` block is silently dropped (neither has axes in OOXML).

## Edge cases

- A chart with `majorTickMark: "out"` round-trips identically to a chart that omits the field — the writer omits the element on the second pass and the reader collapses both shapes to `undefined`.
- Values outside the `ST_TickMark` / `ST_TickLblPos` enums are filtered at every layer (read, write, clone) so a corrupt template cannot leak into the output.
- A `<c:majorTickMark/>` with no `val` attribute reads as `undefined` rather than fabricate a flag.
- Bar / column / line / area / scatter all support tick rendering on both axes; pie / doughnut silently drop the entire `axes` block since OOXML defines no axes for them.
- The element is placed in the spec-required slot inside `<c:catAx>` / `<c:valAx>` — after `<c:numFmt>` and before `<c:crossAx>` — so Excel's strict validator accepts the file.

Refs #136

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm test` (lint + typecheck + vitest, all 2964 tests passing including 33 new tests across reader, writer, and clone for the three tick-rendering elements)
- [x] Reader, writer, and clone tests cover: defaults, every enum value, missing `val`, unknown tokens, OOXML element ordering inside the axis, single-emission guard, every chart family, scatter axis layout (axPos="b" vs "l"), pie/doughnut drop-through, and a full `parseChart -> cloneChart -> writeChart -> parseChart` round-trip with a `writeXlsx` end-to-end check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -543,8 +543,11 @@ for (const sheet of wb.sheets) {
     // e.g. "bottom" "stacked" undefined undefined
 
     // chart.axes carries per-axis labels, gridline visibility, numeric
-    // scaling, and tick-label number format pulled from <c:catAx>/<c:valAx>.
-    // Only populated axes show up — pie/doughnut never do.
+    // scaling, tick-label number format and tick rendering (major /
+    // minor tick mark style and tick-label position) pulled from
+    // <c:catAx>/<c:valAx>. Only populated axes show up — pie/doughnut
+    // never do, and OOXML defaults collapse to undefined so absence
+    // and the default round-trip identically.
     console.log(chart.axes);
     // e.g. {
     //   x: { title: "Quarter" },
@@ -553,6 +556,8 @@ for (const sheet of wb.sheets) {
     //     gridlines: { major: true },
     //     scale: { min: 0, max: 100, majorUnit: 25 },
     //     numberFormat: { formatCode: "$#,##0" },
+    //     majorTickMark: "cross",
+    //     tickLblPos: "low",
     //   },
     // }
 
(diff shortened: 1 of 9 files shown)


productdevbook (commented):
Reviewed the diff end-to-end (read, write, clone, types, README, and the three new test suites).

What I checked:
- **Reader** (`parseAxisTickMark` / `parseAxisTickLblPos`): correctly trims the `val` attribute, validates against the OOXML enums, collapses each per-element default (`out` / `none` / `nextTo`) to `undefined`, and tolerates a missing `val`. The placement inside `parseAxisInfo` mirrors the existing `numberFormat` / `gridlines` / `scale` pattern exactly, so the "all-undefined → return undefined" guard is consistent.
- **Writer** (`buildAxisTickRendering`): the three children are emitted only when pinned, in the spec-required slot (after `<c:numFmt>`, before `<c:crossAx>`). I verified the wiring in both `buildBarAxes` (covers bar / column / line / area) and `buildScatterAxes`. Invalid enum tokens collapse to `undefined` at the `normalize*` step before reaching the XML builder, so a typo'd input cannot leak into OOXML.
- **Clone** (`applyTickMarkOverride` / `applyTickLblPosOverride`): mirrors the `null` (drop) / `undefined` (inherit) / value (replace) grammar that the other axis helpers use. The `out.x` / `out.y` predicates were extended to include the three new fields so an axis composed only of tick rendering still surfaces. Pie / doughnut still drop the entire `axes` block via the existing coercion check.
- **Types**: `ChartAxisTickMark` / `ChartAxisTickLabelPosition` are exported from the public surface; the JSDoc on each `axes.{x,y}.*TickMark` / `tickLblPos` documents the OOXML default. `CloneChartOptions.axes.{x,y}.*` correctly accepts `T | null`.
- **Test coverage** (33 new specs): defaults, every enum value, missing `val`, unknown tokens, single-emission guard, OOXML element ordering, every chart family, scatter axis layout (`axPos="b"` vs `"l"`), pie/doughnut drop-through, and a full `parseChart → cloneChart → writeChart → parseChart` plus end-to-end `writeXlsx` round-trip.

Fixes pushed in this branch:
- Merged `origin/main` to resolve the three test-file conflicts (main concurrently added `plotVisOnly` parse/write/clone suites). Both suites now coexist; ordering and braces were re-stitched manually around the overlapping `writeXlsx` / `ZipReader` boilerplate.
- `pnpm test` (lint + typecheck + vitest, 2988 tests) and `pnpm build` both pass locally on the merged tree.

No remaining concerns from my side. Pending CI green on the merged head before merge.

PR #214, opened 2026-05-03
feat(xlsx): chart category-axis tick skips — read, write, clone-through

productdevbook:
## Summary

Surfaces the per-axis `<c:tickLblSkip>` / `<c:tickMarkSkip>` elements at the read, write, and clone layers so a template's tick-thinning configuration survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

Both elements live exclusively on `CT_CatAx` / `CT_DateAx` in OOXML — they have no slot on `CT_ValAx`. They control how often Excel paints a tick label / mark along a crowded category axis: `1` (the OOXML default) shows every label, `2` shows every other one, and so on. Until now hucre's writer never emitted either element, so a template that pinned a date axis to "show every 7th label" flattened back to "show every label" on every parse -> clone -> write loop. This bridges another axis-configuration gap for the dashboard composition flow tracked in #136.

## API

```ts

// ── Read side ──
console.log(source.axes?.x?.tickLblSkip);  // 7 when the template pinned "show every 7th label",
                                           // undefined when the template used the default 1
console.log(source.axes?.x?.tickMarkSkip); // same shape for the tick-mark stride

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      series: [{ name: "Revenue", values: "B2:B62", categories: "A2:A62" }],
      axes: {
        x: {
          tickLblSkip: 5,  // show every 5th category label
          tickMarkSkip: 5, // and every 5th tick mark
        },
      },
    }],
  }],
});

// ── Clone-through ──
  axes: {
    x: {
      tickLblSkip: 3,    // replace
      tickMarkSkip: null, // drop the inherited stride (writer falls back to "1")
      // tickLblSkip / tickMarkSkip undefined → inherit the source's parsed value
    },
  },
});
```

## Model

```ts
interface ChartAxisInfo {
  /* ...existing fields... */
  tickLblSkip?: number;
  tickMarkSkip?: number;
}

interface SheetChart {
  axes?: {
    x?: { /* ... */; tickLblSkip?: number; tickMarkSkip?: number };
    y?: { /* ... */ };
  };
}

interface CloneChartOptions {
  axes?: {
    x?: { /* ... */; tickLblSkip?: number | null; tickMarkSkip?: number | null };
    y?: { /* ... */ };
  };
}
```

The read-side `ChartAxisInfo.{tickLblSkip,tickMarkSkip}` mirrors the same shape so a parsed value slots straight back into `cloneChart` without transformation. Only the X axis carries the fields on the writer / clone side because the OOXML schema places the elements on category axes only — surfacing them on `axes.y` would be misleading at the type level.

## Behavior

- **Read** — `parseChart` pulls the two elements off `<c:catAx>` only (the parser explicitly skips `<c:valAx>` so a corrupt template carrying a stray skip on a value axis cannot leak a field the writer would never emit). The OOXML default `1` collapses to `undefined` so absence and the default round-trip identically; out-of-range values (non-positive, > 32767, non-numeric) drop rather than fabricate a count Excel would reject.
- **Write** — Each element is emitted only when the matching field is pinned to a non-default value (no defaults injected), keeping fresh charts visually identical to Excel's reference serialization. The OOXML schema's strict child order is preserved: `... -> lblOffset -> tickLblSkip -> tickMarkSkip -> noMultiLvlLbl`. Out-of-range values drop silently (`0`, `-3`, `99999` produce no element); non-integer values round to the nearest integer; the schema boundaries `2` and `32767` are accepted literally.
- **Clone** — Each axis override accepts the standard `undefined` (inherit) / `null` (drop the inherited value, falling back to the writer's "show every tick" default) / value (replace) grammar that mirrors `gridlines` / `scale` / `numberFormat`. When the resolved clone target is `pie` / `doughnut` (no axes) or `scatter` (X axis is a value axis, not a category axis) the entire `axes` block is silently coerced so a stale `tickLblSkip` from a column template never leaks into a chart kind whose schema rejects it.

## Edge cases

- A chart with `tickLblSkip: 1` round-trips identically to a chart that omits the field — the writer omits the element on the second pass and the reader collapses both shapes to `undefined`.
- Values outside the `ST_SkipIntervals` band (`1..32767`) are filtered at every layer (read, write, clone) so a corrupt template cannot leak into the output.
- A `<c:tickLblSkip/>` with no `val` attribute reads as `undefined` rather than fabricate a stride.
- Bar / column / line / area all support tick skips on their category axis; scatter (whose X axis is a value axis) and pie / doughnut (no axes at all) silently drop the entire `axes` block on clone-through.
- Each element is placed in the spec-required slot inside `<c:catAx>` — after `<c:lblOffset>` and before `<c:noMultiLvlLbl>` — so Excel's strict validator accepts the file.

Refs #136

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm test` (lint + typecheck + vitest, all 2983 tests passing including 19 new tests across reader, writer, and clone for the two tick-skip elements)
- [x] Reader, writer, and clone tests cover: defaults, valid values, missing `val`, out-of-range values (drop rather than clamp), schema boundaries (2 / 32767), non-integer rounding, OOXML element ordering inside the catAx, single-emission guard, every chart family, scatter axis layout (value axes only — skips dropped), pie / doughnut drop-through, valAx never carries the elements, and a full `parseChart -> cloneChart -> writeChart -> parseChart` round-trip with a `writeXlsx` end-to-end check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -677,6 +677,15 @@ non-default that keeps hidden cells in the chart). The reader accepts
 the OOXML truthy / falsy spellings (`"1"` / `"true"` / `"0"` /
 `"false"`); unknown values and missing `val` attributes drop to
 `undefined`.
+`ChartAxisInfo.tickLblSkip` and `ChartAxisInfo.tickMarkSkip` surface
+the category-axis tick-thinning knobs (`<c:catAx><c:tickLblSkip val=".."/>`
+and `<c:catAx><c:tickMarkSkip val=".."/>`). Both elements live on
+`CT_CatAx` / `CT_DateAx` only — the reader skips the parse on
+`<c:valAx>` so a corrupt template carrying a stray skip on a value
+axis does not surface a field the writer would never emit anyway. The
+OOXML default `1` (show every label / mark) collapses to `undefined`;
+out-of-range values (non-positive or > 32767) drop rather than clamp
+so a malformed input cannot leak into the writer.
 `ChartSeriesInfo.smooth` surfaces the per-series
 `<c:ser><c:smooth val=".."/>` flag — Excel's "Format Data Series →
 Line → Smoothed line" toggle — only on `line` / `line3D` / `scatter`
@@ -838,6 +847,17 @@ the chart), matching Excel's reference serialization. Pin
 chart (`val="0"`). The writer always emits the element so the
 rendered intent is explicit on roundtrip — no chart family is special-
 cased.
+The `axes.x.tickLblSkip` and `axes.x.tickMarkSkip` fields thin out a
+crowded category axis (`<c:catAx><c:tickLblSkip val=".."/>` and
+`<c:catAx><c:tickMarkSkip val=".."/>`). Pass a positive integer to
+show every Nth label or mark; the OOXML default `1` (show every tick)
+is omitted from the rendered XML so untouched charts match Excel's
+reference serialization byte-for-byte. Out-of-range values
+(non-positive or > 32767) drop silently rather than clamp. Both
+fields live on category axes only — bar / column / line / area
+honour them; scatter (whose two axes are value axes) and pie /
+doughnut (no axes at all) silently ignore them. Non-integer inputs
+round to the nearest integer.
 For line and scatter charts, each `series[i].smooth` flag toggles
 Excel's curved-line variant (`<c:smooth val="..">` inside `<c:ser>`).
 Line series always emit the element — `smooth: true` writes `val="1"`,
@@ -991,6 +1011,15 @@ a `boolean` to replace it. Like `dispBlanksAs` and `varyColors`, the
 field lives on `<c:chart>` and is valid on every chart family, so a
 coercion (line → column, doughnut → pie, etc.) preserves the
 inherited value rather than dropping it.
+The per-axis `axes.x.tickLblSkip` and `axes.x.tickMarkSkip` overrides
+follow the same `undefined` (inherit) / `null` (drop) / number
+(replace) grammar as `gridlines` / `scale` / `numberFormat`. The
+inherited values are dropped silently when the resolved clone target
+is `scatter` (its X axis is a value axis, so the skip would have no
+slot in the rendered chart) and when the target is `pie` or
+`doughnut` (no axes at all) — flattening a column template into a
+scatter clone therefore never leaks a stale catAx skip into the
+output.
 
 #### Walking and adding charts with `getCharts` / `addChart`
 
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
## Review summary

Reviewed the patch end-to-end (`_types.ts`, `chart-reader.ts`, `chart-writer.ts`, `chart-clone.ts`, README, and the three new test suites).

### Correctness
- Reader correctly scopes `<c:tickLblSkip>` / `<c:tickMarkSkip>` to category axes via `axis.local === "catAx" || "dateAx"`. The `dateAx` arm is currently dead-code because `parseAxes` only iterates `catAx` / `valAx` children, but it's harmless defensive scaffolding for the day a `dateAx` consumer is wired up.
- Writer normalisation (`normalizeAxisSkip`) correctly drops `1`, non-finite, and out-of-range values (`<1`, `>32767`), and `Math.round`s non-integers — matches the docs.
- catAx emission is in the spec-required slot (`lblOffset` -> `tickLblSkip` -> `tickMarkSkip` -> `noMultiLvlLbl`). Scatter / pie / doughnut correctly never emit (no `<c:catAx>` rendered).
- Clone resolver follows the project's `undefined` (inherit) / `null` (drop) / value (replace) grammar consistently and now takes `type` to silently strip the skips when the resolved target is `scatter` (no catAx). Pie / doughnut were already short-circuited at the `cloneChart` call site.

### Tests
19 new tests cover read/write/clone defaults, valid values, missing `val`, out-of-range, schema boundaries (2 / 32767), non-integer rounding, single-emission, all chart families, scatter/pie/doughnut drop, valAx never carries the elements, full `parseChart -> cloneChart -> writeChart -> parseChart` round-trip, and a `writeXlsx` end-to-end check. All assertions are strict and meaningful.

### Fixes applied in this push
The branch was conflicting against `main` because the recently merged `majorTickMark` / `minorTickMark` / `tickLblPos` work touched the same axis blocks. Merged `main` in and:
- Appended the new fields after `tickLblPos` in both `SheetChart.axes.x` (`_types.ts`), `CloneChartOptions.axes.x` (`chart-clone.ts`), and `ChartAxisInfo` (`_types.ts`).
- Combined the parser's tick-rendering reads (major/minor/tickLblPos) with the new skip parser; preserved both element-absent guards in `parseAxisInfo`.
- Re-ordered the writer's `AxisRenderOptions` and `normalize*` block so the skip fields trail the tick-rendering fields without changing emission order in the catAx (`buildAxisTickRendering` after `numFmt`, `buildAxisSkips` after `lblOffset`).
- All 3023 vitest tests pass, lint/format/typecheck clean, `pnpm build` produces a clean `dist`.

### Remaining concerns
None blocking. CI is currently in progress on the merge commit; will merge once it goes green.

PR #215, opened 2026-05-03
feat(xlsx): chart axis reverse (orientation) — read, write, clone-through

productdevbook:
## Summary

Surfaces the per-axis `<c:scaling><c:orientation val="maxMin"/></c:scaling>` element at the read, write, and clone layers so a template's reverse-axis configuration survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:orientation>` is the OOXML control behind Excel's "Categories in reverse order" / "Values in reverse order" toggle. On a category axis, reversing flips the order in which categories are drawn (right-to-left on a column chart, top-to-bottom on a bar chart). On a value axis, reversing flips the numeric direction so the maximum sits at the origin and the minimum at the far end. Until now hucre's writer always pinned `"minMax"`, so a template that reversed its date axis (the canonical "most recent first" idiom on a horizontal bar chart) flattened back to forward orientation on every parse -> clone -> write loop. This bridges another axis-configuration gap for the dashboard composition flow tracked in #136.

## API

```ts

// ── Read side ──
console.log(source.axes?.y?.reverse);  // true when the template pinned "maxMin",
                                       // undefined when the template used the default "minMax"

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "bar",
      axes: {
        x: { reverse: true },  // categories drawn top-to-bottom (most recent first)
        y: { reverse: true },  // values drawn right-to-left
      },
    }],
  }],
});

// ── Clone-through ──
  axes: {
    y: {
      reverse: true,   // replace
      // reverse: null  // drop the inherited flag (writer falls back to forward "minMax")
      // reverse: undefined  // inherit the source's parsed value
    },
  },
});
```

## Model

```ts
interface ChartAxisInfo {
  /* ...existing fields... */
  reverse?: boolean;
}

interface SheetChart {
  axes?: {
    x?: { /* ... */; reverse?: boolean };
    y?: { /* ... */; reverse?: boolean };
  };
}

interface CloneChartOptions {
  axes?: {
    x?: { /* ... */; reverse?: boolean | null };
    y?: { /* ... */; reverse?: boolean | null };
  };
}
```

The read-side `ChartAxisInfo.reverse` mirrors the same shape so a parsed value slots straight back into `cloneChart` without transformation.

## Behavior

- **Read** — `parseChart` pulls the value off `<c:scaling><c:orientation val=".."/></c:scaling>` on every `CT_CatAx` / `CT_ValAx` / `CT_DateAx` / `CT_SerAx`. Only `"maxMin"` surfaces `true`; the OOXML default `"minMax"` (and unknown tokens, missing `val` attributes, missing `<c:orientation>`, missing `<c:scaling>`) all collapse to `undefined` so absence and the default round-trip identically.
- **Write** — `<c:orientation>` is always emitted because Excel requires it inside `<c:scaling>`. The writer pins `"maxMin"` only when `reverse === true` — a literal `false`, a non-boolean truthy value (e.g. `"yes"` leaking past the type guard), or an absent field all collapse to the forward `"minMax"` default. Each axis carries its own flag, so reversing X never propagates to Y. The OOXML schema's strict child order is preserved: `logBase` -> `orientation` -> `max` -> `min`.
- **Clone** — Each axis override accepts the standard `undefined` (inherit) / `null` (drop, writer falls back to forward `"minMax"`) / value (replace) grammar that mirrors `gridlines` / `scale` / `numberFormat`. A literal `false` override behaves identically to `null` / `undefined` because the OOXML default and an explicit `false` produce the same on-the-wire output. When the resolved clone target is `pie` / `doughnut` (no axes) the entire `axes` block is silently dropped.

## Edge cases

- A chart with `reverse: false` round-trips identically to a chart that omits the field — the writer emits `"minMax"` either way and the reader collapses both shapes to `undefined`.
- A typo'd `<c:orientation val="diagonal"/>` drops to `undefined` rather than fabricate a flag the writer would pick up.
- `<c:orientation/>` with no `val` attribute reads as `undefined`.
- A chart without any `<c:scaling>` element on the axis reads as `undefined`.
- Bar / column / line / area / scatter all support reverse on both axes; pie / doughnut silently drop the entire `axes` block since OOXML defines no axes for them.
- Mixing `reverse` with other axis fields (gridlines, scale, tick rendering) keeps every field independent — the `resolveAxes` merge does not drop either one when both source and override are populated.

Refs #136

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm test` (lint + typecheck + vitest, all 3019 tests passing including 31 new tests across reader, writer, and clone for the reverse-axis flag)
- [x] Reader, writer, and clone tests cover: defaults, `"maxMin"` value, missing `<c:scaling>` / `<c:orientation>` / `val`, unknown tokens, OOXML element ordering inside `<c:scaling>`, single-axis isolation (X reverse does not propagate to Y), every chart family, scatter axis layout (axPos="b" / "l"), pie / doughnut drop-through, and a full `parseChart -> cloneChart -> writeChart -> parseChart` round-trip with a `writeXlsx` end-to-end check.

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1068,6 +1068,21 @@ export interface SheetChart {
        * reference serialization. See {@link ChartAxisTickLabelPosition}.
        */
       tickLblPos?: ChartAxisTickLabelPosition;
+      /**
+       * Reverse the axis plotting order. Maps to
+       * `<c:scaling><c:orientation val="maxMin"/></c:scaling>` —
+       * Excel's "Categories in reverse order" / "Values in reverse
+       * order" toggle. Default: `false` (the OOXML `"minMax"` default).
+       *
+       * On a category axis, reversing flips the order in which
+       * categories are drawn (right-to-left on a column chart, top-to-
+       * bottom on a bar chart). On a value axis, reversing flips the
+       * numeric direction so the maximum sits at the origin and the
+       * minimum at the far end. Useful when porting templates that
+       * pin a specific reading direction (e.g. dates on a horizontal
+       * bar chart with the most recent at the top).
+       */
+      reverse?: boolean;
     };
     /** Value axis. */
     y?: {
@@ -1093,6 +1108,16 @@ export interface SheetChart {
        * `"nextTo"`. See {@link ChartAxisTickLabelPosition}.
        */
       tickLblPos?: ChartAxisTickLabelPosition;
+      /**
+       * Reverse the value axis plotting order. Maps to
+       * `<c:valAx><c:scaling><c:orientation val="maxMin"/></c:scaling></c:valAx>`.
+       * Default: `false` (the OOXML `"minMax"` default).
+       *
+       * Mirrors {@link SheetChart.axes.x.reverse} for the value axis —
+       * setting `true` flips the numeric direction so the maximum sits
+       * at the origin and the minimum at the far end.
+       */
+      reverse?: boolean;
     };
   };
 }
@@ -1967,6 +1992,17 @@ export interface ChartAxisInfo {
    * shape minimal. See {@link ChartAxisTickLabelPosition}.
    */
   tickLblPos?: ChartAxisTickLabelPosition;
+  /**
+   * Reverse-axis flag pulled from
+   * `<c:scaling><c:orientation val=".."/></c:scaling>`. Surfaces `true`
+   * only when the axis pinned `"maxMin"` (Excel's "Categories /
+   * Values in reverse order" toggle); the OOXML default `"minMax"`
+   * collapses to `undefined` so absence and the default round-trip
+   * identically through {@link cloneChart}. Mirrors the writer-side
+   * {@link SheetChart.axes.x.reverse} field, so a parsed value slots
+   * straight back into a clone target without transformation.
+   */
+  reverse?: boolean;
 }
 
 /**
--- a/src/xlsx/chart-clone.ts
+++ b/src/xlsx/chart-clone.ts
@@ -259,6 +259,13 @@ export interface CloneChartOptions {
        * replaces it.
        */
       tickLblPos?: ChartAxisTickLabelPosition | null;
+      /**
+       * Override the reverse-axis flag. `undefined` (or omitted)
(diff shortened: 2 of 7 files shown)


productdevbook (commented):
Reviewed the chart-axis reverse implementation across reader, writer, and clone layers.

## Strengths
- Clean three-layer surface (`ChartAxisInfo.reverse`, `SheetChart.axes.{x,y}.reverse`, `CloneChartOptions.axes.{x,y}.reverse`) that mirrors the established grammar of `gridlines` / `scale` / `numberFormat` / `tickLblPos`.
- Strict default-collapse policy in the reader: only an explicit `"maxMin"` surfaces `true`; `"minMax"`, missing `<c:scaling>`, missing `<c:orientation>`, missing `val`, and unknown tokens all collapse to `undefined` — absence and the OOXML default round-trip identically.
- Writer correctly pins `"maxMin"` only when `reverse === true` (the `"yes"`-leak test pins this guard) and preserves the strict OOXML CT_Scaling child order (`logBase` → `orientation` → `max` → `min`).
- 31 new tests cover defaults, both axes, scatter axPos isolation, pie/doughnut drop-through, and a full `parseChart -> cloneChart -> writeChart -> writeXlsx` end-to-end round-trip.

## Fixes I applied
- Resolved merge conflicts with `main` (PR #214 added `tickLblSkip` / `tickMarkSkip` on the same axis records). Both feature blocks now coexist side-by-side in `_types.ts`, `chart-reader.ts`, `chart-writer.ts`, `chart-clone.ts`, and across all three test files.
- Extended the README to document the new `reverse` field on the reader, writer, and clone summaries (mirroring how `tickLblSkip` / `tickMarkSkip` are documented).
- Added `reverse` to the per-axis enumerated field list in the writer summary.

## Verification
- `pnpm test` — 3054 tests pass (lint + format + typecheck + vitest).
- `pnpm build` — clean dist.
- No remaining concerns; pending CI on the merge commit.

PR #216, opened 2026-05-03
feat(xlsx): chart roundedCorners frame toggle — read, write, clone-through

productdevbook:
## Summary

Surfaces the `<c:chartSpace><c:roundedCorners val=".."/>` element at the read, write, and clone layers so a template's rounded chart frame survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:roundedCorners>` is the OOXML control behind Excel's "Format Chart Area -> Border -> Rounded corners" toggle. The element sits on `<c:chartSpace>` (a sibling of `<c:chart>`, not a child) because it styles the outer chart frame rather than the plot area. Until now hucre's writer always pinned `val="0"`, so a template that authored a rounded frame flattened back to square corners on every parse -> clone -> write loop. This bridges another chart-frame gap for the dashboard composition flow tracked in #136.

## API

```ts

// ── Read side ──
console.log(source.roundedCorners);  // true when the template pinned val="1",
                                     // undefined when the template used the default val="0"

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      roundedCorners: true,  // soften the chart frame's outer edge
    }],
  }],
});

// ── Clone-through ──
  roundedCorners: false,  // replace
  // roundedCorners: null  // drop the inherited flag (writer falls back to square val="0")
  // roundedCorners: undefined  // inherit the source's parsed value
});
```

## Model

```ts
interface Chart {
  /* ...existing fields... */
  roundedCorners?: boolean;
}

interface SheetChart {
  /* ...existing fields... */
  roundedCorners?: boolean;
}

interface CloneChartOptions {
  /* ...existing fields... */
  roundedCorners?: boolean | null;
}
```

The read-side `Chart.roundedCorners` mirrors the write-side `SheetChart.roundedCorners` so a parsed value slots straight back into `cloneChart` without transformation. The override grammar mirrors `plotVisOnly` / `varyColors` so the chart-frame toggles compose the same way at the call site.

## Behavior

- **Read** — `parseChart` pulls the value off `<c:chartSpace><c:roundedCorners val=".."/>`. Only `"1"` / `"true"` surface `true`; the OOXML default `"0"` / `"false"` (and unknown tokens, missing `val` attributes, missing `<c:roundedCorners>`) all collapse to `undefined` so absence and the default round-trip identically.
- **Write** — `<c:roundedCorners>` is always emitted because Excel's reference serialization includes it on every chart. The writer pins `val="1"` only when `roundedCorners === true` — a literal `false`, an absent field, or a non-boolean value all collapse to the default `val="0"`. The element is the first child of `<c:chartSpace>` per the `CT_ChartSpace` schema sequence, sitting before `<c:chart>`.
- **Clone** — Each override accepts the standard `undefined` (inherit) / `null` (drop, writer falls back to `val="0"`) / `boolean` (replace) grammar that mirrors `plotVisOnly` / `varyColors`. The toggle has no chart-family restriction, so the field carries through every coercion (line -> column, line -> doughnut, etc.) — every authored chart family has a `<c:chartSpace>` wrapper.

## Edge cases

- An explicit `roundedCorners: false` collapses to the OOXML default on re-parse (since absence and `val="0"` are indistinguishable on the read side). This matches how `plotVisOnly: true` / `varyColors` collapse on read.
- Unknown / non-boolean values fed to the writer fall back to `val="0"` rather than fabricate an attribute Excel rejects.

## Test plan

- [x] `pnpm test` passes (3014 tests, including 24 new tests across reader, writer, and clone)
- [x] `parseChart` surfaces `true` only on `<c:roundedCorners val="1"/>` and collapses `val="0"` / absence / unknown / missing attribute to `undefined`
- [x] `writeChart` always emits `<c:roundedCorners>` once on `<c:chartSpace>` before `<c:chart>`, threading the field through every authored family
- [x] `cloneChart` honors inherit / `null` (drop) / value (replace) on the override; carries the value through chart-family coercions; round-trips `true` through `writeXlsx` -> `parseChart`

Refs #136

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1013,6 +1013,22 @@ export interface SheetChart {
    */
   plotVisOnly?: boolean;
   /**
+   * Whether the chart frame is drawn with rounded corners. Maps to
+   * `<c:roundedCorners val=".."/>` on `<c:chartSpace>` (a sibling of
+   * `<c:chart>`, not a child). Mirrors Excel's "Format Chart Area →
+   * Border → Rounded corners" toggle.
+   *
+   * Default: `false` — the OOXML schema default and what every fresh
+   * Excel chart emits. Set `true` to soften the chart frame's outer
+   * edge, useful when matching a dashboard whose other charts already
+   * carry the rounded look from a template.
+   *
+   * The writer always emits the element so the rendered intent is
+   * explicit on roundtrip — Excel itself includes it in every reference
+   * serialization.
+   */
+  roundedCorners?: boolean;
+  /**
    * Per-axis configuration rendered alongside the plot area. The `x`
    * axis is the category axis for bar/column/line/area (or the bottom
    * value axis for scatter); the `y` axis is the value axis. Ignored
@@ -2178,6 +2194,24 @@ export interface Chart {
    * drop to `undefined`.
    */
   plotVisOnly?: boolean;
+  /**
+   * Rounded-corners flag pulled from
+   * `<c:chartSpace><c:roundedCorners val=".."/>`. Reflects Excel's
+   * "Format Chart Area → Border → Rounded corners" toggle, which paints
+   * the chart frame with rounded edges instead of the default square
+   * border.
+   *
+   * The OOXML default `false` collapses to `undefined` so absence and
+   * the default round-trip identically through {@link cloneChart} —
+   * only an explicit `<c:roundedCorners val="1"/>` surfaces `true`.
+   * The reader accepts the OOXML truthy / falsy spellings (`"1"` /
+   * `"true"` / `"0"` / `"false"`); unknown values and missing `val`
+   * attributes drop to `undefined`.
+   *
+   * Note: `<c:roundedCorners>` lives on `<c:chartSpace>`, not inside
+   * `<c:chart>` — the toggle styles the outer frame, not the plot area.
+   */
+  roundedCorners?: boolean;
 }
 
 // ── Workbook ───────────────────────────────────────────────────────
--- a/src/xlsx/chart-clone.ts
+++ b/src/xlsx/chart-clone.ts
@@ -208,6 +208,20 @@ export interface CloneChartOptions {
    */
   plotVisOnly?: boolean | null;
   /**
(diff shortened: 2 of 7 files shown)


productdevbook (commented):
Reviewed the chart-frame `roundedCorners` implementation across reader, writer, and clone layers.

## Strengths
- Clean three-layer surface (`Chart.roundedCorners`, `SheetChart.roundedCorners`, `CloneChartOptions.roundedCorners`) that mirrors the established grammar of `plotVisOnly` / `varyColors`.
- Strict default-collapse policy in the reader: only an explicit `<c:roundedCorners val="1"/>` surfaces `true`; `val="0"`, `"false"`, missing `<c:roundedCorners>`, missing `val`, and unknown tokens all collapse to `undefined` — absence and the OOXML default round-trip identically.
- Writer correctly pins `val="1"` only when `roundedCorners === true`, places the element first in `<c:chartSpace>` per the `CT_ChartSpace` schema (before `<c:chart>`), and emits it exactly once.
- Correct OOXML placement: the element sits on `<c:chartSpace>` (a sibling of `<c:chart>`), not inside `<c:chart>` — matches Excel's reference serialization.
- 24 new tests cover defaults, both spellings (`"1"` / `"true"`), unknown / missing-attr drops, every chart family, the `<c:chart>` ordering invariant, and a full `parseChart -> cloneChart -> writeChart -> writeXlsx` end-to-end round-trip.

## Fixes I applied
- Resolved merge conflicts with `main` (PRs #214 and #215 added `tickLblSkip` / `tickMarkSkip` and axis `reverse` to the same test files). The roundedCorners describe blocks now coexist alongside the axis reverse and tick-skip blocks across `test/chart-clone.test.ts`, `test/charts-write.test.ts`, and `test/charts.test.ts`.
- Extended the README to document the new `roundedCorners` field on the reader, writer, and clone summaries (mirroring how `plotVisOnly` is documented).

## Verification
- `pnpm test` — 3080 tests pass (lint + format + typecheck + vitest).
- `pnpm build` — clean dist.
- No remaining concerns; pending CI on the merge commit.

PR #217, opened 2026-05-03
feat(xlsx): chart axis hidden flag — read, write, clone-through

productdevbook:
## Summary

Surfaces the per-axis `<c:delete val=".."/>` element at the read, write, and clone layers so a template's hidden-axis configuration survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:delete>` is the OOXML control behind Excel's "Format Axis -> Axis Options -> Labels -> Show axis" toggle (and the matching "Delete" context-menu action). When pinned to `val="1"` it collapses the axis line, tick marks, and tick labels off the rendered chart — handy for sparkline-style dashboard tiles where only the data series should remain visible. Up until now hucre's writer always pinned `val="0"`, so a template that hid its value axis flattened back to a visible axis on every parse -> clone -> write loop. This bridges another axis-configuration gap for the dashboard composition flow tracked in #136.

## API

```ts

// ── Read side ──
console.log(source.axes?.x?.hidden);  // true when the template pinned val="1",
                                      // undefined when the template used the default val="0"

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      axes: {
        x: { hidden: true },  // hide the category axis (sparkline look)
        y: { hidden: true },  // hide the value axis too
      },
    }],
  }],
});

// ── Clone-through ──
  axes: {
    x: {
      hidden: true,    // replace
      // hidden: null   // drop the inherited flag (writer falls back to val="0")
      // hidden: false  // identical to null — both collapse to the OOXML default
      // hidden: undefined  // inherit the source's parsed value
    },
  },
});
```

## Model

```ts
interface ChartAxisInfo {
  /* ...existing fields... */
  hidden?: boolean;
}

interface SheetChart {
  axes?: {
    x?: { /* ... */; hidden?: boolean };
    y?: { /* ... */; hidden?: boolean };
  };
}

interface CloneChartOptions {
  axes?: {
    x?: { /* ... */; hidden?: boolean | null };
    y?: { /* ... */; hidden?: boolean | null };
  };
}
```

The read-side `ChartAxisInfo.hidden` mirrors the write-side `SheetChart.axes.[x|y].hidden` so a parsed value slots straight back into `cloneChart` without transformation.

## Behavior

- **Read** — `parseChart` pulls the value off `<c:delete val=".."/>` on every `CT_CatAx` / `CT_ValAx` / `CT_DateAx` / `CT_SerAx`. Only `"1"` / `"true"` surface `true`; the OOXML default `"0"` / `"false"` (and unknown tokens, missing `val` attributes, missing `<c:delete>`) all collapse to `undefined` so absence and the default round-trip identically.
- **Write** — `<c:delete>` is always emitted because Excel's reference serialization includes it on every axis. The writer pins `val="1"` only when `hidden === true` — a literal `false`, an absent field, or a non-boolean value all collapse to the default `val="0"`. The element sits between `<c:scaling>` and `<c:axPos>` per the strict CT_CatAx / CT_ValAx schema sequence.
- **Clone** — Each axis override accepts the standard `undefined` (inherit) / `null` (drop, writer falls back to `val="0"`) / `boolean` (replace) grammar that mirrors `gridlines` / `scale` / `numberFormat`. An override of `false` collapses identically to `null` because the writer treats both as the OOXML default. The flag carries through every chart-family coercion that has axes (line -> column, bar -> scatter, etc.); pie / doughnut clones drop the entire `axes` block because OOXML defines no axes for them.

## Edge cases

- An explicit `hidden: false` round-trips to `undefined` (since absence and `val="0"` are indistinguishable on the read side). This matches how `plotVisOnly: true` / `varyColors` collapse on read.
- Non-boolean inputs fed to the writer fall back to `val="0"` rather than fabricate an attribute Excel rejects.
- Each axis carries its own flag, so hiding X never propagates to Y (and vice versa).
- Bar / column / line / area / scatter all support the flag on both axes; pie / doughnut silently drop the entire `axes` block since OOXML defines no axes for them.

Refs #136

## Test plan

- [x] `pnpm test` passes (3064 tests, including 50 new tests across reader, writer, and clone for the hidden flag)
- [x] Reader, writer, and clone tests cover: defaults, `val="1"` / `val="true"` / `val="0"` / `val="false"`, missing `<c:delete>` / `val`, unknown tokens, OOXML element ordering inside the axis, single-axis isolation (X hidden does not propagate to Y), every chart family, scatter axis layout (axPos="b" / "l"), pie / doughnut drop-through, chart-family coercions (line -> column, bar -> scatter), and a full `parseChart -> cloneChart -> writeChart -> parseChart` round-trip with a `writeXlsx` end-to-end check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -698,6 +698,13 @@ axis does not surface a field the writer would never emit anyway. The
 OOXML default `1` (show every label / mark) collapses to `undefined`;
 out-of-range values (non-positive or > 32767) drop rather than clamp
 so a malformed input cannot leak into the writer.
+`ChartAxisInfo.hidden` surfaces the per-axis `<c:delete val=".."/>`
+flag — Excel's "Format Axis -> Show axis" toggle. Only an explicit
+`val="1"` (axis hidden) surfaces `true`; the OOXML default `val="0"`,
+absence, missing `val`, and unknown tokens all collapse to `undefined`
+so absence and the default round-trip identically. The reader accepts
+the OOXML truthy / falsy spellings (`"1"` / `"true"` / `"0"` /
+`"false"`).
 `ChartSeriesInfo.smooth` surfaces the per-series
 `<c:ser><c:smooth val=".."/>` flag — Excel's "Format Data Series →
 Line → Smoothed line" toggle — only on `line` / `line3D` / `scatter`
@@ -883,6 +890,16 @@ fields live on category axes only — bar / column / line / area
 honour them; scatter (whose two axes are value axes) and pie /
 doughnut (no axes at all) silently ignore them. Non-integer inputs
 round to the nearest integer.
+The per-axis `axes.x.hidden` and `axes.y.hidden` flags toggle
+`<c:catAx><c:delete val=".."/>` / `<c:valAx><c:delete val=".."/>` —
+Excel's "Format Axis -> Show axis" toggle. Set `true` to collapse the
+axis line, tick marks, and tick labels off the rendered chart (handy
+for sparkline-style dashboard tiles). The writer always emits the
+element because Excel's reference serialization includes
+`<c:delete val="0"/>` on every axis; `false` and absence both produce
+that default, while only an explicit `true` emits `val="1"`. The flag
+threads through bar / column / line / area / scatter; pie / doughnut
+silently ignore it because OOXML defines no axes for those families.
 For line and scatter charts, each `series[i].smooth` flag toggles
 Excel's curved-line variant (`<c:smooth val="..">` inside `<c:ser>`).
 Line series always emit the element — `smooth: true` writes `val="1"`,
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
Reviewed the chart axis hidden flag end to end (reader, writer, clone, and all 50 new tests). The implementation matches the conventions established by recent axis PRs:

- **Reader** correctly collapses absence and the OOXML default `val="0"` to `undefined` and only surfaces `true` for `val="1"` / `"true"`, mirroring `parsePlotVisOnly` and the rest of the boolean axis parsers.
- **Writer** always emits `<c:delete>` so untouched charts stay byte-stable with Excel's reference serialization, while `normalizeAxisHidden` defends against non-boolean leaks the same way `roundedCorners` / `plotVisOnly` do. OOXML element ordering (`<c:axId>` -> `<c:scaling>` -> `<c:delete>` -> `<c:axPos>`) is preserved on both `<c:catAx>` and `<c:valAx>` (including scatter's two-`<c:valAx>` layout).
- **Clone** uses the same `undefined` (inherit) / `null` (drop) / `boolean` (replace) grammar as the other axis overrides, with `false` collapsing to `undefined` to match how absence and the default round-trip — consistent with the `tickLblSkip` / `gridlines` override helpers.
- **Pie / doughnut** correctly drop the entire `axes` block, so a coerced clone never leaks a stale flag.

Also resolved the merge conflict against `main` (chart axis `reverse` had landed in parallel and touched the exact same axis-builder lines and override grammar) — kept both knobs side-by-side, ran `pnpm test` (3121 passed), `pnpm build` clean.

CI is green on the merge commit. Merging.

PR #218, opened 2026-05-03
feat(xlsx): chart category-axis label offset — read, write, clone-through

productdevbook:
## Summary

Surfaces the per-axis `<c:lblOffset>` element at the read, write, and clone layers so a template's label-offset configuration survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:lblOffset>` controls the distance between the tick labels and the axis line on a category axis, expressed as a percentage (0..1000) of Excel's default spacing. Up until now hucre's writer always pinned `val="100"`, so a template that pinned a wider gap (e.g. `val="250"`) flattened back to the default on every parse -> clone -> write loop. The element lives exclusively on `CT_CatAx` / `CT_DateAx` in OOXML — it has no slot on `CT_ValAx` or `CT_SerAx`. This bridges another axis-configuration gap for the dashboard composition flow tracked in #136.

## API

```ts

// ── Read side ──
console.log(source.axes?.x?.lblOffset);  // 250 when the template pinned val="250",
                                         // undefined when the template used the default 100

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      axes: {
        x: { lblOffset: 250 },  // push category labels twice as far from the axis
      },
    }],
  }],
});

// ── Clone-through ──
  axes: {
    x: {
      lblOffset: 200,    // replace
      // lblOffset: null   // drop the inherited offset (writer falls back to 100)
      // lblOffset: undefined  // inherit the source's parsed value
    },
  },
});
```

## Model

```ts
interface ChartAxisInfo {
  /* ...existing fields... */
  lblOffset?: number;
}

interface SheetChart {
  axes?: {
    x?: { /* ... */; lblOffset?: number };
  };
}

interface CloneChartOptions {
  axes?: {
    x?: { /* ... */; lblOffset?: number | null };
  };
}
```

The read-side `ChartAxisInfo.lblOffset` mirrors the write-side `SheetChart.axes.x.lblOffset` so a parsed value slots straight back into `cloneChart` without transformation. Same scope as the existing `tickLblSkip` / `tickMarkSkip` knobs — the field surfaces only on `axes.x` because OOXML places `<c:lblOffset>` exclusively on category axes.

## Behavior

- **Read** — `parseChart` pulls the value off `<c:lblOffset val=".."/>` on every `CT_CatAx` / `CT_DateAx`. The OOXML default `100` collapses to `undefined` so absence and the default round-trip identically; out-of-range values (negative or > 1000) drop rather than fabricate. The reader skips the parse on `<c:valAx>` so a corrupt template carrying a stray offset on a value axis does not surface a field the writer would never emit anyway.
- **Write** — `<c:lblOffset>` is always emitted because Excel's reference serialization includes it on every category axis. The writer pins the caller's override when set; absence (and the default `100`) emit `val="100"` so untouched charts match Excel's reference output byte-for-byte. Out-of-range values drop silently rather than clamp — the writer falls back to the default `100`. The element sits between `<c:lblAlgn>` and `<c:tickLblSkip>` per the strict CT_CatAx schema sequence.
- **Clone** — Each axis override accepts the standard `undefined` (inherit) / `null` (drop, writer falls back to `100`) / `number` (replace) grammar that mirrors `tickLblSkip` / `tickMarkSkip`. An override of `100` (the OOXML default) collapses identically to `null`. The flag carries through every chart-family coercion that has a category axis (line -> column, bar -> bar, etc.); scatter / pie / doughnut clones drop the entire `axes` block (or just the offset) because OOXML defines no category axis for them.

## Edge cases

- Out-of-range inputs (`-5`, `9999`, `NaN`) drop both at the parser and writer rather than clamping — a silent clamp would mask a configuration error.
- Non-integer values round to the nearest integer.
- The schema boundaries `0` and `1000` are accepted.
- The element is emitted exactly once per category axis, threaded through bar / column / line / area chart families.
- A scatter or pie / doughnut clone target silently drops an inherited `lblOffset` (its X axis is not a category axis), so flattening a column template into a scatter clone never leaks a stale catAx offset.

Closes Refs #136

## Test plan

- [x] `pnpm test` (lint + typecheck + 3110 vitest) passes.
- [x] `pnpm build` succeeds.
- [x] 30 new lblOffset tests across `charts.test.ts`, `charts-write.test.ts`, `chart-clone.test.ts` cover read, write, clone, end-to-end round-trip, and writeXlsx packaging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -704,63 +704,62 @@ tick skips above — `CT_CatAx` / `CT_DateAx` only — and the same
 default-collapse semantics: the OOXML default `100` (Excel's
 reference label spacing) collapses to `undefined` so absence and the
 default round-trip identically; out-of-range values (negative or
-
-> 1000. drop rather than clamp.
->       `ChartAxisInfo.reverse` surfaces the per-axis
->       `<c:scaling><c:orientation val="maxMin"/></c:scaling>` flag — Excel's
->       "Categories / Values in reverse order" toggle. Only `"maxMin"` surfaces
->       `true`; the OOXML default `"minMax"` (and unknown tokens, missing `val`
->       attributes, missing `<c:orientation>` / `<c:scaling>` elements) all
->       collapse to `undefined` so absence and the default round-trip
->       identically through `cloneChart`. Reverse can fire on either or both
->       axes independently — bar / column / line / area / scatter all support
->       it; pie / doughnut never report it because they have no axes.
->       `Chart.roundedCorners` surfaces the chart-frame
->       `<c:chartSpace><c:roundedCorners val=".."/>` flag — Excel's "Format
->       Chart Area → Border → Rounded corners" toggle. The element sits on
->       `<c:chartSpace>` (a sibling of `<c:chart>`) because it styles the
->       outer frame rather than the plot area. The OOXML default `false`
->       collapses to `undefined` so absence and `<c:roundedCorners val="0"/>`
->       round-trip identically; only an explicit `val="1"` surfaces `true`.
->       The reader accepts the OOXML truthy / falsy spellings (`"1"` / `"true"`
->       / `"0"` / `"false"`); unknown values and missing `val` attributes drop
->       to `undefined`.
->       `ChartSeriesInfo.smooth` surfaces the per-series
->       `<c:ser><c:smooth val=".."/>` flag — Excel's "Format Data Series →
->       Line → Smoothed line" toggle — only on `line` / `line3D` / `scatter`
->       series (the OOXML schema places `<c:smooth>` exclusively on
->       `CT_LineSer` and `CT_ScatterSer`). Absence and the OOXML default
->       `val="0"` both collapse to `undefined`, so only an explicit
->       `<c:smooth val="1"/>` round-trips as `smooth: true`.
->       `ChartSeriesInfo.marker` surfaces the per-series `<c:ser><c:marker>`
->       glyph configuration on `line` / `line3D` / `scatter` series (the
->       schema places `<c:marker>` only on `CT_LineSer` and `CT_ScatterSer`).
->       The reader pulls `symbol` (`circle` / `square` / `diamond` /
->       `triangle` / `x` / `star` / `dot` / `dash` / `plus` / `auto` /
->       `none`), `size` (clamped to the OOXML 2..72 band), and the marker's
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
## Review

The lblOffset implementation looks solid — the read / write / clone layers all follow the existing tickLblSkip / tickMarkSkip patterns symmetrically, and the OOXML schema constraints (`ST_LblOffsetPercent`, `0..1000`, default `100`) are enforced consistently in `parseAxisLblOffset`, `normalizeAxisLblOffset`, and `applyLblOffsetOverride`. The `<c:lblOffset>` element is emitted in the correct schema position (between `<c:lblAlgn>` and `<c:tickLblSkip>`), the writer always emits `val="100"` to match Excel's reference output byte-for-byte, and out-of-range inputs drop rather than clamp to surface programming errors. Test coverage is good: 30 new tests across read / write / clone / e2e roundtrip / writeXlsx packaging, including scope-restriction cases for scatter / pie / doughnut. Type-level scoping (`SheetChart.axes.x.lblOffset` only, not `axes.y`) correctly mirrors the OOXML schema since `<c:lblOffset>` lives only on `CT_CatAx` / `CT_DateAx`.

## Fixes pushed

1. **README markdown corruption**: the original commit's diff inserted a `> 1000)` line at column zero, which markdown parses as a blockquote start. That swallowed every following chart-overview paragraph (`reverse`, `roundedCorners`, `smooth`, `marker`, `stroke`, `invertIfNegative`, regenerated-sheets) into one indented quoted block. Rewrote the `lblOffset` paragraph to spell out "greater than 1000" so it can never trip markdown again, and restored the surrounding paragraphs to their normal prose form.
2. **Merge conflicts with #217 (axis hidden flag)**: both PRs add an axis-level field through the same plumbing (`_types.ts`, `chart-reader`, `chart-writer`, `chart-clone`, plus the three test files). Composed `lblOffset` and `hidden` side-by-side in every layer rather than picking one — they're orthogonal and share the same scope semantics. Test files were the trickiest because git interleaved the new describe blocks; split them back into two consecutive top-level blocks per file.

`pnpm test` (lint + typecheck + 3151 vitest) passes locally and `pnpm build` succeeds. Pending CI on the merge commit before merging.

PR #219, opened 2026-05-03
feat(xlsx): chart legend overlay flag — read, write, clone-through

productdevbook:
## Summary

Surfaces the legend-overlay element at the read, write, and clone layers so a template's overlay configuration survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:legend><c:overlay val=".."/></c:legend>` is the OOXML control behind Excel's "Format Legend -> Show the legend without overlapping the chart" toggle (the checkbox is the inverse of the flag — checked means `val=0`, unchecked means `val=1`). `val=1` draws the legend on top of the plot area so the chart series get the full frame; `val=0` reserves a slot for the legend and shrinks the plot area to make room.

Until now hucre's writer always pinned `val=0`, so a template that overlapped its legend flattened back to a side-anchored legend on every parse -> clone -> write loop. This bridges another chart-level gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.legendOverlay);  // true when the template pinned val="1",
                                    // undefined when the template used the default 0

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      legend: "right",
      legendOverlay: true,  // legend draws on top of the plot area
    }],
  }],
});

// -- Clone-through --
  legendOverlay: true,    // replace
  // legendOverlay: null  // drop the inherited overlay (writer falls back to 0)
  // legendOverlay: undefined // inherit the source's parsed value
});
```

## Model

```ts
interface SheetChart {
  /* ...existing fields... */
  legendOverlay?: boolean;
}

interface Chart {
  /* ...existing fields... */
  legendOverlay?: boolean;
}

interface CloneChartOptions {
  /* ...existing fields... */
  legendOverlay?: boolean | null;
}
```

The read-side `Chart.legendOverlay` mirrors the write-side `SheetChart.legendOverlay` so a parsed value slots straight back into `cloneChart` without transformation. Same shape as the existing `roundedCorners` / `plotVisOnly` toggles.

## Behavior

- **Read** — `parseChart` pulls the value off `<c:legend><c:overlay val=".."/></c:legend>`. The OOXML default `false` (`val="0"`) collapses to `undefined` so absence and the default round-trip identically; only an explicit `val="1"` surfaces `true`. Accepts the OOXML truthy / falsy spellings (`"1"` / `"true"` / `"0"` / `"false"`); unknown values and missing `val` attributes drop to `undefined`. The flag is dropped whenever the chart hides the legend (`<c:delete val="1"/>`) or omits the `<c:legend>` element entirely — there is no overlay slot to surface in either case.
- **Write** — `<c:overlay>` is always emitted inside `<c:legend>` because Excel's reference serialization includes it on every visible legend. The writer pins the caller's override when set; absence (and the default `false`) emit `val="0"` so untouched charts match Excel's reference output byte-for-byte. Non-boolean inputs collapse to `false` to keep the on-the-wire output stable. The element sits between `<c:legendPos>` and the rest of the legend's children per the strict CT_Legend schema sequence. When `legend === false` the writer skips the entire `<c:legend>` element so no orphan overlay can leak through.
- **Clone** — Each override accepts the standard `undefined` (inherit) / `null` (drop, writer falls back to `0`) / `boolean` (replace) grammar that mirrors `roundedCorners` / `plotVisOnly`. The inherited value is silently dropped from the cloned `SheetChart` whenever the resolved `legend` is `false` — a hidden legend has no overlay slot, so leaking the value would carry no on-screen effect. Re-enabling a hidden source legend through `legend: "top"` (or any visible position) on the override re-opens the slot, and an explicit `legendOverlay: true` override threads through.

## Edge cases

- A hidden source legend with `<c:overlay val="1"/>` parses to `legendOverlay: undefined` (since the overlay has no rendering effect on a deleted legend).
- An explicit `legendOverlay: true` override with `legend: false` collapses to `undefined` on the cloned `SheetChart`.
- Non-boolean `legendOverlay` inputs (e.g. truthy strings) collapse to the OOXML `false` default rather than fabricate a flag.
- The element is emitted exactly once per chart (the chart title also carries its own `<c:overlay>` element, scoped to the title block — they do not collide).
- Pie / doughnut / line / bar / column / area / scatter all support the flag identically because `<c:legend>` is a chart-level sibling of `<c:plotArea>`.

## Test plan

- [x] `pnpm test` (lint + typecheck + 3156 vitest) passes.
- [x] `pnpm build` succeeds.
- [x] 35 new legendOverlay tests across `charts.test.ts` (11 read), `charts-write.test.ts` (12 write + round-trip + writeXlsx packaging), `chart-clone.test.ts` (12 clone) cover read, write, clone, end-to-end round-trip, hidden-legend interaction, and per-family threading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -724,6 +724,18 @@ round-trip identically; only an explicit `val="1"` surfaces `true`.
 The reader accepts the OOXML truthy / falsy spellings (`"1"` / `"true"`
 / `"0"` / `"false"`); unknown values and missing `val` attributes drop
 to `undefined`.
+`Chart.legendOverlay` surfaces the legend-overlay flag pulled from
+`<c:legend><c:overlay val=".."/></c:legend>` — Excel's "Format Legend →
+Show the legend without overlapping the chart" toggle (the checkbox is
+the inverse of this flag — checked means `false`, unchecked means
+`true`). The OOXML default `false` collapses to `undefined` so absence
+and `<c:overlay val="0"/>` round-trip identically; only an explicit
+`val="1"` surfaces `true`. The reader accepts the OOXML truthy / falsy
+spellings (`"1"` / `"true"` / `"0"` / `"false"`); unknown values and
+missing `val` attributes drop to `undefined`. The flag is dropped
+whenever `Chart.legend` is `false` or the chart omits the legend
+element entirely — there is no overlay slot on a hidden legend, so the
+parsed shape stays minimal.
 `ChartSeriesInfo.smooth` surfaces the per-series
 `<c:ser><c:smooth val=".."/>` flag — Excel's "Format Data Series →
 Line → Smoothed line" toggle — only on `line` / `line3D` / `scatter`
@@ -908,6 +920,17 @@ outer edge (`val="1"`). The writer always emits the element so the
 rendered intent is explicit on roundtrip — no chart family is
 special-cased, since the toggle styles the outer wrapper rather than
 any chart-family-specific markup.
+The chart-level `legendOverlay` field maps to `<c:overlay val=".."/>`
+inside `<c:legend>` — Excel's "Format Legend → Show the legend without
+overlapping the chart" toggle (the checkbox is the inverse of this
+flag — checked means `false`, unchecked means `true`). Absent it, the
+writer emits the OOXML default `val="0"` (the legend reserves its own
+slot and the plot area shrinks to make room), matching Excel's
+reference serialization. Pin `legendOverlay: true` to draw the legend
+on top of the plot area so the chart series get the full frame
+(`val="1"`). The flag is silently ignored when `legend: false`
+suppresses the entire legend element — there is no overlay slot on a
+hidden legend, so the writer skips emitting any orphaned `<c:overlay>`.
 The `axes.x.tickLblSkip` and `axes.x.tickMarkSkip` fields thin out a
 crowded category axis (`<c:catAx><c:tickLblSkip val=".."/>` and
 `<c:catAx><c:tickMarkSkip val=".."/>`). Pass a positive integer to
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
Reviewed the diff end-to-end. The implementation is symmetrical and consistent with existing chart-level toggles (`roundedCorners`, `plotVisOnly`):

- **Reader**: `parseLegendOverlay` is correctly gated on a visible legend (the call site checks `legend !== undefined && legend !== false`), and `parseLegend` already returns `false` for `<c:delete val="1"/>`, so a hidden-legend overlay is dropped as documented. The OOXML default `val="0"` collapses to `undefined` for round-trip symmetry, and unknown / missing `val` values do too — matches the existing flag conventions.
- **Writer**: `<c:overlay>` is always emitted on a visible legend (matching Excel's reference serialization), placed after `<c:legendPos>`. The `legend === false` path skips the entire `<c:legend>` element so no orphan overlay leaks. `resolveLegendOverlay` collapses non-literal-true to `false`, matching how `roundedCorners` / `plotVisOnly` treat their inputs.
- **Clone**: `resolveLegendOverlay` follows the standard `undefined` (inherit) / `null` (drop) / `boolean` (replace) grammar. The clone gates on the resolved `legend` so an inherited or override `true` against a hidden legend is dropped, and re-enabling a hidden source legend through an override threads the overlay flag through.
- **Tests**: 35 new tests across read, write, clone, round-trip, hidden-legend, every chart family, and `writeXlsx` packaging give thorough coverage.
- **Docs**: README updates cover the read, write, and clone sides clearly with the inverse-checkbox note. No web docs reference the sibling toggles, so no further doc surface to update.

CI green, mergeable, no concerns. Merging.

PR #220, opened 2026-05-03
feat(xlsx): chart category-axis label alignment — read, write, clone-through

productdevbook:
## Summary

Surfaces the per-axis `<c:lblAlgn>` element at the read, write, and clone layers so a template's label-alignment configuration survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:lblAlgn>` controls the horizontal alignment of tick labels along a category axis (`"ctr"` / `"l"` / `"r"` per the OOXML `ST_LblAlgn` enumeration). The element lives exclusively on `CT_CatAx` / `CT_DateAx` in OOXML — `CT_ValAx` and `CT_SerAx` reject it entirely. Up until now hucre's writer always pinned `val="ctr"`, so a template that pinned a flush-left (`"l"`) or flush-right (`"r"`) alignment flattened back to centered on every parse -> clone -> write loop. This bridges another axis-configuration gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.axes?.x?.lblAlgn);  // "l" when the template pinned val="l",
                                        // undefined when the template used the default "ctr"

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      axes: {
        x: { lblAlgn: "l" },  // wrapped category labels align flush-left
      },
    }],
  }],
});

// -- Clone-through --
  axes: {
    x: {
      lblAlgn: "r",       // replace
      // lblAlgn: null    // drop the inherited value (writer falls back to "ctr")
      // lblAlgn: undefined  // inherit the source's parsed value
    },
  },
});
```

## Model

```ts
export type ChartAxisLabelAlign = "ctr" | "l" | "r";

interface ChartAxisInfo {
  /* ...existing fields... */
  lblAlgn?: ChartAxisLabelAlign;
}

interface SheetChart {
  axes?: {
    x?: { /* ... */; lblAlgn?: ChartAxisLabelAlign };
  };
}

interface CloneChartOptions {
  axes?: {
    x?: { /* ... */; lblAlgn?: ChartAxisLabelAlign | null };
  };
}
```

The read-side `ChartAxisInfo.lblAlgn` mirrors the write-side `SheetChart.axes.x.lblAlgn` so a parsed value slots straight back into `cloneChart` without transformation. Same scope as the existing `lblOffset` / `tickLblSkip` / `tickMarkSkip` knobs — the field surfaces only on `axes.x` because OOXML places `<c:lblAlgn>` exclusively on category axes.

## Behavior

- **Read** — `parseChart` pulls the value off `<c:lblAlgn val=".."/>` on every `CT_CatAx` / `CT_DateAx`. The OOXML default `"ctr"` collapses to `undefined` so absence and the default round-trip identically; unknown tokens (e.g. `"left"` / `"center"`) drop rather than fabricate. The reader skips the parse on `<c:valAx>` / `<c:serAx>` so a corrupt template carrying a stray alignment on a value axis does not surface a field the writer would never emit anyway.
- **Write** — `<c:lblAlgn>` is always emitted because Excel's reference serialization includes it on every category axis. The writer pins the caller's override when set; absence (and the default `"ctr"`) emit `val="ctr"` so untouched charts match Excel's reference output byte-for-byte. Unknown tokens drop silently rather than clamp — the writer falls back to the default `"ctr"`. The element sits between `<c:auto>` and `<c:lblOffset>` per the strict CT_CatAx schema sequence.
- **Clone** — Each axis override accepts the standard `undefined` (inherit) / `null` (drop, writer falls back to `"ctr"`) / value (replace) grammar that mirrors `lblOffset` / `tickLblSkip` / `tickMarkSkip`. An override of `"ctr"` (the OOXML default) collapses identically to `null`. The flag carries through every chart-family coercion that has a category axis (line -> column, bar -> bar, etc.); scatter / pie / doughnut clones drop the entire `axes` block (or just the alignment) because OOXML defines no category axis for them.

## Edge cases

- Unknown tokens (`"left"` / `"center"` / `"LEFT"` / empty string) drop both at the parser and writer rather than falling through to the default — a silent fall-through would mask a configuration error.
- Pinning `lblAlgn: "ctr"` (the OOXML default) collapses identically to omitting the field — the writer emits the default in both cases and the parser/cloner reduce to `undefined`.
- The element is emitted exactly once per chart and never on a value axis (scatter / pie / doughnut never see it).
- Bar / column / line / area all support the flag identically because `<c:catAx>` is shared across these families through the same `buildBarAxes` writer path.

## Test plan

- [x] `pnpm test` (lint + typecheck + 3182 vitest) passes.
- [x] `pnpm build` succeeds.
- [x] 33 new lblAlgn tests across `charts.test.ts` (7 read), `charts-write.test.ts` (13 write + round-trip + writeXlsx packaging), `chart-clone.test.ts` (10 clone + end-to-end) cover read, write, clone, end-to-end round-trip, scatter / pie scope, and per-family threading.

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1148,6 +1148,24 @@ export interface SheetChart {
        */
       lblOffset?: number;
       /**
+       * Horizontal alignment of the tick labels on a category axis —
+       * `"ctr"` (center, the OOXML default), `"l"` (left), or `"r"`
+       * (right). Maps to `<c:catAx><c:lblAlgn val=".."/></c:catAx>`.
+       * Useful when category labels are wrapped onto multiple lines
+       * and the default centered alignment looks ragged against a
+       * column chart's left-aligned bars. Excel's UI exposes the
+       * three presets under "Format Axis -> Alignment" on a category
+       * axis only.
+       *
+       * Only meaningful for bar / column / line / area charts (whose X
+       * axis is `<c:catAx>`); silently ignored for scatter (both axes
+       * are value axes) and pie / doughnut (no axes at all). The OOXML
+       * schema (`ST_LblAlgn`) restricts the value to the three tokens
+       * above; unknown tokens are dropped at write time. See
+       * {@link ChartAxisLabelAlign}.
+       */
+      lblAlgn?: ChartAxisLabelAlign;
+      /**
        * Hide the entire axis (line, tick marks, tick labels). Maps to
        * `<c:catAx><c:delete val="1"/></c:catAx>` (or the matching
        * `<c:valAx>` element on scatter). Default: `false` — Excel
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
## Review

Reviewed the patch end-to-end against the recently merged sibling axis-knob features (#214 tick skips, #217 hidden, #218 lblOffset, #219 legendOverlay). The implementation is consistent with the established pattern:

- `parseAxisLblAlgn` is gated behind `isCategoryAxis` and validates against the OOXML `ST_LblAlgn` enum, collapsing the default `"ctr"` and unknown tokens to `undefined` — symmetric with the writer's `normalizeAxisLblAlgn`.
- `<c:lblAlgn>` is emitted between `<c:auto>` and `<c:lblOffset>` per the strict `CT_CatAx` schema sequence; an order test pins this against future regressions.
- Scope is correctly category-axis-only: scatter / pie / doughnut never see the element on read, write, or clone-through.
- Clone helper `applyLblAlgnOverride` mirrors the `undefined` (inherit) / `null` (drop) / value (replace) grammar of the other axis helpers, with `"ctr"` collapsing identically to `null`.
- 33 new tests across the read / write / clone surface (plus an end-to-end `writeXlsx` round trip) — coverage matches the bar set by #218.

## Fixes pushed

- Merged `origin/main` to resolve conflicts caused by #219 (legendOverlay) landing parallel `describe` blocks in the three test files; both feature suites are preserved side by side.
- Extended `README.md` with read-side / write-side / clone-side docs for the new `axes.x.lblAlgn` knob, matching the surrounding entries for `lblOffset` / tick skips / `hidden`.

`pnpm test` (lint + typecheck + 3217 vitest) and `pnpm build` are green locally on the merge commit. Merging once CI clears.

PR #221, opened 2026-05-03
feat(xlsx): chart pie/doughnut series explosion — read, write, clone-through

productdevbook:
## Summary

Surfaces the per-series `<c:explosion>` element at the read, write, and clone layers so a template's pie / doughnut slice-explosion configuration survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:explosion>` controls how far each slice is pulled away from the chart center, expressed as a percentage of the radius. The element lives only on `CT_PieSer` (shared across `<c:pieChart>` / `<c:doughnutChart>` / `<c:pie3DChart>` / `<c:ofPieChart>` via `EG_PieSer`); every other chart family rejects it. Up until now hucre's writer never emitted the element and the parser ignored it, so a template's slice explosion flattened to zero on every parse -> clone -> write loop. This bridges another series-level gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.series?.[0].explosion); // 25  (template pulled the slice 25% out)

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "pie",
      series: [{
        values: "B2:B6",
        categories: "A2:A6",
        explosion: 25,   // pull every slice 25% from the center
      }],
    }],
  }],
});

// -- Clone-through --
  seriesOverrides: [
    { explosion: 50 },     // replace the inherited value
    // { explosion: null }   // drop the inherited value (writer falls back to 0)
    // { explosion: undefined } // inherit the source's parsed value
  ],
});
```

## Model

```ts
export interface ChartSeries {
  /* ...existing fields... */
  explosion?: number;
}

export interface ChartSeriesInfo {
  /* ...existing fields... */
  explosion?: number;
}

export interface CloneChartSeriesOverride {
  /* ...existing fields... */
  explosion?: number | null;
}
```

The read-side `ChartSeriesInfo.explosion` mirrors the write-side `ChartSeries.explosion` so a parsed value slots straight into `cloneChart` without transformation. Same scope as the existing `invertIfNegative` (#209) / `smooth` (#205) per-series knobs — the field surfaces only on pie-family series because OOXML places `<c:explosion>` exclusively on `CT_PieSer`.

## Behavior

- **Read** — `parseExplosion` pulls the value off `<c:ser><c:explosion val=".."/>` on every pie-family series (`pie`, `pie3D`, `doughnut`, `ofPie`). The OOXML default `0` collapses to `undefined` so absence and `0` round-trip identically; negative or non-finite values drop rather than fabricate. Non-integer input rounds for parity with the writer (Excel's UI accepts integer percentages only). The parser ignores stray `<c:explosion>` elements on bar / line / area / scatter templates so a corrupt source does not surface a value the writer would never emit anyway.
- **Write** — `<c:explosion>` is only emitted when the resolved value is `> 0` — `0` matches the OOXML default and absence round-trips identically. A new `clampExplosion` helper rounds non-integer input, drops negatives and non-finites, and clamps the upper bound to 400% (Excel's UI band). The element sits between `<c:spPr>` and `<c:dPt>` / `<c:dLbls>` per the strict CT_PieSer schema sequence. Non-pie / non-doughnut writers never thread the field, so a stray `explosion` on a bar / line / area / scatter series is silently dropped.
- **Clone** — Each series override accepts the standard `undefined` (inherit) / `null` (drop, writer falls back to `0`) / `number` (replace) grammar that mirrors `invertIfNegative` / `smooth`. An override of `0` (the OOXML default) collapses identically to `null`. The field carries through pie -> doughnut and doughnut -> pie flattens (the element lives on both kinds, so the explosion survives); coercion into any other family (`bar`, `column`, `line`, `area`, `scatter`) drops the inherited value silently because OOXML defines no `<c:explosion>` slot for them.

## Edge cases

- `<c:explosion val="0"/>` collapses to `undefined` on read.
- Negative, non-finite, or malformed explosion values drop both at the parser and writer rather than silently substituting `0` — a fall-through would mask a configuration error.
- Pinning `explosion: 0` collapses identically to omitting the field — the writer elides the element in both cases.
- The element is emitted exactly once per series and never on a non-pie / non-doughnut chart family.
- Out-of-band values clamp on write (`9999` -> `400`); the writer's upper bound matches Excel's UI ceiling so a round-trip stays inside the range Excel will render.
- Non-integer input rounds (`33.6` -> `34`) for parity between the parser and writer.
- `pie3DChart` and `ofPieChart` series surface explosion at parse time (CT_Pie3DSer / CT_OfPieSer share CT_PieSer through EG_PieSer); the writer maps both kinds onto the 2D `pie` writer so the element survives the round-trip.
- A doughnut template flattened to pie keeps its explosion (and vice versa); coercion into a non-pie family drops it.

## Test plan

- [x] `pnpm test` (lint + typecheck + 3218 vitest) passes.
- [x] `pnpm build` succeeds.
- [x] 31 new explosion tests across `charts.test.ts` (9 read), `charts-write.test.ts` (12 write + round-trip + writeXlsx packaging), `chart-clone.test.ts` (10 clone + end-to-end) cover read, write, clone, end-to-end round-trip, multi-series independence, doughnut <-> pie flatten, and per-family scope.

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -836,6 +836,22 @@ export interface ChartSeries {
    * the inverted color when the spreadsheet supplies one).
    */
   invertIfNegative?: boolean;
+  /**
+   * Pie / doughnut slice explosion as a percentage of the radius —
+   * the distance the slice is pulled away from the center. Maps to
+   * `<c:explosion val=".."/>` inside the `<c:ser>` element. Only
+   * meaningful for `pie` and `doughnut` charts — the OOXML schema
+   * places `<c:explosion>` exclusively on `CT_PieSer`, so the field
+   * is silently dropped on every other chart family at write time.
+   *
+   * Default: `0` (slices flush against each other). Excel's UI
+   * exposes 0–400% under "Format Data Point → Series Options → Pie
+   * Explosion"; values outside that band are clamped on write so a
+   * round-trip stays inside the range Excel will render. Per-data-point
+   * explosion (one slice pulled away while the rest stay flush) is not
+   * yet supported — the field applies to every slice in the series.
+   */
+  explosion?: number;
 }
 
 /**
@@ -1900,6 +1916,16 @@ export interface ChartSeriesInfo {
    * default and round-trips identically with absence of the field.
    */
   invertIfNegative?: boolean;
+  /**
+   * Slice explosion (in percent of the radius) pulled from
+   * `<c:ser><c:explosion val=".."/>`. Surfaces only on `pie`,
+   * `pie3D`, `doughnut`, and `ofPie` series — the OOXML schema
+   * places `<c:explosion>` exclusively on `CT_PieSer` (which is
+   * shared across the pie family via `EG_PieSer`). The OOXML
+   * default `0` collapses to `undefined` because absence and `0`
+   * round-trip identically through the writer's elision logic.
+   */
+  explosion?: number;
 }
 
 /**
--- a/src/xlsx/chart-clone.ts
+++ b/src/xlsx/chart-clone.ts
@@ -93,6 +93,16 @@ export interface CloneChartSeriesOverride {
    * resolved chart type is anything else.
    */
   invertIfNegative?: boolean | null;
+  /**
+   * Slice-explosion override (in percent of the radius). `undefined`
+   * (or omitted) inherits the source series' `explosion`; `null` drops
+   * the inherited value (the cloned series falls back to the OOXML
+   * default `0`); a finite `number` replaces it wholesale (clamped to
+   * the 0..400% band Excel's UI exposes; `0` collapses to absence).
+   * Only meaningful for `pie` and `doughnut` clones — silently dropped
+   * from the output when the resolved chart type is anything else.
+   */
+  explosion?: number | null;
 }
 
 /**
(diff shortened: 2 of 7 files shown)


productdevbook (commented):
Reviewed the diff end-to-end: read (parseExplosion), write (clampExplosion + emit between c:spPr and c:cat/c:val per CT_PieSer), and clone (resolveExplosion with the standard undefined/null/number grammar) all line up cleanly.

Strengths:
- Symmetry with the 0-collapse rule on both parser and writer keeps absence and the OOXML default round-tripping identically.
- Per-family scoping is enforced at three layers (parser kind guard, writer options threading, clone post-strip after type coercion) which is exactly the right defense-in-depth.
- The 0..400 clamp matches Excel's UI band, mirroring the holeSize precedent.
- Element ordering inside <c:ser> (spPr -> explosion -> dLbls -> cat -> val) matches CT_PieSer's strict sequence.
- 31 new tests cover read, write, clone, multi-series independence, doughnut <-> pie flatten, pie3D/ofPie surfacing, and an end-to-end writeXlsx round-trip — same bar set by the recent invertIfNegative (#209) and smooth (#205) PRs.

Consistent with the project's pattern for per-series knobs. CI is green and mergeable. LGTM, merging.

PR #222, opened 2026-05-03
feat(xlsx): chart category-axis noMultiLvlLbl flag — read, write, clone-through

productdevbook:
## Summary

Surfaces the per-axis `<c:noMultiLvlLbl>` element at the read, write, and clone layers so a template's multi-level-label suppression survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:noMultiLvlLbl>` lives exclusively on `CT_CatAx` in OOXML — even `<c:dateAx>`, `<c:valAx>`, and `<c:serAx>` reject it. It mirrors Excel's "Format Axis -> Multi-level Category Labels" checkbox (the checkbox is the inverse — checked means tiered labels, i.e. `noMultiLvlLbl: false`). When a category range spans multiple columns / rows Excel groups the labels into tiers; setting `noMultiLvlLbl: true` flattens every category onto a single line regardless of the source range's shape. Up until now hucre's writer always pinned `val="0"`, so a template that pinned `val="1"` flattened back to the default on every parse -> clone -> write loop. This bridges another category-axis-configuration gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.axes?.x?.noMultiLvlLbl); // true (template flattened multi-tier labels)

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      series: [{ values: "B2:B10", categories: "A2:B10" }], // 2-column range -> Excel would tier
      axes: { x: { noMultiLvlLbl: true } },                  // flatten back to one line
    }],
  }],
});

// -- Clone-through --
  axes: {
    x: { noMultiLvlLbl: true },     // override the inherited value
    // x: { noMultiLvlLbl: null },  // drop the inherited value (writer falls back to false)
    // x: { noMultiLvlLbl: undefined } // inherit the source's parsed value
  },
});
```

## Model

```ts
export interface SheetChart {
  // ...existing fields...
  axes?: {
    x?: {
      // ...existing fields...
      noMultiLvlLbl?: boolean;
    };
    y?: { /* ... */ };
  };
}

  // ...existing fields...
  noMultiLvlLbl?: boolean;
}

  // ...existing fields...
  axes?: {
    x?: {
      // ...existing fields...
      noMultiLvlLbl?: boolean | null;
    };
    y?: { /* ... */ };
  };
}
```

## Scope rules

- Only meaningful for bar / column / line / area charts (whose X axis is `<c:catAx>`).
- Silently dropped on scatter (both axes are value axes) and pie / doughnut (no axes at all).
- Default `false` (multi-level labels enabled) collapses to `undefined` on read so absence and the default round-trip identically.
- Non-boolean inputs collapse to `false` to keep the on-the-wire output stable, mirroring how `legendOverlay` / `roundedCorners` / axis `hidden` treat their inputs.

## Test plan

- [x] `pnpm test` (oxlint + oxfmt + tsgo + vitest) — 3253 tests passing.
- [x] `parseChart` covers truthy / falsy spellings, default collapse, missing `val`, value-axis rejection, scatter-chart rejection, and co-surfacing alongside title / gridlines / tick skips.
- [x] `writeChart` covers default emit, non-boolean fallback, exactly-one-emit guard, OOXML element ordering, scatter / pie / doughnut suppression, value-axis suppression, and end-to-end packaging via `writeXlsx`.
- [x] `cloneChart` covers inherit / null / false / true grammar, non-boolean fallback, chart-type coercion (bar -> column carries, bar -> scatter strips), composition with other axis overrides, and end-to-end through `parseChart -> cloneChart -> writeChart`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -722,6 +722,18 @@ absence, missing `val`, and unknown tokens all collapse to `undefined`
 so absence and the default round-trip identically. The reader accepts
 the OOXML truthy / falsy spellings (`"1"` / `"true"` / `"0"` /
 `"false"`).
+`ChartAxisInfo.noMultiLvlLbl` surfaces the per-axis
+`<c:noMultiLvlLbl val=".."/>` flag — Excel's "Format Axis ->
+Multi-level Category Labels" checkbox (the checkbox is the inverse:
+checked means tiered labels, i.e. `noMultiLvlLbl: false`). The OOXML
+schema places the element on `CT_CatAx` exclusively (even
+`<c:dateAx>`, `<c:valAx>`, and `<c:serAx>` reject it), so the parser
+ignores the element on every other axis flavour. Only an explicit
+`val="1"` (multi-tier labels collapsed onto a single line) surfaces
+`true`; the OOXML default `val="0"`, absence, missing `val`, and
+unknown tokens all collapse to `undefined` so absence and the default
+round-trip identically. The reader accepts the OOXML truthy / falsy
+spellings (`"1"` / `"true"` / `"0"` / `"false"`).
 `ChartAxisInfo.reverse` surfaces the per-axis
 `<c:scaling><c:orientation val="maxMin"/></c:scaling>` flag — Excel's
 "Categories / Values in reverse order" toggle. Only `"maxMin"` surfaces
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
## Review

I read the diff end-to-end (`_types.ts`, `chart-reader.ts`, `chart-writer.ts`, `chart-clone.ts`, plus the three test files) and cross-checked the implementation against the surrounding axis knobs (`lblAlgn`, `hidden`, `tickLblSkip`, etc.).

### Strengths

- **Reader/writer parity is exact.** `parseAxisNoMultiLvlLbl` mirrors `parseAxisHidden` byte-for-byte (truthy/falsy spelling acceptance, `val="0"` collapsing to `undefined`, missing `val` dropping). The writer's "always emit, only `true` flips to `val=1`" pattern matches `legendOverlay` / `roundedCorners` / axis `hidden`.
- **Scope rules are right.** The element is correctly gated to `<c:catAx>` only on every layer:
  - Reader: `axis.local === "catAx"` short-circuits before `parseAxisNoMultiLvlLbl` runs, so a stray flag on `<c:valAx>` / `<c:dateAx>` / `<c:serAx>` cannot leak through (test covers both single-valAx and scatter's two-valAx case).
  - Writer: only `buildBarAxes` consumes `xNoMultiLvlLbl`; scatter / pie / doughnut never see it.
  - Cloner: `isCatAxisX` (i.e. `type !== "scatter"`) gates the override, matching `lblOffset` / `lblAlgn`.
- **Override grammar is consistent.** `applyNoMultiLvlLblOverride` is structurally identical to `applyHiddenOverride` — same `undefined` (inherit) / `null` (drop) / `boolean` (replace) semantics with non-boolean inputs collapsing to `undefined`.
- **OOXML element ordering is verified.** The schema-order test (`lblOffset` -> `tickLblSkip` -> `tickMarkSkip` -> `noMultiLvlLbl`) pins what was already correct in the writer.
- **Round-trip determinism.** Default `false` collapses to `undefined` on the reader so absence and the OOXML default round-trip identically through `cloneChart`; only `true` survives.
- **Test coverage is thorough.** 31 new test cases cover truthy/falsy spellings, default collapse, missing `val`, value-axis rejection, scatter/pie/doughnut suppression, chart-type coercion (bar -> column carries, bar -> scatter strips), composition with other overrides, exactly-one-emit guards, and end-to-end packaging via `writeXlsx`.

### Doc parity

The PR didn't update README, so I added three paragraphs (`ChartAxisInfo.noMultiLvlLbl`, the writer-side `axes.x.noMultiLvlLbl` flag, and the `cloneChart` override) in commit 3942da9 to match the documentation parity already established for `lblAlgn` / `hidden`.

### Concerns

None. The implementation is consistent with the project's established patterns and the new behaviour is well-tested. Will merge once CI clears the docs commit.

PR #223, opened 2026-05-03
feat(xlsx): chart data-labels showLegendKey flag — read, write, clone-through

productdevbook:
## Summary

Surfaces the `<c:showLegendKey>` element on `<c:dLbls>` at the read, write, and clone layers so a template's `showLegendKey` toggle survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:showLegendKey>` mirrors Excel's "Format Data Labels -> Legend Key" checkbox — when on, Excel paints the legend's color swatch (the small marker / bar shown in the chart legend) inline with every data label. The element lives on `CT_DLbls`, which appears at both the chart-type level (`<c:barChart><c:dLbls>` etc.) and the per-series level (`<c:ser><c:dLbls>`), so the flag threads through every chart family that supports data labels (bar / column / line / area / pie / doughnut / scatter).

Up until now hucre's writer always pinned `val="0"`, so a template that pinned `val="1"` flattened back to the default on every parse -> clone -> write loop. This bridges another data-label-configuration gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.dataLabels?.showLegendKey); // true (template pinned val="1")

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      series: [{ values: "B2:B10", categories: "A2:A10" }],
      dataLabels: {
        showValue: true,
        showLegendKey: true, // paint the legend swatch beside every label
        position: "outEnd",
      },
    }],
  }],
});

// -- Clone-through --
  // dataLabels: undefined  -> inherit (showLegendKey: true carries through)
  // dataLabels: null       -> drop the inherited block entirely
  // dataLabels: { ... }    -> wholesale replace (no per-field merge)
});
```

## Model

```ts
export interface ChartDataLabels {
  // ...existing fields...
  showLegendKey?: boolean;
}

export interface ChartDataLabelsInfo {
  // ...existing fields...
  showLegendKey?: boolean;
}
```

## Scope rules

- Element appears at both the chart-type-level and per-series `<c:dLbls>`. The flag follows the existing `dataLabels` resolution path on both layers (chart-level + series overrides via `seriesOverrides[i].dataLabels`).
- Default `false` (no legend swatch) collapses to `undefined` on read so absence and the OOXML default round-trip identically.
- Non-boolean inputs collapse to `false` on the writer side to keep the on-the-wire output stable, mirroring how the other `show*` toggles (showValue / showCategoryName / showSeriesName / showPercent) treat their inputs.
- A `<c:dLbls>` block whose only meaningful flag is `showLegendKey: true` (every other show toggle off, no position, no separator) still surfaces on the read side — Excel renders the swatch beside otherwise-empty label slots.
- Element ordering inside `<c:dLbls>` follows CT_DLbls: `dLblPos` (when set) before `showLegendKey` before `showVal` / `showCatName` / `showSerName` / `showPercent` / `showBubbleSize` before `separator`.

## Test plan

- [x] `pnpm test` (oxlint + oxfmt + tsgo + vitest) — 3281 tests passing.
- [x] `parseChart` covers truthy / falsy spellings, default collapse, missing `val`, unknown tokens, the showLegendKey-alone-surfaces case, series-level surfacing, and co-surfacing with other show toggles + separator.
- [x] `writeChart` covers default `val="0"` emit, true emit, false-equals-omit, non-boolean fallback, exactly-one-emit guard, OOXML element ordering, series-level threading, multi-family threading (pie / line / scatter), and end-to-end packaging via `writeXlsx`.
- [x] `cloneChart` covers chart-level inherit / null / wholesale-replace, per-series inherit / null / wholesale-replace, composition with other show toggles + position + separator, chart-type coercion (bar -> line) carry, and end-to-end through `parseChart -> cloneChart -> writeChart` plus packaging via `writeXlsx`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -562,10 +562,11 @@ for (const sheet of wb.sheets) {
     // }
 
     // chart.dataLabels surfaces the chart-type-level <c:dLbls> block.
-    // showValue / showCategoryName / showSeriesName / showPercent and
-    // position / separator round-trip through cloneChart unchanged.
+    // showValue / showCategoryName / showSeriesName / showPercent /
+    // showLegendKey and position / separator round-trip through
+    // cloneChart unchanged.
     console.log(chart.dataLabels);
-    // e.g. { showValue: true, position: "outEnd" }
+    // e.g. { showValue: true, showLegendKey: true, position: "outEnd" }
 
     for (const s of chart.series ?? []) {
       console.log(s.kind, s.index, s.name, s.valuesRef, s.categoriesRef, s.color);
@@ -630,10 +631,15 @@ never report one.
 `Chart.dataLabels` mirrors the writer-side `SheetChart.dataLabels`
 and surfaces the toggles Excel carries inside `<c:dLbls>`
 (`showValue`, `showCategoryName`, `showSeriesName`, `showPercent`,
-plus `position` and `separator`). Series-level overrides land on
-`ChartSeriesInfo.dataLabels`; a `<c:dLbls>` block that only contains
-`<c:delete val="1"/>` (Excel's "labels off" idiom) collapses to
-`undefined` rather than a record so callers see the absence cleanly.
+`showLegendKey`, plus `position` and `separator`). Series-level
+overrides land on `ChartSeriesInfo.dataLabels`; a `<c:dLbls>` block
+that only contains `<c:delete val="1"/>` (Excel's "labels off" idiom)
+collapses to `undefined` rather than a record so callers see the
+absence cleanly. `showLegendKey` mirrors Excel's "Format Data Labels
+-> Legend Key" checkbox: pin it `true` to paint the legend's color
+swatch beside every label. The OOXML default `false` collapses to
+`undefined` on the read side so absence and `<c:showLegendKey val="0"/>`
+round-trip identically through `cloneChart`.
 `Chart.holeSize` surfaces `<c:doughnutChart><c:holeSize val=".."/>`
 on doughnut charts so a parsed template can round-trip its hole back
 through `cloneChart`; non-doughnut charts (and doughnut charts that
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
## Review

Reviewed the read / write / clone wiring for `<c:showLegendKey>` plus the test plan.

**Correctness**
- `parseDataLabels` correctly handles the OOXML default (`val="0"` and absence both collapse to `undefined`), only surfaces `true` for `val="1"` / `val="true"`, and exempts `showLegendKey: true` alone from the "block is empty, drop it" collapse so the read-side behavior matches the documented contract.
- `buildDataLabelsBody` keeps the deterministic always-emit pattern, gates `val="1"` on the strict `=== true` check (matching how `legendOverlay`, `roundedCorners`, and axis `hidden` treat their inputs), and preserves the OOXML CT_DLbls element ordering (showLegendKey -> showVal -> showCatName -> showSerName -> showPercent -> showBubbleSize).
- `cloneChart` already shallow-spreads `dataLabels` from `ChartDataLabelsInfo`, so `showLegendKey` threads through automatically without changes — confirmed by the chart-level / per-series / chart-type-coercion tests.

**Edge case + test coverage**
- Read side covers truthy / falsy spellings, default collapse, missing `val`, unknown tokens, the showLegendKey-alone-surfaces case, and series-level surfacing.
- Write side covers default emit, true emit, false-equals-omit, non-boolean fallback, single-emit guard, ordering vs `dLblPos` and other show toggles, series-level threading, multi-family threading (pie / line / scatter), and end-to-end packaging via `writeXlsx`.
- Clone side covers chart-level inherit / null / wholesale-replace, per-series inherit / null / wholesale-replace, composition with other show toggles + position + separator, chart-type coercion (bar -> line) carry, and end-to-end through `parseChart -> cloneChart -> writeChart` plus packaging via `writeXlsx`.

**Type safety / consistency**
- `ChartDataLabels.showLegendKey` and `ChartDataLabelsInfo.showLegendKey` mirror the existing `show*` shape exactly; doc comments are clear about Excel semantics and OOXML defaults.
- README updates land in the same three sections (Reader Notes, model index, dataLabels options) used by the other show toggles, including the legend-key-specific paragraph that explains the OOXML default round-trip.

**Fixes applied**
- Merged `origin/main` to resolve the conflicts caused by #222 (`noMultiLvlLbl`) landing first. Kept both feature describe blocks (`showLegendKey` + `noMultiLvlLbl`) as siblings in `test/chart-clone.test.ts`, `test/charts-write.test.ts`, and `test/charts.test.ts`.
- Re-ran `pnpm test` (oxlint + oxfmt + tsgo + vitest) — all 3317 tests pass — and `pnpm build` — obuild succeeds.

No remaining concerns. Pending CI green, ready to merge.

PR #224, opened 2026-05-03
feat(xlsx): chart title overlay flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:title><c:overlay val=".."/></c:title>` at the read, write, and clone layers so a template's chart-title overlay survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:title><c:overlay>` mirrors Excel's "Format Chart Title -> Show the title without overlapping the chart" toggle (the checkbox is the inverse of the flag — checked means `false`, unchecked means `true`). `val=1` draws the title on top of the plot area so the chart series get the full frame; `val=0` reserves a slot for the title above the plot area and shrinks the plot area to make room.

Up until now hucre's writer always pinned `val=0`, so a template that overlapped its title flattened back to a vertically-stacked title on every parse -> clone -> write loop. Sibling of the existing `legendOverlay` flag — both elements share a local name but live on different parents (`<c:title>` vs `<c:legend>`), so the two compose freely. This bridges another chart-level gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.titleOverlay); // true when the template pinned val="1",
                                  // undefined when the template used the default 0

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      title: "Revenue",
      titleOverlay: true,  // title draws on top of the plot area
    }],
  }],
});

// -- Clone-through --
  titleOverlay: true,    // replace
  // titleOverlay: null  // drop the inherited overlay (writer falls back to 0)
  // titleOverlay: undefined // inherit the source's parsed value
});
```

## Model

```ts
interface SheetChart {
  // ...existing fields...
  titleOverlay?: boolean;
}

interface Chart {
  // ...existing fields...
  titleOverlay?: boolean;
}

interface CloneChartOptions {
  // ...existing fields...
  titleOverlay?: boolean | null;
}
```

The read-side `Chart.titleOverlay` mirrors the write-side `SheetChart.titleOverlay` so a parsed value slots straight back into `cloneChart` without transformation. Same shape as the existing `legendOverlay` / `roundedCorners` / `plotVisOnly` toggles.

## Behavior

- **Read** — `parseChart` pulls the value off `<c:title><c:overlay val=".."/></c:title>`. The OOXML default `false` (`val="0"`) collapses to `undefined` so absence and the default round-trip identically; only an explicit `val="1"` surfaces `true`. Accepts the OOXML truthy / falsy spellings (`"1"` / `"true"` / `"0"` / `"false"`); unknown values and missing `val` attributes drop to `undefined`. The flag is dropped whenever the chart omits the `<c:title>` element entirely. The lookup is scoped to direct `<c:title>` children, so a sibling's `<c:overlay>` (e.g. the legend's) cannot leak into the title flag.
- **Write** — `<c:overlay>` is always emitted inside `<c:title>` because Excel's reference serialization includes it on every visible title. The writer pins the caller's override when set; absence (and the default `false`) emit `val="0"` so untouched charts match Excel's reference output byte-for-byte. Non-boolean inputs collapse to `false`. The element sits after `<c:tx>` per the strict CT_Title schema sequence. When the chart renders no title (no `title` set or `showTitle: false`) the writer skips the entire `<c:title>` element so no orphan overlay can leak through.
- **Clone** — Each override accepts the standard `undefined` (inherit) / `null` (drop, writer falls back to `0`) / `boolean` (replace) grammar that mirrors `legendOverlay` / `roundedCorners`. The inherited value is silently dropped from the cloned `SheetChart` whenever the resolved chart renders no title (resolved `title` is `undefined` or `showTitle === false`) — a missing title has no overlay slot. Re-introducing a missing source title through `title: "..."` on the override re-opens the slot, and an explicit `titleOverlay: true` override threads through.

## Edge cases

- A chart with no `<c:title>` element parses to `titleOverlay: undefined`.
- An explicit `titleOverlay: true` override with `title: null` (or `showTitle: false`) collapses to `undefined` on the cloned `SheetChart`.
- Non-boolean `titleOverlay` inputs collapse to the OOXML `false` default rather than fabricate a flag.
- The element is emitted exactly once per title (the legend also carries its own `<c:overlay>` element, scoped to the legend block — they do not collide). The reader scopes its lookup to direct `<c:title>` children so a sibling's `<c:overlay>` cannot leak in.
- Pie / doughnut / line / bar / column / area / scatter all support the flag identically because `<c:title>` is a chart-level sibling of `<c:plotArea>`.
- Composes independently with `legendOverlay`: pinning one does not change the other on either the writer or the clone path.

## Test plan

- [x] `pnpm test` (oxlint + oxfmt + tsgo + vitest) passes — 3325 vitest tests.
- [x] `pnpm build` succeeds.
- [x] 13 new `parseChart` tests cover the truthy / falsy spellings, default collapse, missing val, missing element, missing title, multi-family threading (line / bar / pie / doughnut), independence from legend overlay, sibling-element scoping, and co-surfacing alongside other chart-level toggles.
- [x] 14 new `writeChart` tests cover the default emit, true / false threading, OOXML element ordering, exactly-one-emit guard, missing-title and `showTitle: false` suppression, multi-family threading, independent composition with legend overlay, non-boolean fallback, parse-side round-trip (true / default / explicit-false), and end-to-end packaging via `writeXlsx`.
- [x] 15 new `cloneChart` tests cover inherit / null / replace grammar, chart-type coercion (line -> column / doughnut), missing-title / `showTitle: false` drop guards, override-wins-when-re-introducing-title, independent composition with `legendOverlay` clone-through, and end-to-end through `parseChart -> cloneChart -> writeChart` plus packaging via `writeXlsx`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -765,6 +765,17 @@ missing `val` attributes drop to `undefined`. The flag is dropped
 whenever `Chart.legend` is `false` or the chart omits the legend
 element entirely — there is no overlay slot on a hidden legend, so the
 parsed shape stays minimal.
+`Chart.titleOverlay` surfaces the title-overlay flag pulled from
+`<c:title><c:overlay val=".."/></c:title>` — Excel's "Format Chart
+Title → Show the title without overlapping the chart" toggle (the
+checkbox is the inverse of this flag — checked means `false`, unchecked
+means `true`). The OOXML default `false` collapses to `undefined` so
+absence and `<c:overlay val="0"/>` round-trip identically; only an
+explicit `val="1"` surfaces `true`. The reader accepts the OOXML
+truthy / falsy spellings (`"1"` / `"true"` / `"0"` / `"false"`);
+unknown values and missing `val` attributes drop to `undefined`. The
+flag is dropped whenever the chart omits the `<c:title>` element
+entirely — there is no overlay slot to surface in that case.
 `ChartSeriesInfo.smooth` surfaces the per-series
 `<c:ser><c:smooth val=".."/>` flag — Excel's "Format Data Series →
 Line → Smoothed line" toggle — only on `line` / `line3D` / `scatter`
@@ -960,6 +971,19 @@ on top of the plot area so the chart series get the full frame
 (`val="1"`). The flag is silently ignored when `legend: false`
 suppresses the entire legend element — there is no overlay slot on a
 hidden legend, so the writer skips emitting any orphaned `<c:overlay>`.
+The chart-level `titleOverlay` field maps to `<c:overlay val=".."/>`
+inside `<c:title>` — Excel's "Format Chart Title → Show the title
+without overlapping the chart" toggle (the checkbox is the inverse of
+this flag — checked means `false`, unchecked means `true`). Absent it,
+the writer emits the OOXML default `val="0"` (the title reserves its
+own slot above the plot area and the plot area shrinks to make room),
+matching Excel's reference serialization. Pin `titleOverlay: true` to
+draw the title on top of the plot area so the chart series get the
+full frame (`val="1"`). The flag is silently ignored when no title is
+emitted (no `title` set or `showTitle: false`) — there is no `<c:title>`
+block to host the overlay child in either case. Independent of
+`legendOverlay`: the legend and title `<c:overlay>` elements live on
+different parents, so the two flags compose freely.
 The `axes.x.tickLblSkip` and `axes.x.tickMarkSkip` fields thin out a
 crowded category axis (`<c:catAx><c:tickLblSkip val=".."/>` and
 `<c:catAx><c:tickMarkSkip val=".."/>`). Pass a positive integer to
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
Reviewed the chart title-overlay implementation across read / write / clone layers and verified locally:

- pnpm test: 3325 tests pass (lint + format + typecheck + vitest)
- pnpm build: succeeds

Notes from the review:

- Reader scopes the lookup to direct `<c:title>` children via `findChild(title, "overlay")`, so a sibling `<c:legend><c:overlay>` cannot leak into `titleOverlay`. The dedicated test (`does not pull <c:overlay> from a sibling element by mistake`) confirms this.
- The default-collapse symmetry (OOXML `val="0"` / `"false"` -> `undefined`) matches the established `legendOverlay` / `roundedCorners` / `plotVisOnly` convention so `parseChart -> cloneChart -> writeChart` round-trips identically for both absence and the explicit default.
- Writer always emits `<c:overlay>` inside `<c:title>` matching Excel's reference serialization, in the correct CT_Title position (after `<c:tx>`). The non-boolean fallback (`titleOverlay === true` only) prevents stray truthy strings from producing `val="1"`.
- The clone-side `titleRendered` guard `(out.showTitle ?? Boolean(out.title)) && out.title !== undefined` correctly mirrors the writer's `showTitle && chart.title` gate, so the inherited overlay flag never survives to a clone with no title block to host it. Both the `title: null` and `showTitle: false` paths drop the flag, and the override grammar (`undefined` inherit / `null` drop / `boolean` replace) matches `legendOverlay` for free composition.
- Test coverage is thorough across all three layers (13 reader + 14 writer + 15 clone), including round-trips through `writeXlsx`, multi-family threading, independence from `legendOverlay`, and the missing-`val` / unknown-`val` edge cases.
- README is updated in all three relevant sections (read / write / clone).

CI is already green on the head commit. Merging.

PR #226, opened 2026-05-03
feat(xlsx): chart drop / hi-low lines — read, write, clone-through

productdevbook:
## Summary

Surfaces the chart-type-level `<c:dropLines/>` and `<c:hiLowLines/>` elements at the read, write, and clone layers so a template's connector-line configuration survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:dropLines>` mirrors Excel's "Add Chart Element -> Lines -> Drop Lines" toggle (vertical reference lines connecting each data point to the category axis). `<c:hiLowLines>` mirrors Excel's "Add Chart Element -> Lines -> High-Low Lines" toggle (vertical connectors between the highest and lowest series values at each category position; the same connector painted on stock charts). Both elements are bare — their mere presence paints the connector lines, so the model is a simple boolean toggle.

Up until now hucre's writer had no slot for either element, so a template that pinned drop / hi-low lines flattened back to no connectors on every parse -> clone -> write loop. This bridges another visualization gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts
import { readXlsx, writeXlsx, cloneChart } from "hucre";

// -- Read side --
const wb = await readXlsx(templateBytes);
const source = wb.sheets[0].charts![0];
console.log(source.dropLines);  // true when the template pinned <c:dropLines/>
console.log(source.hiLowLines); // true when the template pinned <c:hiLowLines/>

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "line",
      series: [
        { name: "High", values: "B2:B10", categories: "A2:A10" },
        { name: "Low", values: "C2:C10", categories: "A2:A10" },
      ],
      dropLines: true,   // vertical lines from each point down to the X axis
      hiLowLines: true,  // vertical connector between high / low at each X
      anchor: { from: { row: 6, col: 0 } },
    }],
  }],
});

// -- Clone-through --
const clone = cloneChart(source, {
  anchor: { from: { row: 14, col: 0 } },
  // dropLines: undefined -> inherit
  // dropLines: null      -> drop inherited flag
  // dropLines: true      -> pin
  // dropLines: false     -> equivalent to null (writer treats both as default)
});
```

## Model

```ts
// Read side
export interface Chart {
  // ...existing fields...
  dropLines?: boolean;   // <c:lineChart> / <c:line3DChart> / <c:areaChart> / <c:area3DChart>
[...]

PR #227, opened 2026-05-03
feat(xlsx): chart line up/down bars — read, write, clone-through

productdevbook:
## Summary

Surfaces the chart-type-level `<c:upDownBars>` element at the read, write, and clone layers so a template's up / down bars configuration survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh line chart.

`<c:upDownBars>` mirrors Excel's "Add Chart Element -> Up/Down Bars" toggle on a line chart — vertical bars at each category that connect a paired series, where the bar's top tracks the higher value and bottom tracks the lower one. The element is the visualization Excel paints natively on line-style stock charts (open / close differences on a price series). The model at this layer is a plain presence flag (`upDownBars: boolean`); the writer emits the element with Excel's reference `<c:gapWidth val="150"/>` child whenever the flag is `true` on a line chart.

Up until now hucre's writer had no slot for the element, so a template that pinned up / down bars flattened back to no bars on every parse -> clone -> write loop. This bridges another visualization gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.upDownBars); // true when the template pinned <c:upDownBars/>

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Stock",
    rows: [...],
    charts: [{
      type: "line",
      series: [
        { name: "High", values: "B2:B10", categories: "A2:A10" },
        { name: "Low",  values: "C2:C10", categories: "A2:A10" },
      ],
      upDownBars: true, // vertical bars connecting High / Low at each category
    }],
  }],
});

// -- Clone-through --
  // upDownBars: undefined -> inherit
  // upDownBars: null      -> drop inherited flag (no element emitted)
  // upDownBars: true      -> pin
  // upDownBars: false     -> equivalent to absence (writer collapses to default)
});
```

## Model

```ts
// Read side
export interface Chart {
  // ...existing fields...
  upDownBars?: boolean;
}

// Write side
export interface SheetChart {
  // ...existing fields...
  upDownBars?: boolean;
}

// Clone side
  // ...existing fields...
  upDownBars?: boolean | null;
}
```

## Scope rules

- The OOXML schema places `<c:upDownBars>` on `CT_LineChart` / `CT_Line3DChart` / `CT_StockChart` exclusively. The writer only authors `<c:lineChart>`, so the flag emits only when `type: "line"`; bar / column / pie / doughnut / area / scatter silently drop it.
- The reader surfaces the flag from any line-flavored chart-type body (line / line3D / stock); a stray `<c:upDownBars>` on bar / column / pie / doughnut / area / scatter chart-type elements is ignored.
- Element ordering inside `<c:lineChart>` follows CT_LineChart: `<c:upDownBars>` lands before `<c:marker>` and `<c:axId>`.
- The writer emits a default `<c:gapWidth val="150"/>` child to mirror Excel's reference serialization. `<c:upBars>` / `<c:downBars>` are intentionally omitted — Excel paints the default white-up / black-down bars on a fresh toggle. A richer model (custom gap widths, per-bar styling) can layer on later if needed.
- The clone layer drops the flag on a flatten to anything but `line` (line -> column flattens drop the element) so a stale up/down-bars hint never leaks into a chart-type element whose schema rejects it.

## Test plan

- [x] `pnpm exec vitest run --dir test` — 3443 tests passing.
- [x] `pnpm typecheck` — tsgo clean.
- [x] `pnpm exec oxlint src test` — 0 errors.
- [x] `pnpm exec oxfmt --check src test` — clean.
- [x] `parseChart` covers bare element, element with `<c:gapWidth>` child, element with `<c:upBars>` / `<c:downBars>` children, absence collapse, scope drops on bar / area / scatter, surfacing on stock + line3D bodies, and co-existence with chart-level toggles.
- [x] `writeChart` covers absence default, `true` emit with default gapWidth=150, `false` collapse to absence, OOXML element ordering (before `<c:marker>` / `<c:axId>`), single-emit guard, drop on every non-line chart family, parseChart round-trip, and end-to-end packaging via `writeXlsx`.
- [x] `cloneChart` covers chart-level inherit / null / wholesale-replace, addition on a source that lacked the flag, drops on flatten to column / area / pie / scatter, carry through stock -> line flatten, and end-to-end through `parseChart -> cloneChart -> writeChart` plus packaging via `writeXlsx`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -827,6 +827,19 @@ series (the OOXML schema places `<c:invertIfNegative>` exclusively on
 `CT_BarSer` / `CT_Bar3DSer`). Absence and the OOXML default
 `val="0"` both collapse to `undefined`, so only an explicit
 `<c:invertIfNegative val="1"/>` round-trips as `invertIfNegative: true`.
+`Chart.upDownBars` surfaces the line-chart up / down bars flag pulled
+from `<c:lineChart><c:upDownBars>...</c:upDownBars></c:lineChart>` —
+Excel's "Add Chart Element -> Up/Down Bars" toggle. The element
+paints a vertical bar at each category whose top tracks the higher
+series value and bottom tracks the lower one (typically used to
+visualize open / close differences on a line-style stock chart).
+Surfaces `true` whenever the element is present (with or without the
+optional `<c:gapWidth>` / `<c:upBars>` / `<c:downBars>` children — the
+model is a plain presence flag at this layer); absence collapses to
+`undefined`. Only line-flavored chart-type bodies surface the field —
+`CT_LineChart` / `CT_Line3DChart` / `CT_StockChart` per the OOXML
+schema; a stray `<c:upDownBars>` on a bar / column / pie / doughnut /
+area / scatter chart is ignored.
 Sheets that hucre actively regenerates because they
 also carry hucre-managed images currently keep the chart bodies but
 lose the in-drawing chart anchor — merging hucre's drawing output
@@ -1006,6 +1019,18 @@ emitted (no `title` set or `showTitle: false`) — there is no `<c:title>`
 block to host the overlay child in either case. Independent of
 `legendOverlay`: the legend and title `<c:overlay>` elements live on
 different parents, so the two flags compose freely.
+The chart-level `upDownBars` field maps to `<c:upDownBars>` inside
+`<c:lineChart>` — Excel's "Add Chart Element -> Up/Down Bars" toggle
+on a line chart. The element paints a vertical bar at each category
+whose top tracks the higher series value and bottom tracks the lower
+one (typically used to visualize open / close differences on a
+line-style stock chart). The OOXML default is absence, so the writer
+omits the element on a fresh chart. Pin `upDownBars: true` on a
+`type: "line"` chart to emit the element with Excel's reference
+`<c:gapWidth val="150"/>` child. The flag is silently ignored on
+every other chart family — `<c:upDownBars>` lives exclusively on
+`CT_LineChart` / `CT_Line3DChart` / `CT_StockChart` per the OOXML
+schema, and the writer never authors the 3D / stock variants.
 The `axes.x.tickLblSkip` and `axes.x.tickMarkSkip` fields thin out a
 crowded category axis (`<c:catAx><c:tickLblSkip val=".."/>` and
 `<c:catAx><c:tickMarkSkip val=".."/>`). Pass a positive integer to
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
Reviewed the chart up / down bars bridge end to end and resolved the merge conflict against `main` (PR #226 — drop / hi-low lines — landed in the meantime).

## Review summary

- Read-side gating in `parseChart` correctly restricts `<c:upDownBars>` to `lineChart` / `line3DChart` / `stockChart` per CT_LineChart / CT_Line3DChart / CT_StockChart. The bare-presence model surfaces `true` on the parent element regardless of whether the optional `<c:gapWidth>` / `<c:upBars>` / `<c:downBars>` children are present, matching how Excel actually paints the toggle. Stray elements on bar / column / pie / doughnut / area / scatter chart-type bodies drop silently — confirmed by the per-family negative tests.
- Write-side ordering in `buildLineChart` correctly slots `<c:upDownBars>` after the new `<c:dropLines>` / `<c:hiLowLines>` slots and before `<c:marker>`, matching the CT_LineChart sequence (grouping, varyColors?, ser*, dLbls?, dropLines?, hiLowLines?, upDownBars?, marker?, axId+). Only literal `true` emits the element, mirroring the precedent set by the bare-element writers nearby. The default `<c:gapWidth val="150"/>` child mirrors Excel's reference serialization for a freshly toggled "Add Chart Element -> Up/Down Bars".
- Clone-through resolver `resolveUpDownBars` mirrors `resolveDropLines` / `resolveHiLowLines` (inherit on `undefined`, drop on `null`, replace on `boolean`). The line-only scope gate after type coercion ensures a stale up/down-bars hint never leaks into a column / pie / doughnut / area / scatter clone — and the stock -> line carry test confirms a stock template's flag survives the typical flatten.
- Test coverage spans bare element, gapWidth-only, upBars / downBars children, absence collapse, scope drops on bar / area / scatter, surfacing on stock + line3D bodies, single-emit guard, OOXML element ordering (before `<c:marker>` / `<c:axId>`), end-to-end round trip via `parseChart -> cloneChart -> writeChart -> parseChart`, and packaging via `writeXlsx` with `ZipReader` extraction. README updates land in the read / write / clone sections in the same shape as the recently merged drop / hi-low lines docs.

## Fixes

- Resolved a textual merge conflict in `README.md`, `src/xlsx/chart-reader.ts`, `src/xlsx/chart-writer.ts`, `test/charts.test.ts`, `test/charts-write.test.ts`, and `test/chart-clone.test.ts`. Both this branch and `main` (drop / hi-low lines PR #226) appended their describe blocks at the end of each test file and added parallel chart-level toggle plumbing in the source / docs. Kept main's content first, then the upDownBars content immediately after — preserving both sets of tests with no behavior change. The line-chart writer now emits `<c:dropLines>` / `<c:hiLowLines>` / `<c:upDownBars>` in CT_LineChart schema sequence (dropLines before hiLowLines before upDownBars before marker), so a chart that pins all three flags still serializes in the schema-defined order.
- Reran `pnpm test` (oxlint + oxfmt + tsgo + vitest) and `pnpm build` locally — 3502 tests pass and obuild succeeds.

## Notes

- `resolveUpDownBars` is functionally identical to `resolveDropLines` / `resolveHiLowLines`. Folding the three into one shared boolean-flag resolver would shave a few lines, but keeping per-flag resolvers matches the pattern used for the other chart-level toggles and makes the JSDoc self-documenting. Not blocking — could be a follow-up cleanup once the chart-level toggle catalog stabilizes.

CI is pending on the merge commit. Will merge once it goes green.

PR #228, opened 2026-05-03
feat(xlsx): chart built-in style preset — read, write, clone-through

productdevbook:
## Summary

Surfaces the chart-space-level `<c:style val=".."/>` element at the read, write, and clone layers so a template's built-in chart-style preset survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch.

`<c:style>` mirrors Excel's "Chart Design -> Chart Styles" gallery — each integer in 1–48 picks one of the numbered presets that cycle a colored background, gridline density, border, and label styling across the chart. The element sits on `<c:chartSpace>` (a sibling of `<c:chart>`, not a child) per `CT_ChartSpace`, after `<c:roundedCorners>` and before `<c:chart>`.

Up until now hucre's writer skipped the element entirely and the reader ignored it, so a templated chart flattened back to the unstyled default on every parse -> clone -> write loop. This bridges another chart-frame gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// ── Read side ──
console.log(source.style); // 27 when the template pinned <c:style val="27"/>,
                           // undefined when the template omitted the element

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      style: 27, // pin the gallery preset
    }],
  }],
});

// ── Clone-through ──
  style: 4,        // replace
  // style: null   // drop the inherited preset (writer skips <c:style>)
  // style: undefined // inherit the source's parsed value
});
```

## Model

```ts
interface Chart {
  /* ...existing fields... */
  style?: number;
}

interface SheetChart {
  /* ...existing fields... */
  style?: number;
}

interface CloneChartOptions {
  /* ...existing fields... */
  style?: number | null;
}
```

The read-side `Chart.style` mirrors the write-side `SheetChart.style` so a parsed value slots straight back into `cloneChart` without transformation. The override grammar mirrors `roundedCorners` / `plotVisOnly` so the chart-frame toggles compose the same way at the call site.

## Scope rules

- The OOXML schema declares `<c:style>` `val` as `xsd:unsignedByte` in the gallery range 1–48. Out-of-range / non-integer / missing values drop to absence on both read and write — the reader collapses them to `undefined` and the writer skips emission rather than emit a token Excel would reject.
- Element ordering inside `<c:chartSpace>` follows the `CT_ChartSpace` sequence: `<c:style>` lands after `<c:roundedCorners>` and before `<c:chart>`. The chart-frame helper applies on every chart family — the writer threads `style` through bar / column / line / pie / doughnut / area / scatter identically, since the preset has no chart-family restriction.
- Unlike `roundedCorners` (always emitted with the explicit default `0`), the writer skips the element entirely when `style` is unset so a fresh chart matches Excel's implicit default look. Excel's reference serialization for a fresh chart pins style 2, but a chart that omits the element renders identically — collapsing absence to absence keeps the parsed shape minimal so a roundtrip of an unstyled chart returns no `style` field.
- The clone layer carries `style` through every chart-type flatten (line -> column / doughnut / pie / area / scatter) because `<c:style>` lives on `<c:chartSpace>` and is valid on every chart family.

## Test plan

- [x] `pnpm test` — 3532 tests passing (47 new), 0 lint errors, 0 type errors, oxfmt clean.
- [x] `parseChart` covers default-shape (val="2"), templated mid-range (val="27"), range bounds (1 / 48), absence, out-of-range (0 / 49 / 255), non-integer (3.5 / -1 / "two" / "3px"), missing val attribute, and co-existence with `roundedCorners` / `plotVisOnly` / `dispBlanksAs` / `varyColors`.
- [x] `writeChart` covers absence skip, `style: 27` emit, range bounds emit, OOXML element ordering (after `<c:roundedCorners>`, before `<c:chart>`), single-emit guard, out-of-range drop (0 / 49 / 100 / 255 / -3), non-integer drop (3.5 / NaN / Infinity), every chart family (bar / column / line / pie / doughnut / area / scatter), `parseChart` round-trip, and end-to-end packaging via `writeXlsx`.
- [x] `cloneChart` covers chart-level inherit / null / wholesale-replace, addition on a source that lacked the field, carry through every flatten (line -> column / doughnut), and end-to-end through `parseChart -> cloneChart -> writeChart` plus packaging via `writeXlsx`.

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1143,6 +1143,26 @@ export interface SheetChart {
    */
   upDownBars?: boolean;
   /**
+   * Built-in chart style preset. Maps to `<c:style val=".."/>` on
+   * `<c:chartSpace>` (a sibling of `<c:chart>`, not a child). Mirrors
+   * Excel's "Chart Design -> Chart Styles" gallery — each integer
+   * picks one of the 48 numbered presets that cycle a colored
+   * background, gridline density, border, and label styling across
+   * the chart.
+   *
+   * Default: omitted — when the field is absent the writer skips the
+   * element and Excel renders the chart with its application default
+   * look. Set an integer in the OOXML range (1–48) to pin a preset;
+   * out-of-range and non-integer values are silently dropped rather
+   * than emit a token Excel would reject.
+   *
+   * Useful when matching a dashboard whose other charts already
+   * carry a particular preset look from a template — clone-through
+   * preserves the parsed value so a fresh chart and a templated chart
+   * compose side by side without manual re-styling.
+   */
+  style?: number;
+  /**
    * Per-axis configuration rendered alongside the plot area. The `x`
    * axis is the category axis for bar/column/line/area (or the bottom
    * value axis for scatter); the `y` axis is the value axis. Ignored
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed the diff and changed files end-to-end.

**Correctness** — `<c:style>` is correctly placed on `<c:chartSpace>` (sibling of `<c:chart>`), positioned after `<c:roundedCorners>` and before `<c:chart>` per CT_ChartSpace. Reader/writer/clone shape checks (integer + 1–48 range) match the `xsd:unsignedByte` schema and stay consistent across the three layers.

**Edge cases** — Reader uses a strict `/^\d+$/` regex so `"3px"`/`"3.5"`/`"-1"` are rejected before `Number()`. Writer guards `Number.isInteger` against NaN/Infinity/floats. Clone follows the established `undefined → inherit / null → drop / number → replace` grammar matching `roundedCorners`/`plotVisOnly`. Since `<c:style>` lives on chart-space and has no chart-family restriction, setting `out.style` once (rather than threading through each type-flatten branch) is the right call.

**Test coverage** — 47 new tests across the three test files cover defaults, mid-range, range bounds, absence, out-of-range (0/49/100/255/-3), non-integer (3.5/NaN/±Infinity/"two"/"3px"), missing val, ordering, single-emit guard, every chart family (bar/column/line/pie/doughnut/area/scatter), parse round-trip, packaging via `writeXlsx`, clone-through with line→column and line→doughnut flattens, and co-existence with `roundedCorners`/`plotVisOnly`/`dispBlanksAs`/`varyColors`.

**Conventions** — Doc comments mirror the patterns in `resolveRoundedCorners` and `parseRoundedCorners`. The decision to skip the element on absence (rather than emit Excel's reference `val="2"`) keeps round-trip minimal which matches the rest of the chart-frame handling.

CI is green on the head commit. No remaining concerns — merging.

PR #229, opened 2026-05-03
feat(xlsx): chart value-axis display units — read, write, clone-through

productdevbook:
## Summary

Surfaces the per-value-axis `<c:dispUnits>` element at the read, write, and clone layers so a template's display-unit preset survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:dispUnits>` mirrors Excel's "Format Axis -> Display units" dropdown — every numeric tick label is divided by the chosen preset before rendering, so a chart whose source range stores raw amounts (e.g. `1_500_000`) can show compact tick labels (`1.5` with an optional "Millions" annotation) without modifying the underlying cells. The element sits as the very last child of `<c:valAx>` (after `<c:minorUnit>`) per `CT_ValAx` in ECMA-376 §21.2.2.32, and holds an XSD choice between `<c:builtInUnit val=".."/>` and `<c:custUnit val=".."/>`, optionally followed by `<c:dispUnitsLbl>`.

The model at this layer surfaces only the built-in path (`unit: ChartAxisDispUnit`) plus a presence flag for the bare `<c:dispUnitsLbl>` annotation (`showLabel: boolean`). The custom numeric divisor and rich-text label customization are intentionally out of scope.

Up until now hucre's writer had no slot for the element, so a template that pinned a display-unit preset flattened back to raw numbers on every parse -> clone -> write loop. This bridges another axis-rendering gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.axes?.y?.dispUnits);
// { unit: "millions", showLabel: true } when the template pinned both;
// undefined when the template omitted <c:dispUnits> entirely.

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      axes: {
        y: { dispUnits: { unit: "millions", showLabel: true } },
      },
    }],
  }],
});

// -- Clone-through --
  axes: {
    y: {
      // dispUnits: undefined           -> inherit
      // dispUnits: null                -> drop (no element emitted)
      // dispUnits: "thousands"         -> shorthand replace
      // dispUnits: { unit, showLabel } -> full replace
      dispUnits: "thousands",
    },
  },
});
```

## Model

```ts
export type ChartAxisDispUnit =
  | "hundreds" | "thousands" | "tenThousands" | "hundredThousands"
  | "millions" | "tenMillions" | "hundredMillions"
  | "billions" | "trillions";

export interface ChartAxisDispUnits {
  unit: ChartAxisDispUnit;
  showLabel?: boolean;
}

// Read side
  // ...existing fields...
  dispUnits?: ChartAxisDispUnits;
}

// Write side — accepts the shorthand string for ergonomics
export interface SheetChart {
  axes?: {
    x?: { /* ...existing... */ dispUnits?: ChartAxisDispUnits | ChartAxisDispUnit };
    y?: { /* ...existing... */ dispUnits?: ChartAxisDispUnits | ChartAxisDispUnit };
  };
}

// Clone side
  axes?: {
    x?: { /* ...existing... */ dispUnits?: ChartAxisDispUnits | ChartAxisDispUnit | null };
    y?: { /* ...existing... */ dispUnits?: ChartAxisDispUnits | ChartAxisDispUnit | null };
  };
}
```

## Scope rules

- The OOXML schema places `<c:dispUnits>` exclusively on `CT_ValAx`. The reader skips the parse on every other axis flavour (catAx / dateAx / serAx); a stray element on a corrupt template never surfaces. The writer drops the field on category axes (the X axis on bar / column / line / area) and on pie / doughnut (no axes at all); the value-axis side of bar / column / line / area (the Y axis) and both axes on scatter (both are value axes) honour the field.
- Element ordering inside `<c:valAx>` follows CT_ValAx: `<c:dispUnits>` lands as the last child, after `<c:minorUnit>`.
- The OOXML accepts the nine `ST_BuiltInUnit` tokens (`hundreds`, `thousands`, `tenThousands`, `hundredThousands`, `millions`, `tenMillions`, `hundredMillions`, `billions`, `trillions`); unknown tokens drop to `undefined` rather than fabricate a value the schema would reject.
- `showLabel: true` emits a bare `<c:dispUnitsLbl/>` so Excel paints its automatic annotation alongside the axis. The rich-text label customization (`<a:p>` / `<a:r>` inside `<c:dispUnitsLbl>`) is intentionally not surfaced — callers needing a custom label string can layer it on later.
- The custom-divisor variant `<c:custUnit val=".."/>` is intentionally not surfaced. A template that uses it parses as `dispUnits: undefined` and a clone that wants to author one needs the built-in preset path.
- The clone layer drops the X-axis preset on a flatten to anything other than scatter (e.g. scatter -> column flattens drop the X-axis preset because catAx rejects the element).

## Test plan

- [x] `pnpm exec vitest run --dir test` — 3533 tests passing (87 new tests for parseChart / writeChart / cloneChart).
- [x] `pnpm typecheck` — tsgo clean.
- [x] `pnpm exec oxlint src test` — 0 errors on changed files.
- [x] `pnpm exec oxfmt --check src test` — clean.
- [x] `parseChart` covers built-in unit on valAx, showLabel surfacing via `<c:dispUnitsLbl>`, scope drop on catAx, unknown ST_BuiltInUnit token rejection, missing / malformed `<c:builtInUnit>`, custUnit variant ignored, both axes on scatter, absence default.
- [x] `writeChart` covers absence default, single-string shorthand, full object form, showLabel emit / drop, unknown token rejection, OOXML element ordering (after `<c:minorUnit>`), category-axis scope drop, both scatter axes, parseChart round-trip, single-emit guard, end-to-end packaging via `writeXlsx`.
- [x] `cloneChart` covers chart-level inherit / null / wholesale-replace, addition on a source that lacked the field, shorthand override, unknown-token drop, scope drops on flatten to pie / column from scatter, scatter -> scatter carry-through, end-to-end through `parseChart -> cloneChart -> writeChart` plus packaging via `writeXlsx`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -762,6 +762,26 @@ emit order. The OOXML default `crosses: "autoZero"` collapses to
 distinct from the `"autoZero"` default which defers to Excel's
 auto-placement). Unknown semantic tokens and non-numeric `val`
 attributes drop rather than fabricate values the writer would reject.
+`ChartAxisInfo.dispUnits` surfaces the per-value-axis display-unit
+preset pulled from `<c:valAx><c:dispUnits><c:builtInUnit val=".."/></c:dispUnits></c:valAx>`
+— Excel's "Format Axis -> Display units" dropdown. The element rescales
+the numeric tick labels by the chosen preset (e.g. `"millions"` divides
+every label by 1e6) so a chart whose source range stores raw amounts
+can show compact tick labels without modifying the underlying cells.
+The OOXML schema places the element exclusively on `CT_ValAx`, so
+`dispUnits` only surfaces on the value-axis side of bar / column / line
+/ area charts (the Y axis) and on both axes of scatter charts (both are
+value axes); category axes (`<c:catAx>`) and pie / doughnut never
+surface the field. The reader keeps the parsed `unit` token and the
+presence of `<c:dispUnitsLbl>` (`showLabel: true` when Excel paints its
+automatic annotation alongside the axis); the alternative
+`<c:custUnit val=".."/>` (custom numeric divisor) and any rich-text
+`<c:dispUnitsLbl>` body are intentionally not surfaced. The OOXML
+schema accepts the nine `ST_BuiltInUnit` tokens (`"hundreds"`,
+`"thousands"`, `"tenThousands"`, `"hundredThousands"`, `"millions"`,
+`"tenMillions"`, `"hundredMillions"`, `"billions"`, `"trillions"`);
+unknown tokens drop to `undefined` rather than fabricate a value the
+writer would never emit.
 `Chart.roundedCorners` surfaces the chart-frame
 `<c:chartSpace><c:roundedCorners val=".."/>` flag — Excel's "Format
 Chart Area → Border → Rounded corners" toggle. The element sits on
@@ -912,7 +932,7 @@ charts; `lineGrouping` and `areaGrouping` accept
 `top` / `bottom` / `left` / `right` / `topRight` / `false`, and
 `altText` / `frameTitle` flow through to the drawing's `xdr:cNvPr`
 attributes for screen readers.
-`axes: { x: { title, gridlines, scale, numberFormat, majorTickMark, minorTickMark, tickLblPos, reverse, crosses, crossesAt }, y: { title, gridlines, scale, numberFormat, majorTickMark, minorTickMark, tickLblPos, reverse, crosses, crossesAt } }`
+`axes: { x: { title, gridlines, scale, numberFormat, majorTickMark, minorTickMark, tickLblPos, reverse, crosses, crossesAt, dispUnits }, y: { title, gridlines, scale, numberFormat, majorTickMark, minorTickMark, tickLblPos, reverse, crosses, crossesAt, dispUnits } }`
 attaches per-axis labels, gridlines, numeric scaling, the tick-label
 number format and the tick-rendering trio — `x` lands inside
 `<c:catAx>` (or the X value axis for scatter), `y` inside the value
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
## Review

Reviewed the read / write / clone-through implementation for `<c:dispUnits>` on value axes.

**Correctness**
- Reader scope is gated correctly: `parseAxisDispUnits` is only invoked when `axis.local === "valAx"`, which mirrors the OOXML schema (the element exists exclusively on `CT_ValAx`). A stray element on `<c:catAx>` / `<c:dateAx>` / `<c:serAx>` from a corrupt template never surfaces.
- Built-in unit token validation goes through a single shared `ReadonlySet<ChartAxisDispUnit>` in each layer (reader, writer, clone). Unknown tokens drop to `undefined` rather than fabricate a value the OOXML `ST_BuiltInUnit` enum would reject.
- The `<c:custUnit>` variant and the rich-text `<c:dispUnitsLbl>` body are intentionally not surfaced; the docstrings call this out explicitly. A `<c:dispUnits>` element carrying only `<c:custUnit val="500"/>` collapses to `undefined`, which is consistent with the "ignore what we cannot round-trip" stance.
- Element ordering inside `<c:valAx>` is correct: `buildAxisDispUnits(...)` is appended after `buildAxisTickUnits(...)` (which emits `<c:majorUnit>` / `<c:minorUnit>`), satisfying CT_ValAx's `<c:dispUnits>` "last child" position.
- Scatter is handled correctly on both sides — both axes route through `<c:valAx>` so `xDispUnits` and `yDispUnits` both pick up; the bar / column / line / area builder ignores `xDispUnits` since the X axis is `<c:catAx>`. The clone resolver collapses the X-axis override to `undefined` when the resolved type is non-scatter so a stale hint never leaks into the writer.

**Edge cases**
- `showLabel: false` and `showLabel: undefined` both omit the `<c:dispUnitsLbl/>` element. Only `showLabel === true` emits.
- Pie / doughnut have no axes; the writer short-circuits and never builds `<c:valAx>` for those families, so a pinned `dispUnits` is silently ignored.
- `null` override grammar matches the rest of the clone surface (undefined inherits, null drops, value replaces).
- The single-emit guard test (`occurrences === 2` for opening + closing tag) protects against a regression that double-emits.

**Test coverage**
- 31 new tests across `parseChart`, `writeChart`, `cloneChart` covering: built-in token surfacing, `showLabel` round-trip, scope drops on `catAx` / `dateAx` / `serAx`, unknown-token rejection, missing / malformed `<c:builtInUnit>`, `<c:custUnit>` ignored, both scatter axes, OOXML element ordering, scatter -> column flatten dropping the X-axis preset, end-to-end packaging through `writeXlsx`, and the parse -> clone -> write round-trip.

**Conflict resolution**
- The PR was conflicting against main because the chart-style-preset describe blocks landed at the end of `test/charts.test.ts`, `test/charts-write.test.ts`, and `test/chart-clone.test.ts` — the same place this PR appended its own dispUnits describe blocks. Resolved by keeping both block sequences side-by-side.

**Local validation**
- `pnpm test` (lint + typecheck + 3563 vitest tests) passes.
- `pnpm build` (obuild) succeeds.
- `oxfmt --check` clean; `oxlint` reports only pre-existing warnings (none introduced by this PR).

LGTM. Pending CI completion before merge.

PR #230, opened 2026-05-03
feat(xlsx): chart editing locale (lang) — read, write, clone-through

productdevbook:
## Summary

Surfaces the chart-space `<c:lang>` element at the read, write, and clone layers so a template's editing-locale hint survives `parseChart -> cloneChart -> writeXlsx` and can be authored from scratch on a fresh chart.

`<c:lang val=".."/>` is the IETF BCP-47 culture name Excel records on every authored chart (e.g. `en-US`, `tr-TR`, `zh-Hans-CN`). Excel uses the value to drive locale-sensitive defaults inside the chart — decimal / group separators on unformatted axis ticks, default text-font fallback, and the locale recorded for in-chart text runs. The element sits on `<c:chartSpace>` (a sibling of `<c:chart>`, not a child) per `CT_ChartSpace` in ECMA-376 §21.2.2.29, between `<c:date1904>` and `<c:roundedCorners>`.

Up until now hucre's writer had no slot for the element, so a template that pinned `tr-TR` flattened back to the host workbook's editing language on every parse -> clone -> write loop. Excel's reference serialization for a fresh chart authored on an English locale emits `<c:lang val="en-US"/>` — the writer here intentionally does not pin a default so an unmarked chart and an `en-US` chart do not silently diverge on roundtrip.

The model is a plain string (`lang?: string`); the writer emits the element when the value matches the BCP-47 shape `[A-Za-z]{2,3}(-[A-Za-z0-9]{2,8})*` Excel itself emits, dropping malformed / non-string values rather than emit a token Excel would reject (`<c:lang>` is `xsd:language` in the schema). The reader applies the same shape check, so absence and malformed tokens both round-trip to `undefined` identically. The clone layer follows the same `string | null | undefined` override grammar as `style` / `roundedCorners` so the chart-space toggles compose the same way at the call site.

This bridges another chart-space gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.lang);
// "tr-TR" when the template pinned it; undefined when the template
// omitted <c:lang> entirely.

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      lang: "tr-TR",
    }],
  }],
});

// -- Clone-through --
  // lang: undefined  -> inherit
  // lang: null       -> drop (no element emitted; Excel falls back to
  //                            workbook editing language)
  // lang: "de-DE"    -> replace
});
```

## Test plan

- [x] `pnpm test` (lint + typecheck + 3563 vitest tests, all passing)
- [x] `pnpm build`
- [x] New unit tests cover: writer emission + ordering, reader shape filter, clone inherit / null / replace / flatten, end-to-end packaging round-trip, and composition with the existing `roundedCorners` / `style` chart-space toggles

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1163,6 +1163,31 @@ export interface SheetChart {
    */
   style?: number;
   /**
+   * Editing-locale hint. Maps to `<c:lang val=".."/>` on
+   * `<c:chartSpace>` (a sibling of `<c:chart>`, not a child). The
+   * value is an RFC-1766 / IETF BCP-47 culture name such as `en-US`,
+   * `tr-TR`, or `de-DE` — Excel uses it to drive locale-sensitive
+   * defaults within the chart (decimal / group separators on
+   * unformatted axis ticks, default text font fallback, and the
+   * locale recorded for any in-chart text runs).
+   *
+   * Default: omitted — when the field is absent the writer skips the
+   * element and Excel falls back to the workbook's editing language.
+   * Excel's reference serialization for a fresh chart authored on an
+   * English locale emits `<c:lang val="en-US"/>`; the writer does
+   * not pin a default so an unmarked chart and an `en-US` chart do
+   * not silently diverge on roundtrip.
+   *
+   * Useful when restamping a templated chart for a different locale,
+   * or carrying a translated dashboard's `tr-TR` / `de-DE` hint
+   * through the parse → clone → write loop. Only well-formed culture
+   * names are emitted — unrecognized shapes are silently dropped
+   * rather than emit a token Excel would reject. The token has to
+   * match `[A-Za-z]{2,3}(-[A-Za-z0-9]{2,8})*` (the IETF language tag
+   * subset `<c:lang>` accepts under `xsd:language`).
+   */
+  lang?: string;
+  /**
    * Per-axis configuration rendered alongside the plot area. The `x`
    * axis is the category axis for bar/column/line/area (or the bottom
    * value axis for scatter); the `y` axis is the value axis. Ignored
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed the diff and the rendered file layout — looks clean and consistent with the rest of the chart-space toggle family (`style`, `roundedCorners`, `plotVisOnly`).

What I checked:
- Schema ordering: writer threads `<c:lang>` before `<c:roundedCorners>` and before `<c:style>`, matching CT_ChartSpace (`date1904?, lang?, roundedCorners?, ..., style?`). Verified by the explicit ordering tests in `charts-write.test.ts`.
- Reader / writer symmetry: both apply the same BCP-47 shape regex `^[A-Za-z]{2,3}(-[A-Za-z0-9]{2,8})*$`, so absence and malformed tokens (`english`, `en_US`, `en-`, `-US`, empty, etc.) round-trip to `undefined` identically — no fabricated defaults.
- Clone grammar: `string | null | undefined` mirrors `style`, with `null` dropping the inherited value at clone time and an unset writer skipping the element. Tests cover inherit / replace / null / flatten across line → column and line → doughnut.
- No-op writer behavior: when `lang` is unset the writer emits no element (no `<c:lang/>` self-close), so a fresh chart and an unmarked chart serialize identically — confirmed by the unset-round-trip test.
- Coverage: parser, writer, clone, end-to-end packaging via `writeXlsx`, and composition with `roundedCorners` + `style` are all exercised.

CI: build passing on the head commit. No remaining concerns; merging.

PR #225, opened 2026-05-03
feat(xlsx): chart axis crosses / crossesAt — read, write, clone-through

productdevbook:
## Summary

Surfaces the per-axis `<c:crosses val=\"..\"/>` and `<c:crossesAt val=\"..\"/>` elements at the read, write, and clone layers so a template's axis-crossing pin survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:crosses>` and `<c:crossesAt>` mirror Excel's "Format Axis -> Axis Options -> Vertical/Horizontal axis crosses" toggle. `crosses` accepts the OOXML `ST_Crosses` tokens (`\"autoZero\"` / `\"min\"` / `\"max\"`) and `crossesAt` accepts a literal numeric pin (e.g. `crossesAt: 100` to put the perpendicular axis at value 100). The two elements live in an XSD choice in `CT_CatAx` / `CT_ValAx` / `CT_DateAx` / `CT_SerAx` — only one may legally appear at a time — so the writer favours `crossesAt` when both are set, mirroring how the reader handles the same pair on a malformed template. `crossesAt: 0` is preserved (a numeric pin at zero, distinct from the `\"autoZero\"` default which defers to Excel's auto-placement).

Up until now hucre's writer always pinned `<c:crosses val=\"autoZero\"/>`, so a template with a non-default crossing pin flattened back to the default on every parse -> clone -> write loop. This bridges another axis-configuration gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts
import { readXlsx, writeXlsx, cloneChart } from \"hucre\";

// -- Read side --
const wb = await readXlsx(templateBytes);
const source = wb.sheets[0].charts![0];
console.log(source.axes?.y?.crosses);   // \"max\" when the template pinned val=\"max\"
console.log(source.axes?.y?.crossesAt); // 50 when the template pinned <c:crossesAt val=\"50\"/>

// -- Write side --
await writeXlsx({
  sheets: [{
    name: \"Dashboard\",
    rows: [...],
    charts: [{
      type: \"column\",
      series: [{ values: \"B2:B10\", categories: \"A2:A10\" }],
      axes: {
        x: { crosses: \"min\" },     // category axis crosses at the low end
        y: { crossesAt: 0 },        // value axis crosses at zero (numeric pin)
      },
      anchor: { from: { row: 6, col: 0 } },
    }],
  }],
});

// -- Clone-through --
const clone = cloneChart(source, {
[...]

PR #231, opened 2026-05-03
feat(xlsx): chart axis crossBetween — read, write, clone-through

productdevbook:
## Summary

Surfaces the per-value-axis `<c:crossBetween>` element at the read, write, and clone layers so a template's cross-between mode survives `parseChart -> cloneChart -> writeXlsx` and can be authored from scratch on a fresh chart.

`<c:crossBetween>` is the OOXML `ST_CrossBetween` toggle (`"between"` / `"midCat"`) that controls whether the perpendicular axis crosses BETWEEN data points (a half-category gap on each end of the plot area) or AT the midpoint of each category (data points sit ON the perpendicular-axis ticks rather than between them). The element is value-axis-only per ECMA-376 Part 1, §21.2.2.10 (CT_ValAx → CT_CrossBetween) — `<c:catAx>` / `<c:dateAx>` / `<c:serAx>` reject it.

Excel emits a per-family default on every freshly-drawn chart: `"between"` on bar / column / line / area Y axes and `"midCat"` on both scatter axes. The reader collapses the family-default value to `undefined` so absence and the default round-trip identically through `cloneChart`; only a non-default override (e.g. a template that pinned `"midCat"` on a line chart's Y axis to land the first / last data point flush with the value-axis line) surfaces. The writer pins the family default when the caller leaves `crossBetween` unset so untouched charts still match Excel's reference serialization byte-for-byte.

This bridges another axis-rendering gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.axes?.y?.crossBetween);
// "midCat" when the template pinned a non-default value on the value
// axis; undefined when the template inherited the family default.

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "line",
      // Land the first / last data point flush with the value-axis line.
      axes: { y: { crossBetween: "midCat" } },
    }],
  }],
});

// -- Clone-through --
  axes: {
    y: {
      // crossBetween: undefined  -> inherit
      // crossBetween: null       -> drop (writer falls back to the
      //                                   per-family default)
      // crossBetween: "between"  -> replace
      crossBetween: "between",
    },
  },
});
```

## Model

```ts
export type ChartAxisCrossBetween = "between" | "midCat";

// Read side
  // ...existing fields...
  crossBetween?: ChartAxisCrossBetween;
}

// Write side
export interface SheetChart {
  axes?: {
    x?: { /* ...existing... */ crossBetween?: ChartAxisCrossBetween };
    y?: { /* ...existing... */ crossBetween?: ChartAxisCrossBetween };
  };
}

// Clone side
  axes?: {
    x?: { /* ...existing... */ crossBetween?: ChartAxisCrossBetween | null };
    y?: { /* ...existing... */ crossBetween?: ChartAxisCrossBetween | null };
  };
}
```

## Scope rules

- The OOXML schema places `<c:crossBetween>` exclusively on `CT_ValAx`. The reader skips the parse on every other axis flavour (catAx / dateAx / serAx); a stray element on a corrupt template never surfaces. The writer drops the field on category axes (the X axis on bar / column / line / area) and on pie / doughnut (no axes at all); the value-axis side of bar / column / line / area (the Y axis) and both axes on scatter (both are value axes) honour the field.
- Element ordering inside `<c:valAx>` follows CT_ValAx: `<c:crossBetween>` lands after `<c:crosses>`/`<c:crossesAt>` and before `<c:majorUnit>`.
- The OOXML accepts the two `ST_CrossBetween` tokens (`"between"`, `"midCat"`); unknown tokens drop to the per-family default rather than fabricate a value the schema would reject.
- The writer always emits the element because Excel's reference serialization requires it on every `<c:valAx>`. The per-family default (`"between"` on bar / column / line / area Y axes; `"midCat"` on scatter axes) keeps untouched charts byte-for-byte identical to Excel's output.
- The reader's family-default collapse mirrors the writer so a chart that inherited Excel's default round-trips minimally through `cloneChart`. The clone layer drops the X-axis override on a flatten to anything other than scatter (e.g. scatter -> column flattens drop the X-axis pin because catAx rejects the element).

## Test plan

- [x] `pnpm test` (lint + typecheck + 3595 vitest tests, all passing — 32 new tests)
- [x] `pnpm build`
- [x] `parseChart` covers default-collapse on column (`"between"`) and scatter (`"midCat"`), non-default override surfacing on both families, scope drop on catAx, unknown ST_CrossBetween token rejection, missing val attribute, absence default.
- [x] `writeChart` covers absence default per family, single-token override, OOXML element ordering (after `<c:crosses>`, before `<c:majorUnit>`), category-axis scope drop, pie / doughnut scope drop, both scatter axes, parseChart round-trip including default-collapse, single-emit guard, end-to-end packaging via `writeXlsx`.
- [x] `cloneChart` covers chart-level inherit / null / wholesale-replace, addition on a source that lacked the field, unknown-token drop, scope drops on flatten to pie / column from scatter, scatter -> scatter carry-through, composition with other axis fields (crosses, dispUnits) without dropping unrelated state, end-to-end through `parseChart -> cloneChart -> writeChart` plus packaging via `writeXlsx`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -782,6 +782,22 @@ schema accepts the nine `ST_BuiltInUnit` tokens (`"hundreds"`,
 `"tenMillions"`, `"hundredMillions"`, `"billions"`, `"trillions"`);
 unknown tokens drop to `undefined` rather than fabricate a value the
 writer would never emit.
+`ChartAxisInfo.crossBetween` surfaces the per-value-axis cross-between
+mode pulled from `<c:valAx><c:crossBetween val=".."/></c:valAx>` — the
+OOXML `ST_CrossBetween` enum (`"between"` / `"midCat"`) that controls
+whether the perpendicular axis crosses BETWEEN data points (a half-
+category gap on each end of the plot area) or AT the midpoint of each
+category (data points sit ON the perpendicular-axis ticks). The element
+is required on every `<c:valAx>` and Excel emits a per-family default
+on every freshly-drawn chart — `"between"` on bar / column / line /
+area Y axes; `"midCat"` on both scatter axes — so the reader collapses
+the family-default value to `undefined` and only surfaces a non-default
+override (e.g. a template that pinned `"midCat"` on a line chart's Y
+axis to land the first / last data point flush with the value-axis
+line). The OOXML schema places the element exclusively on `CT_ValAx`,
+so `crossBetween` only surfaces on value-axis sides; category axes
+(`<c:catAx>`) and pie / doughnut never carry one. Unknown tokens drop
+to `undefined`.
 `Chart.roundedCorners` surfaces the chart-frame
 `<c:chartSpace><c:roundedCorners val=".."/>` flag — Excel's "Format
 Chart Area → Border → Rounded corners" toggle. The element sits on
@@ -932,7 +948,7 @@ charts; `lineGrouping` and `areaGrouping` accept
 `top` / `bottom` / `left` / `right` / `topRight` / `false`, and
 `altText` / `frameTitle` flow through to the drawing's `xdr:cNvPr`
 attributes for screen readers.
-`axes: { x: { title, gridlines, scale, numberFormat, majorTickMark, minorTickMark, tickLblPos, reverse, crosses, crossesAt, dispUnits }, y: { title, gridlines, scale, numberFormat, majorTickMark, minorTickMark, tickLblPos, reverse, crosses, crossesAt, dispUnits } }`
+`axes: { x: { title, gridlines, scale, numberFormat, majorTickMark, minorTickMark, tickLblPos, reverse, crosses, crossesAt, dispUnits, crossBetween }, y: { title, gridlines, scale, numberFormat, majorTickMark, minorTickMark, tickLblPos, reverse, crosses, crossesAt, dispUnits, crossBetween } }`
 attaches per-axis labels, gridlines, numeric scaling, the tick-label
 number format and the tick-rendering trio — `x` lands inside
 `<c:catAx>` (or the X value axis for scatter), `y` inside the value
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
## Review

Solid implementation that follows the established display-units (#229) / dispBlanksAs (#225) pattern closely. Verified the full read / write / clone path:

**Correctness**
- `<c:crossBetween>` is correctly scoped to `CT_ValAx` per ECMA-376 §21.2.2.10 — the reader gates on `axis.local === "valAx"`, the writer's catAx builder ignores `xCrossBetween`, and the resolver collapses the X-axis override on a flatten to anything other than scatter. Pie / doughnut never reach the axis builders.
- Element ordering inside `<c:valAx>` lands after `<c:crosses>`/`<c:crossesAt>` and before `<c:majorUnit>` per CT_ValAx, matching the existing `<c:dispUnits>` placement.
- The reader's family-default collapse is mirrored on both sides — `"between"` on bar / column / line / area Y axes; `"midCat"` on both scatter axes — so absence and the writer-emitted default round-trip identically through `cloneChart`.
- Unknown ST_CrossBetween tokens drop to `undefined` on read, write, and clone — the writer falls back to the family default rather than emit a value the schema would reject.

**Type safety / consistency**
- `ChartAxisCrossBetween` discriminated union mirrors `ChartAxisCrosses` / `ChartAxisDispUnit`.
- `applyCrossBetweenOverride` follows the same `undefined` / `null` / value grammar as the other override helpers in `chart-clone.ts`.

**Test coverage**
- 32 new tests across parse / write / clone covering family-default collapse on both column and scatter, non-default override surfacing, scope drops on catAx / pie, unknown / missing val rejection, OOXML element ordering, single-emit guard, scatter -> column flatten drop, scatter -> scatter carry-through, composition with `crosses` / `dispUnits`, and end-to-end packaging via `writeXlsx`.

**Fixes applied**
- Resolved CONFLICTING state by merging `origin/main` in. Conflicts were limited to the three test files where the `lang` describe block (PR #230, just merged) landed on overlapping line ranges. Preserved both blocks verbatim with `lang` first (matching main) and `crossBetween` appended after.
- `pnpm test` (lint + typecheck + 3626 vitest tests, all passing) and `pnpm build` confirmed locally.

No remaining concerns. Will merge once CI clears.

PR #232, opened 2026-05-03
feat(xlsx): chart date system (date1904) — read, write, clone-through

productdevbook:
## Summary

Surfaces the `<c:chartSpace><c:date1904>` element at the read, write, and clone layers so a template's chart-level date-system toggle survives `parseChart -> cloneChart -> writeXlsx` and can be authored from scratch.

`<c:date1904>` mirrors the host workbook's `<workbookPr date1904="1"/>` flag for chart date-axis interpretation — `true` anchors dates inside the chart to the 1904 base (Excel for Mac's legacy epoch where day 0 falls on 1904-01-01); `false` (the OOXML default) is the 1900 base. The element sits at the head of `<c:chartSpace>` per CT_ChartSpace (ECMA-376 Part 1, §21.2.2.29), before `<c:lang>` and `<c:roundedCorners>`. Until now hucre's writer skipped the element entirely and the reader ignored it, so a chart copied from a 1904-based template silently shifted by 1462 days when the host workbook used a different date system on every parse -> clone -> write loop.

The model is a plain boolean (`date1904?: boolean`); the writer emits `<c:date1904 val="1"/>` only when the chart pins `true` and skips emission for `false` / unset because `val="0"` is the OOXML default and absence renders identically. The reader collapses both the default value and absence to `undefined` so a fresh chart and a chart that pinned the default round-trip identically. The clone layer follows the same `boolean | null | boolean` override grammar as `roundedCorners` / `plotVisOnly` so the chart-space toggles compose the same way at the call site.

This bridges another chart-space gap for the dashboard composition flow tracked in #136 and closes the trio of chart-space siblings (`date1904` / `lang` / `roundedCorners`) that sit ahead of `<c:style>` in the schema sequence.

Refs #136

## API

```ts

// -- Read side --
console.log(source.date1904);
// true when the template pinned <c:date1904 val="1"/>; undefined when
// the template inherited the default (the OOXML `val="0"` collapses
// to undefined alongside absence so absence and the default round-
// trip identically).

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "line",
      // Anchor the chart's date references to the 1904 base even
      // though the host workbook uses 1900.
      date1904: true,
    }],
  }],
});

// -- Clone-through --
  // date1904: undefined  -> inherit the source's parsed value
  // date1904: null       -> drop the inherited value (writer skips
  //                          the element; Excel falls back to the
  //                          host workbook's date system)
  // date1904: true       -> pin the 1904 base
  // date1904: false      -> collapses to absence (the writer would
  //                          skip val="0" anyway because it matches
  //                          the OOXML default)
  date1904: true,
});
```

## Model

```ts
// Read side
export interface Chart {
  // ...existing fields...
  date1904?: boolean;
}

// Write side
export interface SheetChart {
  // ...existing fields...
  date1904?: boolean;
}

// Clone side
  // ...existing fields...
  date1904?: boolean | null;
}
```

## Scope rules

- The OOXML schema places `<c:date1904>` exclusively on `<c:chartSpace>` (a sibling of `<c:chart>`). The writer threads the element at the head of the chartSpace children, before `<c:lang>` / `<c:roundedCorners>` / `<c:style>` per the CT_ChartSpace sequence so a strict validator never sees the children transposed.
- The writer emits `<c:date1904 val="1"/>` only when the chart pins `true`. `false` and unset both collapse to absence because the OOXML default is `val="0"` and a re-parse of either renders identically — the minimal-emission contract every other chart-space toggle (`roundedCorners`, `style`, `lang`) follows.
- The reader accepts the OOXML truthy / falsy spellings (`"1"` / `"true"` / `"0"` / `"false"`); unknown values and missing `val` attributes drop to `undefined` rather than fabricate a flag Excel would not emit.
- `<c:date1904>` is family-agnostic — the toggle governs date interpretation across the whole chart document, so the clone layer carries the value through every flatten (line -> column / doughnut / scatter) without dropping it.

## Test plan

- [x] `pnpm test` (lint + typecheck + vitest under `--dir=test`, all 3625 tests passing — 31 new tests covering reader / writer / clone)
- [x] `pnpm build`
- [x] `parseChart` covers `val="1"` / `val="true"` surfacing, `val="0"` / `val="false"` default-collapse, absence, missing `val`, unknown-token drop, co-existence with `<c:lang>` / `<c:roundedCorners>` / `<c:style>`.
- [x] `writeChart` covers absence default, `false` default-collapse, `<c:date1904 val="1"/>` emission, CT_ChartSpace ordering (before `<c:lang>` and `<c:roundedCorners>`), single-emit guard, every chart family (bar / column / line / pie / doughnut / area / scatter), parseChart round-trip, end-to-end packaging via `writeXlsx`.
- [x] `cloneChart` covers chart-level inherit / null / wholesale-replace, `false` collapse-to-absence, addition on a source that lacked the field, scope retention through flatten (line -> column / doughnut), composition with `lang` / `roundedCorners` / `style` without dropping unrelated state, end-to-end through `parseChart -> cloneChart -> writeChart` plus packaging via `writeXlsx`, schema-sequence ordering verification on the rendered chartSpace.

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1188,6 +1188,36 @@ export interface SheetChart {
    */
   lang?: string;
   /**
+   * Date-system hint. Maps to `<c:date1904 val=".."/>` on
+   * `<c:chartSpace>` (a sibling of `<c:chart>`, not a child). The flag
+   * mirrors the host workbook's `<workbookPr date1904="1"/>` toggle —
+   * `true` means the chart's date-axis values are interpreted under
+   * the 1904 base (Excel for Mac's legacy epoch where day 0 falls on
+   * 1904-01-01); `false` (the OOXML default) is the 1900 base.
+   *
+   * Default: omitted — when the field is absent the writer skips the
+   * element entirely and Excel falls back to the workbook's date
+   * system. Excel itself emits `<c:date1904 val="0"/>` on every
+   * authored chart even under the 1900 base; the writer does not pin
+   * that default so an unmarked chart re-parses to `undefined` (the
+   * minimal-shape contract every other chart-space toggle follows).
+   *
+   * Useful when restamping a chart from a 1904-based template into a
+   * 1900-based workbook (or vice versa) — pinning the field keeps the
+   * chart's date references anchored to the source's epoch even after
+   * the host changes. The value is emitted as `<c:date1904 val="1"/>`
+   * for `true` and skipped (rather than `val="0"`) for `false` so the
+   * writer's behavior matches absence — a re-parse drops the default
+   * back to `undefined` either way.
+   *
+   * Note: `<c:date1904>` lives on `<c:chartSpace>` (per CT_ChartSpace
+   * the element sits at the head of the sequence, before `<c:lang>`
+   * and `<c:roundedCorners>`), not inside `<c:chart>` — the toggle
+   * governs date interpretation across the whole chart document, not
+   * just the plot area.
+   */
+  date1904?: boolean;
+  /**
    * Per-axis configuration rendered alongside the plot area. The `x`
    * axis is the category axis for bar/column/line/area (or the bottom
    * value axis for scatter); the `y` axis is the value axis. Ignored
@@ -2971,6 +3001,39 @@ export interface Chart {
    * document, not just the plot area.
    */
   lang?: string;
+  /**
+   * Date-system flag pulled from `<c:chartSpace><c:date1904 val=".."/>`.
+   * Mirrors the host workbook's `<workbookPr date1904="1"/>` toggle —
+   * `true` signals that date-axis values inside the chart are
+   * interpreted under the 1904 base (Excel for Mac's legacy epoch
+   * where day 0 falls on 1904-01-01); the OOXML default `false` is
+   * the 1900 base.
+   *
+   * Surfaces `true` only when the chart pinned `<c:date1904 val="1"/>`
+   * (the non-default state). The default `val="0"` and absence both
+   * collapse to `undefined` so absence and the default round-trip
+   * identically through {@link cloneChart}. The reader accepts the
+   * OOXML truthy / falsy spellings (`"1"` / `"true"` / `"0"` /
+   * `"false"`); unknown values and missing `val` attributes drop to
+   * `undefined`.
+   *
+   * Useful when copying a chart authored on Excel for Mac (or any
+   * 1904-based template) into a 1900-based workbook — pinning the
+   * flag keeps the chart's date references anchored to the source's
+   * epoch instead of silently shifting by 1462 days when the host
+   * date system flips. Excel's reference serialization for a fresh
+   * chart authored on a 1900-based workbook emits `<c:date1904
+   * val="0"/>`, but a chart that omits the element renders identically;
+   * surfacing only the non-default value preserves the minimal-shape
+   * contract the rest of {@link Chart} follows.
+   *
+   * Note: `<c:date1904>` lives on `<c:chartSpace>` (per CT_ChartSpace
+   * the element sits at the head of the sequence, before `<c:lang>`
+   * and `<c:roundedCorners>`), not inside `<c:chart>` — the toggle
+   * governs date interpretation across the whole chart document, not
+   * just the plot area.
+   */
+  date1904?: boolean;
 }
 
 // ── Workbook ───────────────────────────────────────────────────────
--- a/src/xlsx/chart-clone.ts
+++ b/src/xlsx/chart-clone.ts
@@ -333,6 +333,26 @@ export interface CloneChartOptions {
    */
   lang?: string | null;
   /**
+   * Override `<c:date1904>` (the chart-space date-system toggle).
+   *
+   * `undefined` (or omitted) inherits the source's parsed `date1904`.
+   * `null` drops the inherited value so the writer skips the element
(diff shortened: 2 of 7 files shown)


productdevbook (commented):
## Review

Solid implementation that mirrors the chart-space toggle patterns established in #230 (lang) and the earlier roundedCorners / plotVisOnly work. Verified the full read / write / clone path:

**Correctness**
- `<c:date1904>` is correctly placed at the head of `<c:chartSpace>` per CT_ChartSpace (ECMA-376 §21.2.2.29). The writer threads it before `<c:lang>` / `<c:roundedCorners>` / `<c:style>` and before `<c:chart>`, and the test suite asserts the order both at the chart-XML level and end-to-end through writeXlsx.
- Reader / writer / clone follow the minimal-shape contract every conditional chart-space toggle (`lang`, `style`) uses: only the non-default `val="1"` surfaces. Absence and `val="0"` round-trip identically through cloneChart, and a `false` override on the clone layer collapses back to absence.
- Reader accepts `xsd:boolean` spellings (`"1"` / `"true"` / `"0"` / `"false"`); unknown tokens drop to `undefined` rather than fabricate a flag.
- The toggle is family-agnostic — it lives on `<c:chartSpace>`, sibling of `<c:chart>` — so the clone layer carries it through every flatten (line -> column / doughnut / scatter), which the tests verify.

**Type safety / consistency**
- `date1904?: boolean` on `SheetChart` and `Chart`, `boolean | null` on `CloneChartOptions`, mirrors `roundedCorners` / `plotVisOnly`.
- `resolveDate1904` (clone side) follows the same `undefined` / `null` / value grammar as the other override helpers.

**Test coverage**
- 31 new tests covering reader (val="1" / "true" / "0" / "false" / missing val / unknown tokens / co-existence with other chart-space children), writer (absence default, false collapse, val="1" emission, CT_ChartSpace ordering against lang / roundedCorners / style, single-emit guard, every chart family, parseChart round-trip, end-to-end packaging), and clone (inherit / null / wholesale-replace, false collapse-to-absence, scope retention through flatten, composition with lang / roundedCorners / style, end-to-end packaging, schema-sequence ordering).

**Minor note (non-blocking)**
- The inline writer comment claims "every other chart-space toggle follows the same minimal-emission contract" — this is true for `lang` and `style` but `roundedCorners` always emits (matching Excel's reference serialization). The new field follows the lang / style pattern rather than the roundedCorners pattern, which is the right call for a non-default-only toggle.

CI is green on the head commit and `mergeable` is true. No remaining concerns — merging.

productdevbook (commented):
## Review

Solid implementation that mirrors the chart-space toggle patterns established in #230 (lang) and the earlier roundedCorners / plotVisOnly work. Verified the full read / write / clone path:

**Correctness**
- `<c:date1904>` is correctly placed at the head of `<c:chartSpace>` per CT_ChartSpace (ECMA-376 §21.2.2.29). The writer threads it before `<c:lang>` / `<c:roundedCorners>` / `<c:style>` and before `<c:chart>`, and the test suite asserts the order both at the chart-XML level and end-to-end through writeXlsx.
- Reader / writer / clone follow the minimal-shape contract every conditional chart-space toggle (`lang`, `style`) uses: only the non-default `val="1"` surfaces. Absence and `val="0"` round-trip identically through cloneChart, and a `false` override on the clone layer collapses back to absence.
- Reader accepts `xsd:boolean` spellings (`"1"` / `"true"` / `"0"` / `"false"`); unknown tokens drop to `undefined` rather than fabricate a flag.
- The toggle is family-agnostic — it lives on `<c:chartSpace>`, sibling of `<c:chart>` — so the clone layer carries it through every flatten (line -> column / doughnut / scatter), which the tests verify.

**Type safety / consistency**
- `date1904?: boolean` on `SheetChart` and `Chart`, `boolean | null` on `CloneChartOptions`, mirrors `roundedCorners` / `plotVisOnly`.
- `resolveDate1904` (clone side) follows the same `undefined` / `null` / value grammar as the other override helpers.

**Test coverage**
- 31 new tests covering reader (val="1" / "true" / "0" / "false" / missing val / unknown tokens / co-existence with other chart-space children), writer (absence default, false collapse, val="1" emission, CT_ChartSpace ordering against lang / roundedCorners / style, single-emit guard, every chart family, parseChart round-trip, end-to-end packaging), and clone (inherit / null / wholesale-replace, false collapse-to-absence, scope retention through flatten, composition with lang / roundedCorners / style, end-to-end packaging, schema-sequence ordering).

**Fixes applied**
- The PR went CONFLICTING after #231 (axis crossBetween) merged into main because both branches appended describe blocks to the same three test files. Resolved by merging `origin/main` in: kept the date1904 blocks verbatim and appended the crossBetween blocks from main below them in each test file. README also re-synced to bring back the crossBetween documentation that main introduced.
- `pnpm test` (lint + typecheck + 165327 vitest tests, all passing) and `pnpm build` confirmed locally.

**Minor note (non-blocking)**
- The inline writer comment claims "every other chart-space toggle follows the same minimal-emission contract" — this is true for `lang` and `style` but `roundedCorners` always emits (matching Excel's reference serialization). The new field follows the lang / style pattern rather than the roundedCorners pattern, which is the right call for a non-default-only toggle.

No remaining concerns. Will merge once CI clears.

PR #233, opened 2026-05-03
feat(xlsx): chart showDLblsOverMax — read, write, clone-through

productdevbook:
## Summary

Surfaces the chart-level `<c:showDLblsOverMax>` element at the read, write, and clone layers so a template's "Show data labels for values over maximum scale" toggle survives `parseChart` → `cloneChart` → `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:showDLblsOverMax>` is the OOXML element behind Excel's "Format Axis → Labels → Show data labels for values over maximum scale" checkbox. When the box is unchecked, data labels are suppressed for any point whose value exceeds the value axis's pinned `<c:max>` and the field surfaces `false`. The element sits at the tail of `CT_Chart` (after `<c:dispBlanksAs>` and before `<c:extLst>`) per ECMA-376 Part 1, §21.2.2.29. Until now hucre's writer skipped the element entirely and `parseChart` did not surface the value, so a template that pinned a tight axis ceiling and stripped over-max labels could not round-trip through clone. This bridges another chart-level configuration gap for the dashboard composition flow tracked in #136.

## API

```ts

// ── Read side ──
console.log(source.showDLblsOverMax); // false on a chart that strips over-max labels,
                                       // undefined when the template uses the OOXML default

// ── Write side ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [
      // Strip data labels off any point whose value exceeds the pinned axis max.
      {
        type: "column",
        axes: { y: { scale: { max: 100 } } },
        showDLblsOverMax: false,
      },
    ],
  }],
});

// ── Clone-through ──
  showDLblsOverMax: false,            // replace
  // showDLblsOverMax: null,          // drop the inherited flag (writer falls back to OOXML default true)
  // showDLblsOverMax: undefined,     // inherit the source's parsed value
});
```

## Model

```ts
interface SheetChart {
  /** ...existing fields... */
  showDLblsOverMax?: boolean;
}

interface Chart {
  /** ...existing fields... */
  showDLblsOverMax?: boolean;
}

interface CloneChartOptions {
  /** ...existing fields... */
  showDLblsOverMax?: boolean | null;
}
```

The read-side `Chart.showDLblsOverMax` mirrors the same shape so a parsed value slots straight back into `cloneChart` without transformation.

## Behavior

- **Read** — `parseChart` pulls `<c:showDLblsOverMax val=".."/>` off `<c:chart>`. The OOXML default `true` collapses to `undefined` so absence and the default round-trip identically; only an explicit `val="0"` surfaces `false`. The reader accepts the OOXML truthy / falsy spellings (`"1"` / `"true"` / `"0"` / `"false"`); unknown values and missing `val` attributes drop to `undefined`.
- **Write** — The writer always emits `<c:showDLblsOverMax>` on `<c:chart>` (matching the always-emit contract `<c:plotVisOnly>` and `<c:dispBlanksAs>` follow). Absent `showDLblsOverMax`, the writer emits `val="1"` (the OOXML schema default). An explicit `chart.showDLblsOverMax === false` flips it to `val="0"`.
- **Clone** — `options.showDLblsOverMax` accepts the standard `undefined` (inherit) / `null` (drop the inherited flag, falling back to the writer's default `true`) / `boolean` (replace) grammar that mirrors `plotVisOnly` / `dispBlanksAs`. The inherited value carries through every coercion (column → pie, doughnut → line, etc.) since the element lives on `<c:chart>` and is valid on every chart family.

## Edge cases

- A chart with `showDLblsOverMax: true` round-trips identically to a chart that omits the field — the writer emits `val="1"` and the reader collapses both shapes to `undefined`.
- A chart with `showDLblsOverMax: false` round-trips as `false` because that is the non-default value.
- A `<c:showDLblsOverMax/>` element with no `val` attribute reads as `undefined` rather than fabricate a flag Excel would not emit.
- The element is placed in the spec-required slot inside `<c:chart>` — after `<c:dispBlanksAs>` and before `<c:extLst>` — so Excel's strict validator accepts the file.
- The toggle has no observable effect on a chart whose value axis auto-scales (no point exceeds the auto-computed max); it only matters when the caller pinned a tighter axis ceiling via `axes.y.scale.max`.

Refs #136

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm test` (lint + typecheck + vitest, all 3651 tests passing including 25 new `showDLblsOverMax` tests across reader, writer, and clone)
- [x] Reader, writer, and clone tests cover: defaults, true / false / absent, the `"true"` / `"false"` spellings, missing `val`, unknown tokens, OOXML element ordering inside `<c:chart>`, single-emission guard, every chart family, and a full `parseChart -> cloneChart -> writeXlsx -> parseChart` round-trip.

### DIFF
--- a/README.md
+++ b/README.md
@@ -695,6 +695,16 @@ non-default that keeps hidden cells in the chart). The reader accepts
 the OOXML truthy / falsy spellings (`"1"` / `"true"` / `"0"` /
 `"false"`); unknown values and missing `val` attributes drop to
 `undefined`.
+`Chart.showDLblsOverMax` surfaces the chart-level
+`<c:chart><c:showDLblsOverMax val=".."/>` flag — Excel's "Format Axis
+→ Labels → Show data labels for values over maximum scale" checkbox.
+The OOXML default `true` (labels render for every point) collapses to
+`undefined` so absence and `<c:showDLblsOverMax val="1"/>` round-trip
+identically; only an explicit `val="0"` surfaces `false` (the
+non-default that strips labels off any point whose value exceeds the
+pinned `<c:max>`). The reader accepts the OOXML truthy / falsy
+spellings (`"1"` / `"true"` / `"0"` / `"false"`); unknown values and
+missing `val` attributes drop to `undefined`.
 `ChartAxisInfo.tickLblSkip` and `ChartAxisInfo.tickMarkSkip` surface
 the category-axis tick-thinning knobs (`<c:catAx><c:tickLblSkip val=".."/>`
 and `<c:catAx><c:tickMarkSkip val=".."/>`). Both elements live on
@@ -1040,6 +1050,17 @@ the chart), matching Excel's reference serialization. Pin
 chart (`val="0"`). The writer always emits the element so the
 rendered intent is explicit on roundtrip — no chart family is special-
 cased.
+The chart-level `showDLblsOverMax` field maps to
+`<c:showDLblsOverMax val=".."/>` on `<c:chart>` — Excel's "Format
+Axis → Labels → Show data labels for values over maximum scale"
+checkbox. Absent it, the writer emits the OOXML default `val="1"`
+(labels render for every point regardless of the axis ceiling),
+matching Excel's reference serialization. Pin `showDLblsOverMax: false`
+to strip labels off any point whose value exceeds the pinned `<c:max>`
+(`val="0"`). The writer always emits the element so the rendered
+intent is explicit on roundtrip — no chart family is special-cased,
+since the toggle lives on `<c:chart>` and is valid on every chart
+family.
 The chart-level `roundedCorners` field maps to
 `<c:roundedCorners val=".."/>` on `<c:chartSpace>` (a sibling of
 `<c:chart>`, not a child) — Excel's "Format Chart Area → Border →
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
Reviewed the diff end-to-end against the established `plotVisOnly` / `dispBlanksAs` / `roundedCorners` patterns and the implementation looks solid:

**Correctness**
- Reader pulls `<c:showDLblsOverMax>` directly off `<c:chart>` so its OOXML position (after `<c:dispBlanksAs>`, before `<c:extLst>`) is irrelevant; the writer emits it at the spec-required slot at the tail of CT_Chart.
- The default-collapse contract (`val="1"` / absence -> `undefined`, `val="0"` -> `false`) matches `plotVisOnly` exactly, so absence and OOXML default round-trip identically.
- Clone resolver follows the standard `undefined` (inherit) / `null` (drop) / `boolean` (replace) grammar — same as `plotVisOnly` / `dispBlanksAs`.

**Edge cases / type safety**
- Reader rejects unknown tokens and missing `val` cleanly (drops to `undefined` rather than fabricating a flag).
- Both `"1"`/`"true"` and `"0"`/`"false"` spellings accepted.
- Field is valid on every chart family per OOXML, and the tests cover line→column and line→doughnut coercions to confirm the value carries through flatten paths.

**Tests**
- 25 new tests across reader, writer, and clone covering defaults, false / true / absent, alternate spellings, missing `val`, unknown tokens, single-emission guard, OOXML element ordering, every chart family, and a full `parseChart -> cloneChart -> writeXlsx -> parseChart` round-trip. Coverage is on par with the comparable plotVisOnly / dispBlanksAs PRs.

**Doc follow-up I added**
- Pushed `docs(readme): document chart showDLblsOverMax field` (commit f1b7356) — adds reader / writer / clone paragraphs to README.md alongside the existing chart-level toggle docs so the new field is discoverable from the top-level docs the way `plotVisOnly` / `roundedCorners` already are. All 3651 tests still green locally with the doc change.

Pending CI on the doc commit; will merge once green.

PR #234, opened 2026-05-03
feat(xlsx): chart data table — read, write, clone-through

productdevbook:
## Summary

Surfaces the chart-level `<c:plotArea><c:dTable>...</c:dTable></c:plotArea>` element (Excel's "Add Chart Element -> Data Table" toggle) at the read, write, and clone layers so a template's data-table configuration survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:dTable>` paints a small table of the underlying series values beneath the plot area for a quick read of the numbers behind the picture — the same surface Excel exposes under "Format Data Table" with four boolean checkboxes for the borders, outline, and legend keys. The OOXML schema (`CT_DTable`) requires those four children: `<c:showHorzBorder>`, `<c:showVertBorder>`, `<c:showOutline>`, and `<c:showKeys>`. The element sits inside `<c:plotArea>` after the axes per `CT_PlotArea` (ECMA-376 Part 1, §21.2.2.145), so only chart families that have axes (bar / column / line / area / scatter) carry a slot for it; pie / doughnut have no axes at all and reject the element.

Until now hucre's writer skipped the element entirely and the reader ignored it, so a template that pinned a data table beneath the chart silently lost it on every parse → clone → write loop. This bridges another chart-level configuration gap for the dashboard composition flow tracked in #136.

## API

```ts

// ── Read side ──
console.log(source.dataTable);
// { showHorzBorder: true, showVertBorder: true, showOutline: true, showKeys: false }

// ── Write side — boolean shorthand (every flag true) ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [
      {
        type: "column",
        dataTable: true,
      },
    ],
  }],
});

// ── Write side — object form with per-flag overrides ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [
      {
        type: "line",
        series: [...],
        // Paint the table grid but hide the legend swatches because
        // the chart already carries a separate legend.
        dataTable: { showKeys: false },
      },
    ],
  }],
});

// ── Clone-through ──
  dataTable: { showKeys: false, showOutline: false }, // replace wholesale
  // dataTable: true,                                  // pin every flag true
  // dataTable: null,                                  // drop the inherited block
  // dataTable: false,                                 // alias for null
  // dataTable: undefined,                             // inherit the source's parsed value
});
```

## Model

```ts
interface ChartDataTable {
  showHorzBorder?: boolean;
  showVertBorder?: boolean;
  showOutline?: boolean;
  showKeys?: boolean;
}

interface SheetChart {
  /** ...existing fields... */
  dataTable?: boolean | ChartDataTable;
}

interface Chart {
  /** ...existing fields... */
  dataTable?: ChartDataTable;
}

interface CloneChartOptions {
  /** ...existing fields... */
  dataTable?: ChartDataTable | boolean | null;
}
```

The read-side `Chart.dataTable` always surfaces the object shape so a parsed value slots straight back into `cloneChart` without transformation. The write-side `SheetChart.dataTable` accepts the boolean shorthand for ergonomics — `dataTable: true` is equivalent to `dataTable: {}` and yields the same on-the-wire output (every flag `true`).

## Behavior

- **Read** — `parseChart` pulls `<c:dTable>` off `<c:plotArea>`. Each of the four boolean children round-trips literally because all four are required on `CT_DTable` and Excel always emits them; the reader does not collapse any per-field default. Children that are missing or carry an unknown `val` attribute drop to `undefined` rather than fabricate a flag the file did not pin (the writer falls back to the OOXML reference defaults — every flag `true` — on round-trip). Absence of the `<c:dTable>` element collapses the whole `dataTable` field to `undefined`.
- **Write** — The writer emits `<c:dTable>` only when the chart pins a non-suppression value (`true` or an object); `undefined` and `false` skip the element. When emitted, all four boolean children are serialized in `CT_DTable` schema order (`showHorzBorder`, `showVertBorder`, `showOutline`, `showKeys`), each falling back to `true` for any unspecified field. The element is silently dropped on `pie` / `doughnut` charts because those families have no axes and the OOXML schema places no slot for `<c:dTable>` on their plot areas.
- **Clone** — `options.dataTable` accepts `undefined` (inherit) / `null` (drop) / `false` (drop, alias for `null`) / `true` (pin every flag default `true`) / object (replace wholesale, no per-field merge). The clone layer drops the field on pie / doughnut targets so a templated bar / line / scatter chart with a pinned data table does not leak the element into a doughnut clone whose schema rejects it.

## Edge cases

- A chart with `dataTable: { showHorzBorder: true, showVertBorder: true, showOutline: true, showKeys: true }` round-trips identically to a chart that uses `dataTable: true` — the writer emits the same XML and the reader surfaces the same shape.
- `dataTable: false` and `dataTable: null` (in the clone API) both collapse to absence on the writer side. The two are kept distinct in the API surface for ergonomic alignment with the `boolean | object` shape but produce the same XML.
- An empty object (`dataTable: {}`) accepts every default — equivalent to `dataTable: true`.
- A `<c:dTable>` element with every boolean child malformed (no `val` attribute) parses as an empty `ChartDataTable` (`{}`), preserving the gating intent — the file is requesting a data table even if every flag is malformed.
- The element is placed in the spec-required slot inside `<c:plotArea>` — after the last axis (`</c:catAx>` / `</c:valAx>`) and before the optional `<c:spPr>` — so Excel's strict validator accepts the file.
- Coercing a line / column source with a pinned data table to a `pie` or `doughnut` clone drops the table because pie / doughnut have no axes; the clone layer handles the coercion silently rather than emit XML Excel would refuse.

Refs #136

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm vitest run --dir=test` (3698 tests passing including 35 new `dataTable` tests across reader, writer, and clone)
- [x] Reader, writer, and clone tests cover: defaults, the four-flags-`true` shape, mixed (per-flag) shapes, all-flags-`false`, the `"true"` / `"false"` spellings, missing `val`, unknown tokens, OOXML element ordering inside `<c:plotArea>`, single-emission guard, every chart family (with the pie / doughnut drop), pie / doughnut coercion drops, and a full `parseChart -> cloneChart -> writeXlsx -> parseChart` round-trip.

### DIFF
--- a/README.md
+++ b/README.md
@@ -705,6 +705,20 @@ non-default that strips labels off any point whose value exceeds the
 pinned `<c:max>`). The reader accepts the OOXML truthy / falsy
 spellings (`"1"` / `"true"` / `"0"` / `"false"`); unknown values and
 missing `val` attributes drop to `undefined`.
+`Chart.dataTable` surfaces the chart-level
+`<c:plotArea><c:dTable>...</c:dTable></c:plotArea>` block — Excel's
+"Add Chart Element → Data Table" toggle (the small grid of underlying
+series values painted beneath the plot area). The reader returns a
+`ChartDataTable` object with up to four boolean children
+(`showHorzBorder`, `showVertBorder`, `showOutline`, `showKeys`); each
+flag round-trips literally because all four are required on
+`CT_DTable` and Excel always emits every one. Children with missing or
+unknown `val` attributes drop to `undefined` rather than fabricate a
+flag the file did not pin, and absence of the `<c:dTable>` element
+collapses the whole field to `undefined`. Only chart families with
+axes (`bar`, `column`, `line`, `area`, `scatter`) carry a slot for the
+element — the OOXML schema places `<c:dTable>` inside `<c:plotArea>`
+alongside the axes, so pie / doughnut surface `undefined`.
 `ChartAxisInfo.tickLblSkip` and `ChartAxisInfo.tickMarkSkip` surface
 the category-axis tick-thinning knobs (`<c:catAx><c:tickLblSkip val=".."/>`
 and `<c:catAx><c:tickMarkSkip val=".."/>`). Both elements live on
@@ -1061,6 +1075,21 @@ to strip labels off any point whose value exceeds the pinned `<c:max>`
 intent is explicit on roundtrip — no chart family is special-cased,
 since the toggle lives on `<c:chart>` and is valid on every chart
 family.
+The chart-level `dataTable` field maps to
+`<c:plotArea><c:dTable>...</c:dTable></c:plotArea>` — Excel's "Add
+Chart Element → Data Table" toggle. The field accepts a
+`boolean | ChartDataTable` discriminated union: pass `true` for
+Excel's reference defaults (every border drawn, the outline frame on,
+and the legend keys shown next to each series row), pass an object to
+opt individual children in or out (each field maps to the matching
+`<c:dTable>` boolean child — `showHorzBorder`, `showVertBorder`,
+`showOutline`, `showKeys`), or omit it (or pass `false`) to suppress
+the element entirely. When emitted, the writer always serializes all
+four children in `CT_DTable` schema order, falling back to the OOXML
+reference default `true` for any flag the caller leaves unset. The
+field is silently dropped on `pie` / `doughnut` charts because those
+families have no axes and the OOXML schema places no slot for
+`<c:dTable>` on their plot areas.
 The chart-level `roundedCorners` field maps to
 `<c:roundedCorners val=".."/>` on `<c:chartSpace>` (a sibling of
 `<c:chart>`, not a child) — Excel's "Format Chart Area → Border →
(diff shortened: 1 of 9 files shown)


productdevbook (commented):
Reviewed the diff end-to-end against the established `plotVisOnly` / `dispBlanksAs` / `roundedCorners` / `showDLblsOverMax` patterns and the implementation looks solid:

**Correctness**
- Reader pulls `<c:dTable>` directly off `<c:plotArea>` and surfaces a `ChartDataTable` literally — every required boolean child round-trips because `CT_DTable` requires all four and Excel always emits them.
- Writer emits the four children (`showHorzBorder`, `showVertBorder`, `showOutline`, `showKeys`) in CT_DTable schema order and places `<c:dTable>` after the axes inside `<c:plotArea>` per CT_PlotArea (ECMA-376 §21.2.2.145), matching Excel's reference layout.
- The pie / doughnut drop is correctly handled at three layers (writer's `resolveDataTable`, clone resolver short-circuit before `resolveDataTable`, plus the absence of axes fallback) — the OOXML schema places no slot for `<c:dTable>` on those families and the implementation cleanly avoids leaking the element.
- Discriminated `boolean | object` shape on `SheetChart.dataTable` and the `boolean | object | null` clone grammar mirror the conventions already used by other chart-level block toggles.

**Edge cases / type safety**
- Reader accepts the OOXML truthy / falsy spellings (`"1"` / `"true"` / `"0"` / `"false"`) and drops unknown tokens / missing `val` to `undefined` rather than fabricating flags.
- Empty `<c:dTable>` with malformed children parses to `{}` — keeps the gating intent (the file requested a table) while letting the writer fall back to OOXML defaults on round-trip.
- Stray non-boolean inputs in writer's `resolveDataTable` (e.g. `null`, `undefined` per-field) collapse to the default `true` rather than emit invalid XML.

**Tests**
- 35 new tests across reader, writer, and clone covering defaults, every-flag-true / every-flag-false / mixed shapes, alternate spellings, missing `val`, unknown tokens, OOXML element ordering inside `<c:plotArea>`, single-emission guard, every chart family with axes, pie / doughnut drop, line→column and line→doughnut coercions, and a full `parseChart -> cloneChart -> writeXlsx -> parseChart` round-trip. Coverage is on par with the comparable plotVisOnly / dispBlanksAs / showDLblsOverMax PRs.

**Updates I added**
- Merged latest `main` into the branch (resolved automatically — only the showDLblsOverMax addition from #233 sits adjacent in the same files).
- Pushed `docs(readme): document chart dataTable field` (commit 8726b50) — adds reader / writer / clone paragraphs to README.md alongside the existing chart-level toggle docs (`plotVisOnly` / `dispBlanksAs` / `showDLblsOverMax`) so the new field is discoverable from the top-level docs the way the surrounding toggles already are.
- Local `pnpm test` (lint + typecheck + 3723 vitest tests) and `pnpm build` both green.

Pending CI on the merge + doc commit; will merge once green.

PR #235, opened 2026-05-03
feat(xlsx): chart-space protection — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:chartSpace><c:protection>...</c:protection>` (CT_Protection, ECMA-376 Part 1, §21.2.2.142) at the read, write, and clone layers so a template's chart-level lock survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:protection>` declares which interactions Excel still permits when the host worksheet is itself protected via `<sheetProtection>` — the chart inherits the sheet's protection boundary, and these flags decide whether the user can still move the frame, re-pick data ranges, retheme the look, click into chart elements, or use the floating "+ Chart Elements" / right-click affordances. The OOXML schema lists five independently optional `CT_Boolean` children: `<c:chartObject>`, `<c:data>`, `<c:formatting>`, `<c:selection>`, `<c:userInterface>`. Every flag defaults to `false` (action permitted); `true` locks the action.

The element sits on `<c:chartSpace>` between `<c:style>` / `<c:pivotSource>` and `<c:chart>` per CT_ChartSpace, so every chart family — pie / doughnut / bar / column / line / area / scatter — can carry it (the slot has no axis dependency, unlike `<c:dTable>`). Until now hucre's writer skipped the element entirely and the reader ignored it, so a chart copied from a locked-down template silently lost its protection on every parse → clone → write loop. This bridges another chart-level configuration gap for the dashboard composition flow tracked in #136.

## API

```ts

// ── Read side ──
console.log(source.protection);
// { formatting: true, selection: true }   ← only the flags the file pinned

// ── Write side — boolean shorthand (bare <c:protection> shell) ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [
      {
        type: "column",
        protection: true,
      },
    ],
  }],
});

// ── Write side — object form with per-flag overrides ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [
      {
        type: "line",
        series: [...],
        // Lock the look (formatting + selection) but leave the data
        // references and on-canvas affordances untouched.
        protection: { formatting: true, selection: true },
      },
    ],
  }],
});

// ── Clone-through ──
  protection: { chartObject: true, data: true }, // replace wholesale
  // protection: true,                            // pin the bare shell (all flags false)
  // protection: null,                            // drop the inherited block
  // protection: false,                           // alias for null
  // protection: undefined,                       // inherit the source's parsed value
});
```

## Model

```ts
interface ChartProtection {
  chartObject?: boolean;
  data?: boolean;
  formatting?: boolean;
  selection?: boolean;
  userInterface?: boolean;
}

interface SheetChart {
  /** ...existing fields... */
  protection?: boolean | ChartProtection;
}

interface Chart {
  /** ...existing fields... */
  protection?: ChartProtection;
}

interface CloneChartOptions {
  /** ...existing fields... */
  protection?: ChartProtection | boolean | null;
}
```

The read-side `Chart.protection` always surfaces the object shape so a parsed value slots straight back into `cloneChart` without transformation. The write-side `SheetChart.protection` accepts the boolean shorthand for ergonomics — `protection: true` is equivalent to `protection: {}` and yields the same on-the-wire output (the bare `<c:protection>` shell with every flag at the OOXML default `false`).

## Behavior

- **Read** — `parseChart` pulls `<c:protection>` off `<c:chartSpace>`. Each of the five boolean children is independently optional on `CT_Protection`, so the reader only surfaces the flags the file actually pinned (mirroring how every CT_Protection child is optional). Children that are missing or carry an unknown `val` attribute drop to `undefined` for that field rather than fabricate a flag the file did not pin. The element itself is the gating signal — a `<c:protection>` block with no resolvable children surfaces as `{}` (mirroring how `dataTable` handles a malformed `<c:dTable>`), so a chart that authors the bare element round-trips literally.
- **Write** — The writer emits `<c:protection>` only when the chart pins a non-suppression value (`true` or an object); `undefined` and `false` skip the element. When emitted, all five boolean children are serialized in `CT_Protection` schema order (`chartObject`, `data`, `formatting`, `selection`, `userInterface`), each falling back to `false` for any unspecified field — the always-emit contract matches `<c:plotVisOnly>` / `<c:dispBlanksAs>` so the rendered intent is explicit on roundtrip. Stray non-boolean inputs leaking through the type guard collapse to the OOXML default `false` rather than emit a token Excel rejects.
- **Clone** — `options.protection` accepts `undefined` (inherit) / `null` (drop) / `false` (drop, alias for `null`) / `true` (pin the bare shell with every flag `false`) / object (replace wholesale, no per-field merge). Unlike `dataTable`, the resolver applies to every chart family — `<c:protection>` lives on `<c:chartSpace>` (not inside `<c:plotArea>`), so pie / doughnut clones still carry the inherited block instead of stripping it.

## Notes

Excel only enforces these flags when the host worksheet is itself protected (the chart inherits the sheet's protection boundary). On an unprotected sheet the element round-trips literally but has no observable runtime effect — it sits dormant until `<sheetProtection>` activates the lock.

Refs #136

## Test plan

- [x] `pnpm test` (lint + typecheck + vitest, 88 files, 3759 tests, all passing)
- [x] `pnpm build` (obuild, 183 files, 1.17 MB)
- [x] New `parseChart — chart-space protection` describe (10 tests) covering the fully-pinned shape, partial overrides, missing / malformed `val` attributes, the bare `<c:protection/>` element, and pie-chart roundtrip
- [x] New `writeChart — chart-space protection` describe (15 tests) covering suppression, the boolean shorthand, per-field overrides, schema-order emission, the chartSpace placement, every chart family, and writeXlsx integration
- [x] New `cloneChart — chart-space protection` describe (10 tests) covering inherit / replace / drop / coerce-into-doughnut grammar plus a full template → clone → write → re-parse loop

### DIFF
--- a/README.md
+++ b/README.md
@@ -719,6 +719,22 @@ collapses the whole field to `undefined`. Only chart families with
 axes (`bar`, `column`, `line`, `area`, `scatter`) carry a slot for the
 element — the OOXML schema places `<c:dTable>` inside `<c:plotArea>`
 alongside the axes, so pie / doughnut surface `undefined`.
+`Chart.protection` surfaces the chart-space-level
+`<c:chartSpace><c:protection>...</c:protection>` block — Excel's
+chart-level lock that pairs with the host worksheet's
+`<sheetProtection>` to decide which interactions remain allowed when
+the parent sheet is protected. The reader returns a `ChartProtection`
+object with up to five independently optional boolean children
+(`chartObject`, `data`, `formatting`, `selection`, `userInterface`);
+each flag is independently optional on `CT_Protection`, so the reader
+only surfaces the flags the file actually pinned. Children with
+missing or unknown `val` attributes drop to `undefined` rather than
+fabricate a flag the file did not pin. The element itself is the
+gating signal — a `<c:protection>` block with no resolvable children
+surfaces as an empty `{}`, and absence of the element collapses the
+field to `undefined`. Every chart family carries a slot — the element
+lives on `<c:chartSpace>` (a sibling of `<c:chart>`), not inside
+`<c:plotArea>`, so pie / doughnut still surface the block.
 `ChartAxisInfo.tickLblSkip` and `ChartAxisInfo.tickMarkSkip` surface
 the category-axis tick-thinning knobs (`<c:catAx><c:tickLblSkip val=".."/>`
 and `<c:catAx><c:tickMarkSkip val=".."/>`). Both elements live on
@@ -1090,6 +1106,25 @@ reference default `true` for any flag the caller leaves unset. The
 field is silently dropped on `pie` / `doughnut` charts because those
 families have no axes and the OOXML schema places no slot for
 `<c:dTable>` on their plot areas.
+The chart-level `protection` field maps to
+`<c:chartSpace><c:protection>...</c:protection>` — Excel's chart-level
+lock that pairs with the host worksheet's `<sheetProtection>` to
+decide which interactions remain allowed when the parent sheet is
+protected. The field accepts a `boolean | ChartProtection`
+discriminated union: pass `true` for the bare `<c:protection>` shell
+with every flag at the OOXML default `false` (action permitted), pass
+an object to opt individual children in (each field maps to the
+matching `<c:protection>` boolean child — `chartObject`, `data`,
+`formatting`, `selection`, `userInterface`), or omit it (or pass
+`false`) to suppress the element entirely. When emitted, the writer
+always serializes all five children in `CT_Protection` schema order,
+falling back to the OOXML default `false` (action permitted) for any
+flag the caller leaves unset. The element sits on `<c:chartSpace>`
+(not inside `<c:plotArea>`), so every chart family — including pie /
+doughnut — carries a slot for it. Excel only enforces these flags
+when the host worksheet is itself protected; on an unprotected sheet
+the element round-trips literally but has no observable runtime
+effect.
 The chart-level `roundedCorners` field maps to
 `<c:roundedCorners val=".."/>` on `<c:chartSpace>` (a sibling of
 `<c:chart>`, not a child) — Excel's "Format Chart Area → Border →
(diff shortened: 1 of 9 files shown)


productdevbook (commented):
Reviewed the diff end-to-end and the implementation looks solid.

**What I checked**
- `parseChart` correctly gates on `<c:protection>` presence and treats every CT_Boolean child as independently optional, mirroring the precedent `dataTable` set for malformed `val` handling. The empty-shell `{}` carve-out for templates that author the bare element is the right call.
- `writeChart` always emits the five children in CT_Protection schema order (`chartObject`, `data`, `formatting`, `selection`, `userInterface`) — matches the always-emit contract `<c:plotVisOnly>` / `<c:dispBlanksAs>` follow. Stray non-boolean inputs collapse to `false`, no invalid tokens leak.
- `<c:protection>` is inserted between `<c:style>` and `<c:chart>` on `<c:chartSpace>`, which is correct CT_ChartSpace order (since hucre does not emit `clrMapOvr` / `pivotSource` between them).
- `cloneChart` resolver follows the same `undefined` (inherit) / `null` (drop) / `false` (drop alias) / `true` (bare shell) / object (replace wholesale) grammar as `dataTable`. Unlike `dataTable`, the resolver applies to every chart family, which correctly reflects that `<c:protection>` lives on `<c:chartSpace>`, not inside `<c:plotArea>`.
- 35 new tests across `charts.test.ts`, `charts-write.test.ts`, and `chart-clone.test.ts` cover the fully-pinned shape, partial overrides, missing / malformed `val`, the bare element, every chart family (including pie / doughnut / scatter), schema-order emission, the chartSpace placement, and the full template → clone → write → re-parse loop.

**Fix pushed**
- Added README sections under `### Charts` (reader, writer, and clone-grammar paragraphs) so `protection` is discoverable alongside `dataTable` / `showDLblsOverMax`. Matched the existing prose style verbatim.
- Re-ran `pnpm test` (88 files, 3759 tests, all passing) and `pnpm build` (183 files, 1.17 MB) locally — both green.

No remaining concerns. Will merge once CI completes.

PR #236, opened 2026-05-03
feat(xlsx): chart data labels number format — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:dLbls><c:numFmt formatCode=".." sourceLinked=".."/>` at the read, write, and clone layers so a template's data-label number format survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:numFmt>` inside `<c:dLbls>` mirrors Excel's "Format Data Labels -> Number" panel — pinning a custom format string (`"0.00%"`, `"$#,##0.00"`, `"#,##0"`, etc.) on the rendered label values. Until now hucre's writer skipped the element entirely and the reader ignored it, so a dashboard template that pinned currency / percent labels lost the format on every parse -> clone -> write loop.

Reuses the existing `ChartAxisNumberFormat` shape so a single number-format helper can thread through both the axis and the data-labels code paths — the write-side `ChartDataLabels.numberFormat` and the read-side `ChartDataLabelsInfo.numberFormat` share field semantics with `ChartAxisInfo.numberFormat`, so a parsed value flows straight back into the writer / clone without transformation. This bridges another chart-level configuration gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// ── Read side ──
console.log(source.dataLabels?.numberFormat);
// { formatCode: "0.00%" }   ← when the template pinned a percent format

// Per-series override surfaces on ChartSeriesInfo.dataLabels:
console.log(source.series?.[0].dataLabels?.numberFormat);
// { formatCode: "$#,##0", sourceLinked: false }

// ── Write side — chart-level ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [
      {
        type: "column",
        dataLabels: {
          showValue: true,
          numberFormat: { formatCode: "$#,##0" },
        },
      },
    ],
  }],
});

// ── Write side — sourceLinked=true (Excel inherits the source cell format) ──
dataLabels: {
  showValue: true,
  numberFormat: { formatCode: "General", sourceLinked: true },
}

// ── Write side — per-series override ──
series: [
  {
    name: "Margin",
    values: "B2:B6",
    dataLabels: {
      showValue: true,
      numberFormat: { formatCode: "0.00%" },
    },
  },
],

// ── Clone-through ──
  // dataLabels: undefined            // inherit the parsed numberFormat
  // dataLabels: null                 // drop the inherited block (and its numFmt)
  // dataLabels: { showValue: true,
  //   numberFormat: { formatCode: "0.00%" } }  // wholesale replace
});
```

## Model

```ts
interface ChartDataLabels {
  // ...existing fields...
  numberFormat?: ChartAxisNumberFormat;
}

interface ChartDataLabelsInfo {
  // ...existing fields...
  numberFormat?: ChartAxisNumberFormat;
}
```

The read-side `ChartDataLabelsInfo.numberFormat` mirrors the write-side `ChartDataLabels.numberFormat` so a parsed value slots straight back into `cloneChart` without transformation. Same shape as the existing axis-side `ChartAxisInfo.numberFormat` — the writer / parser / clone all reuse the existing `ChartAxisNumberFormat` interface (`{ formatCode: string; sourceLinked?: boolean }`).

## Behavior

- **Read** — `parseChart` pulls the value off `<c:dLbls><c:numFmt formatCode=".." sourceLinked=".."/>`. `formatCode` is required; absence of the element, missing / non-string / empty `formatCode` all collapse to `undefined` so a malformed pin cannot fabricate a record. `sourceLinked` accepts the OOXML truthy / falsy spellings (`"1"` / `"true"` / `"0"` / `"false"`); absence and the OOXML default `"0"` collapse to `undefined` so absence and the default round-trip identically. The lookup is scoped to direct `<c:dLbls>` children — a `<c:numFmt>` nested inside a per-point `<c:dLbl>` does not leak into the block-level record. The `numberFormat` field alone is enough to surface a `dataLabels` record (Excel still applies the format pin even when no `show*` toggle is on, since a series-level override can flip the labels on without restating the format).
- **Write** — `<c:numFmt>` sits at the head of the CT_DLbls sequence (before `<c:dLblPos>` and the show* toggles). The writer skips emission entirely when `numberFormat` is unset / `formatCode` is missing / non-string / empty, so a fresh chart matches Excel's reference shape (no override means Excel inherits from the source cells). `sourceLinked` is normalized to a literal boolean — only `true` survives, every other shape (`undefined` / `false` / non-boolean) defaults to `val="0"`. Format codes are emitted through the standard XML attribute escape path so accounting / conditional formats containing `<`, `>`, `&`, `"` round-trip cleanly.
- **Clone** — Lives inside the existing `dataLabels` clone grammar (chart-level: `undefined` inherits / `null` drops / object replaces wholesale; per-series: `undefined` inherits / `null` drops / `false` suppresses / object replaces wholesale). The inherited block carries `numberFormat` through automatically because `ChartDataLabelsInfo` and `ChartDataLabels` share the same field semantics — no separate helper needed.

## Edge cases

- A chart with no `<c:numFmt>` element parses to `numberFormat: undefined`.
- Empty / missing `formatCode` collapses to no emission rather than fabricate an attribute Excel rejects.
- A per-point `<c:dLbl><c:numFmt>` does not leak into the block-level record — the lookup is scoped to direct `<c:dLbls>` children.
- Non-string `formatCode` and non-boolean `sourceLinked` inputs collapse safely (no emission / `val="0"` default).
- The element is emitted exactly once per `<c:dLbls>` block — both chart-level and series-level can carry their own pin independently.
- All chart families (bar / column / line / pie / doughnut / area / scatter) support the field identically because `<c:dLbls>` is shared across every chart-type element.
- Format codes containing XML-special characters (e.g. `[<=9999999]"$"#,##0`) are escaped through the standard XML attribute path.
- Clone-through respects the wholesale `dataLabels` override — pinning a new `dataLabels` block on the clone drops the inherited `numberFormat` along with the rest of the parsed shape, matching how the existing `showLegendKey` / `position` / `separator` fields handle the override.

## Test plan

- [x] `pnpm vitest run --dir=test` (88 files / 3761 tests in worktree) passes.
- [x] `pnpm lint` and `pnpm typecheck` pass with no new errors.
- [x] `pnpm build` succeeds.
- [x] 11 new `parseChart` tests cover the truthy / falsy `sourceLinked` spellings, default collapse, missing `formatCode`, missing element, empty `formatCode`, surfacing-as-only-field guard, series-level threading, multi-family threading (line / pie / scatter), composition with other dLbls fields, and per-point `<c:dLbl>` scoping.
- [x] 14 new `writeChart` tests cover the default emit, `sourceLinked` true / false threading, OOXML element ordering (`numFmt` before `dLblPos` / show*), exactly-one-emit guard, missing-numberFormat / empty-formatCode / non-string-formatCode suppression, multi-family threading (pie / line / scatter), series-level threading, XML escaping for format codes with special characters, non-boolean `sourceLinked` fallback, parse-side round-trip (with and without `sourceLinked`), and end-to-end packaging via `writeXlsx`.
- [x] 11 new `cloneChart` tests cover inherit / null / wholesale-replace grammar at chart and series scope, override-pins-on-source-without-numberFormat, chart-type coercion (bar -> line), composition alongside other dLbls fields, and end-to-end through `parseChart -> cloneChart -> writeChart` plus packaging via `writeXlsx`.

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -618,6 +618,22 @@ export interface ChartDataLabels {
    * `" "`, `"; "`, `"\n"` (newline).
    */
   separator?: string;
+  /**
+   * Number format applied to the value rendered inside each data label.
+   * Mirrors Excel's "Format Data Labels -> Number" panel — pinning a
+   * `formatCode` such as `"0.00%"`, `"$#,##0.00"`, or `"#,##0"` lets a
+   * dashboard chart show currency / percent labels without the source
+   * cells carrying the format. Same shape as
+   * {@link ChartAxisNumberFormat} so a single number-format helper can
+   * thread through both the axis and the data-labels code paths.
+   *
+   * Maps to `<c:numFmt formatCode=".." sourceLinked=".."/>` inside
+   * `<c:dLbls>`. The element sits right after `<c:dLbl>` (per the
+   * `CT_DLbls` schema sequence — `dLbl* -> numFmt? -> spPr? -> txPr? ->
+   * dLblPos? -> show*`). Omit to fall back to whatever Excel inherits
+   * from the source cell formatting.
+   */
+  numberFormat?: ChartAxisNumberFormat;
 }
 
 /**
@@ -2332,6 +2348,18 @@ export interface ChartDataLabelsInfo {
   showLegendKey?: boolean;
   position?: ChartDataLabelPosition;
   separator?: string;
+  /**
+   * Mirror of {@link ChartDataLabels.numberFormat}. Surfaces the
+   * `<c:numFmt formatCode=".." sourceLinked=".."/>` parsed from the
+   * source `<c:dLbls>` block — Excel's "Format Data Labels -> Number"
+   * panel pin. Same shape as the axis-side
+   * {@link ChartAxisNumberFormat} so the parsed value can be fed
+   * straight back into {@link cloneChart} or {@link writeXlsx} without
+   * transformation. Absent when the source data-labels block has no
+   * `<c:numFmt>` element or when the parsed `formatCode` is missing /
+   * empty.
+   */
+  numberFormat?: ChartAxisNumberFormat;
 }
 
 /**
--- a/src/xlsx/chart-reader.ts
+++ b/src/xlsx/chart-reader.ts
@@ -1285,6 +1285,18 @@ function parseDataLabels(el: XmlElement): ChartDataLabelsInfo | undefined {
     if (text.length > 0) out.separator = text;
   }
 
+  // `<c:numFmt formatCode=".." sourceLinked=".."/>` mirrors Excel's
+  // "Format Data Labels -> Number" panel — pinning a custom number
+  // format on the rendered label values. Same shape as the axis-side
+  // `<c:numFmt>` so the parsed value can be fed straight back into
+  // {@link cloneChart}. The element sits early in the CT_DLbls
+  // sequence (after the optional `<c:dLbl>` instances), so the lookup
+  // is scoped to direct `<c:dLbls>` children — a `<c:numFmt>` nested
+  // inside a per-point `<c:dLbl>` does not leak into the block-level
+  // record.
+  const numFmt = parseDataLabelsNumberFormat(el);
+  if (numFmt) out.numberFormat = numFmt;
+
   // Empty record is meaningless to a consumer — collapse to undefined.
   if (
     out.position === undefined &&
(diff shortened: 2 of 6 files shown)


productdevbook (commented):
Reviewed the diff end-to-end against the existing axis-side `<c:numFmt>` plumbing — the new `numberFormat` field is a clean extension that reuses the `ChartAxisNumberFormat` shape across read / write / clone, matching the conventions established by recent chart PRs.

Strengths:
- `resolveDataLabelsNumberFormat` mirrors `normalizeAxisNumberFormat` (same empty / non-string `formatCode` collapse, same `sourceLinked === true` normalization), so a parsed value flows straight back into the writer without transformation.
- The element is correctly placed at the head of the CT_DLbls sequence (before `<c:dLblPos>` and the show* toggles), and the test asserts the ordering explicitly.
- Edge-case coverage is thorough: missing / empty / non-string `formatCode`, non-boolean `sourceLinked`, per-point `<c:dLbl>` scoping, XML escaping for accounting / conditional format codes, multi-family threading (bar / column / line / pie / scatter), end-to-end `writeXlsx` packaging, and parse → clone → write round-trips at both chart and series scope.
- The 'numberFormat alone is enough to surface a dataLabels record' guard is the right call — Excel still applies the pin when a per-series override flips labels on.

Minor stylistic note (non-blocking): the new `parseDataLabelsNumberFormat` decodes `sourceLinked` inline (`'1' || .toLowerCase() === 'true'`) instead of delegating to the existing `parseBoolAttr` helper that the axis-side parser uses. The behavioral difference is whitespace-trimming on the attribute value, which is unlikely to matter for Excel-emitted files; mentioning only because future readers may wonder why two near-identical numFmt parsers shape `sourceLinked` slightly differently. Not worth blocking the merge.

CI is green, tests pass, no concerns. Merging.

PR #237, opened 2026-05-03
feat(xlsx): chart-level line marker visibility — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:lineChart><c:marker val=".."/>` (the chart-level CT_Boolean variant of CT_LineChart `<c:marker>`, ECMA-376 Part 1 §21.2.2.103) at the read, write, and clone layers so a template's "Line vs. Line with Markers" choice survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The chart-level `<c:marker>` element gates whether per-series markers paint at all on a line chart — flipping `val="0"` collapses the chart to the bare "Line" preset look, while `val="1"` (Excel's reference for every authored line chart) lets per-series CT_Marker blocks render their symbol / size / fill. Until now hucre's writer hard-coded `val="1"` and the reader ignored the element entirely, so a templated dashboard line chart that pinned `val="0"` silently re-painted markers on every parse -> clone -> write loop. This bridges another chart-level configuration gap for the dashboard composition flow tracked in #136.

Distinct from the per-series `<c:marker>` (CT_Marker, with `<c:symbol>` / `<c:size>` / `<c:spPr>` children) hucre already supports — the chart-level gate decides whether markers paint, while the per-series block picks the symbol / size / fill that paints when the gate is open.

Refs #136

## API

```ts

// ── Read side ──
console.log(source.showLineMarkers);
// false   ← when the template pinned <c:marker val="0"/>
// undefined ← when the template emits the default <c:marker val="1"/> (or absence)

// ── Write side — suppress markers chart-wide ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [
      {
        type: "line",
        showLineMarkers: false,   // emits <c:marker val="0"/>
      },
    ],
  }],
});

// ── Write side — explicit on (matches Excel default behavior) ──
showLineMarkers: true            // emits <c:marker val="1"/>

// ── Write side — undefined / omitted ──
// (no field set)                 // emits <c:marker val="1"/> — back-compat
//                                // with every line chart hucre wrote before

// ── Clone-through ──
  // showLineMarkers: undefined            // inherit the parsed value
  // showLineMarkers: null                 // drop the inherited value
  //                                       // (writer falls back to <c:marker val="1"/>)
  // showLineMarkers: false                // wholesale replace — gate off
  // showLineMarkers: true                 // explicit gate on (collapses to default on emit)
});
```

## Model

```ts
interface SheetChart {
  // ...existing fields...
  showLineMarkers?: boolean;
}

interface Chart {
  // ...existing fields...
  showLineMarkers?: boolean;
}

interface CloneChartOptions {
  // ...existing fields...
  showLineMarkers?: boolean | null;
}
```

The read-side `Chart.showLineMarkers` and write-side `SheetChart.showLineMarkers` share field semantics so a parsed value slots straight back into `cloneChart` without transformation. The clone option mirrors `upDownBars` / `dropLines` / `hiLowLines` grammar — `undefined` inherits, `null` drops the inherited value (writer falls back to default), and a `boolean` replaces.

## Behavior

- **Read** — `parseChart` pulls the value off the first `<c:lineChart>` element's `<c:marker>` child. The OOXML / Excel default `val="1"` and absence both collapse to `undefined` so absence and the default round-trip identically through `cloneChart` — only an explicit `val="0"` surfaces `false`. Accepts the OOXML truthy / falsy spellings (`"1"` / `"true"` / `"0"` / `"false"`); unknown values, missing `val` attributes, and a missing element all drop to `undefined`. The lookup is scoped to `<c:lineChart>` only — `CT_Line3DChart` / `CT_StockChart` and every non-line family have no slot for the chart-level CT_Boolean variant per the OOXML schema, so a stray `<c:marker val=".."/>` on those bodies is ignored. Per-series `<c:marker>` (CT_Marker, sibling of `<c:idx>`) still surfaces through `ChartSeriesInfo.marker` — the chart-level gate and the per-series styling carry separately.
- **Write** — The writer always emits `<c:marker val="0|1"/>` inside `<c:lineChart>` (Excel's reference behavior — every authored line chart includes the element). Defaults to `val="1"` when `showLineMarkers` is unset, `true`, or any non-boolean leaking through the type guard; flips to `val="0"` only on a literal `false`. The element sits in CT_LineChart schema position — after `<c:dLbls>` / `<c:dropLines>` / `<c:hiLowLines>` / `<c:upDownBars>`, before `<c:axId>`. Silently dropped on every other chart family — the writer's `<c:barChart>` / `<c:areaChart>` / `<c:pieChart>` / `<c:doughnutChart>` / `<c:scatterChart>` builders never author the chart-level `<c:marker>` block.
- **Clone** — Lives inside the existing line-only chart-level grammar (mirrors `upDownBars` / `dropLines` / `hiLowLines`): `undefined` inherits / `null` drops / `boolean` replaces. The resolver gates the carry on the resolved write-side family — flattening a line template into a column / pie / doughnut / area / scatter clone drops the inherited flag silently so a host that cannot carry `<c:marker>` never picks one up.

## Edge cases

- A line chart with no `<c:marker>` parses to `showLineMarkers: undefined`.
- The Excel default `<c:marker val="1"/>` collapses to `undefined` so absence and the default round-trip identically; only the non-default `val="0"` survives.
- A bare `<c:marker/>` (no `val`) collapses to `undefined` — the writer always emits the attribute, so a real chart never emits the bare shape.
- Non-boolean `showLineMarkers` inputs (e.g. a stray `"off"` leaking through the type guard) collapse to the default `val="1"`.
- The chart-level `<c:marker>` (CT_Boolean) and per-series `<c:marker>` (CT_Marker, with `<c:symbol>` / `<c:size>` children) compose freely — both surface independently on parse, both emit independently on write.
- Stray `<c:marker val=".."/>` on `<c:line3DChart>` / `<c:stockChart>` / `<c:barChart>` / `<c:areaChart>` / `<c:pieChart>` / `<c:doughnutChart>` / `<c:scatterChart>` does not surface — the reader scopes the lookup to `<c:lineChart>` per the OOXML schema.
- Composes alongside other line-only chart-level toggles (`dropLines` / `hiLowLines` / `upDownBars`) — the writer respects the CT_LineChart schema sequence (`<c:marker>` sits at the tail before `<c:axId>`).

## Test plan

- [x] `pnpm vitest run --dir=test` (88 files / 3799 tests in worktree) passes.
- [x] `pnpm lint` and `pnpm typecheck` pass with no new errors.
- [x] `pnpm build` succeeds (1.18 MB / 183 files).
- [x] 11 new `parseChart` tests cover the non-default `val="0"` surface, default `val="1"` collapse, absence, OOXML truthy / falsy spellings, missing-attr / unknown-token drops, line3D / stock / bar / column / pie / doughnut / area / scatter no-slot guards, per-series-marker scope guard, and composition with other line-only chart-level fields.
- [x] 13 new `writeChart` tests cover the default emit, explicit `true` / `false` threading, exactly-one-emit guard, OOXML schema-order placement (`<c:marker>` after `<c:upDownBars>` / `<c:dropLines>` / `<c:hiLowLines>`, before `<c:axId>`), bar / column / pie / doughnut / area / scatter no-emit guards, non-boolean fallback, parse-side round-trip (false / undefined / true), per-series marker composition, and end-to-end packaging via `writeXlsx` (alone and alongside every line-only optional block).
- [x] 13 new `cloneChart` tests cover inherit / null / wholesale-replace grammar, override-on-source-without-flag, line → column / area / pie / doughnut / scatter flatten guards, stock → line carry, composition with other line-only toggles, and end-to-end through `parseChart -> cloneChart -> writeChart` plus packaging via `writeXlsx`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -935,6 +935,19 @@ model is a plain presence flag at this layer); absence collapses to
 `CT_LineChart` / `CT_Line3DChart` / `CT_StockChart` per the OOXML
 schema; a stray `<c:upDownBars>` on a bar / column / pie / doughnut /
 area / scatter chart is ignored.
+`Chart.showLineMarkers` surfaces the chart-level marker visibility
+flag pulled from `<c:lineChart><c:marker val=".."/></c:lineChart>` —
+Excel's "Line vs. Line with Markers" chart-type distinction. The flag
+gates whether per-series markers paint at all on a line chart;
+per-series `<c:marker>` blocks (CT_Marker, with `<c:symbol>` /
+`<c:size>`) still pick the symbol / size / fill that paints when the
+gate is open. The Excel-default `val="1"` and absence both collapse to
+`undefined` so absence and the default round-trip identically through
+{@link cloneChart} — only an explicit `val="0"` surfaces `false`. The
+chart-level slot lives exclusively on `CT_LineChart` per the OOXML
+schema; `CT_Line3DChart` / `CT_StockChart` and every non-line family
+have no slot for the chart-level CT_Boolean variant, so a stray
+`<c:marker val=".."/>` on those bodies is ignored.
 Sheets that hucre actively regenerates because they
 also carry hucre-managed images currently keep the chart bodies but
 lose the in-drawing chart anchor — merging hucre's drawing output
@@ -1192,6 +1205,21 @@ omits the element on a fresh chart. Pin `upDownBars: true` on a
 every other chart family — `<c:upDownBars>` lives exclusively on
 `CT_LineChart` / `CT_Line3DChart` / `CT_StockChart` per the OOXML
 schema, and the writer never authors the 3D / stock variants.
+The chart-level `showLineMarkers` field maps to `<c:marker val=".."/>`
+inside `<c:lineChart>` — the chart-level CT_Boolean variant that gates
+whether per-series markers paint at all on a line chart (Excel's
+"Line vs. Line with Markers" chart-type distinction). The writer
+always emits the element (Excel's reference behavior — every authored
+line chart includes one), defaulting to `val="1"` when the field is
+unset or `true`. Pin `showLineMarkers: false` to suppress markers
+chart-wide so the line renders as a clean stroke without the
+per-point dots. The flag is silently ignored on every other chart
+family — the OOXML schema places the chart-level `<c:marker>`
+exclusively on `CT_LineChart`, and `CT_Line3DChart` / `CT_StockChart`
+and every non-line family have no slot for it. Independent of any
+per-series marker styling: this gate sits at the chart level and
+decides whether markers paint, while the per-series block picks the
+symbol / size / fill that paints when the gate is open.
 The `axes.{x,y}.dispUnits` field controls the per-value-axis
 display-unit preset (`<c:valAx><c:dispUnits><c:builtInUnit val=".."/></c:dispUnits></c:valAx>`)
 — Excel's "Format Axis -> Display units" dropdown. Each numeric tick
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
Reviewed the diff end-to-end. The implementation cleanly mirrors the existing line-only chart-level grammar (dropLines / hiLowLines / upDownBars) at all three layers:

- Reader scopes the lookup to <c:lineChart>, correctly rejecting CT_Line3DChart / CT_StockChart and every non-line family per the OOXML schema for the chart-level CT_Boolean variant of <c:marker>.
- Writer emits <c:marker val="0|1"/> in the correct CT_LineChart sequence position (after upDownBars, before axId+) and respects Excel's always-emit reference behavior so back-compat output is unchanged.
- Clone resolver follows the canonical undefined-inherits / null-drops / boolean-replaces grammar and gates the carry on the resolved chart type.

The asymmetric collapse contract (default val="1" and absence both surface as undefined; only the non-default val="0" surfaces false) is the right call for round-trip symmetry. Per-series <c:marker> (CT_Marker) handling is correctly kept independent. Test coverage is thorough — 11 reader + 13 writer + clone-through tests cover defaults, OOXML truthy/falsy spellings, schema-scope guards on every other family, exactly-once emit, schema-order placement, and end-to-end round-trips through writeXlsx.

CI is green, no concerns. Merging.

PR #238, opened 2026-05-03
feat(xlsx): chart series lines — read, write, clone-through

productdevbook:
## Summary

Surfaces the chart-type-level `<c:serLines/>` element at the read, write, and clone layers so a template's series-lines configuration survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:serLines>` mirrors Excel's "Add Chart Element -> Lines -> Series Lines" toggle — connector lines drawn between paired data points across consecutive series in a stacked bar / column chart (visible only on `stacked` / `percentStacked` groupings; Excel disables the toggle on a clustered chart). Up until now hucre's writer had no slot for the element, so a stacked-bar template that pinned the connectors flattened back to no series lines on every parse -> clone -> write loop. This bridges another visualization gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.serLines); // true when the template pinned <c:serLines/>

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [{
      type: "column",
      barGrouping: "stacked",
      series: [
        { name: "Q1", values: "B2:B6", categories: "A2:A6" },
        { name: "Q2", values: "C2:C6", categories: "A2:A6" },
      ],
      serLines: true, // connector lines between same-category data points
    }],
  }],
});

// -- Clone-through --
  type: "column",
  // serLines: undefined -> inherit
  // serLines: null      -> drop inherited flag
  // serLines: true      -> pin
  // serLines: false     -> equivalent to null (writer treats both as default)
});
```

## Model

```ts
// Read side
export interface Chart {
  // ...existing fields...
  serLines?: boolean; // <c:barChart> / <c:ofPieChart>
}

// Write side
export interface SheetChart {
  // ...existing fields...
  serLines?: boolean; // emitted on type === "bar" | "column"
}

// Clone-through
  // ...existing fields...
  serLines?: boolean | null;
}
```

## Scope rules

- The element is bare (no `val` attribute), so the reader surfaces `true` on presence and `undefined` on absence; the writer emits the element only when the field is literally `true` (`false`, absence, and non-boolean inputs all collapse to no element).
- `<c:serLines>` is read off the first `<c:barChart>` / `<c:ofPieChart>` in the plot area. Stray elements on chart families whose schema rejects them (line / area / pie / doughnut / scatter / stock) are ignored.
- The bar writer pins `<c:serLines>` between `<c:overlap>` and the first `<c:axId>` to honour the CT_BarChart sequence (barDir, grouping, varyColors?, ser*, dLbls?, gapWidth?, overlap?, serLines*, axId+). The element pins regardless of grouping — Excel only renders the connectors on stacked / percentStacked but the writer honours the toggle on a clustered chart too (matches Excel's own behavior when the toggle is flipped on a clustered chart).
- The clone-through carries `serLines` across the bar ↔ column coercion (both map to `<c:barChart>` with a different `barDir`) and drops it on every other type-coercion. Non-boolean overrides drop rather than fall through to the inherited value, mirroring the existing `dropLines` / `hiLowLines` / `upDownBars` resolvers.

## Test plan

- [x] `pnpm test` (oxlint + oxfmt + tsgo + vitest) — 3835 tests passing (+39).
- [x] `parseChart` covers presence / absence on `<c:barChart>` (with and without nested `<c:spPr>`), the schema-scope gate (stray elements on line / pie / area drop), `<c:ofPieChart>` (the second OOXML host), combo-workbook surfacing, and a clustered bar chart pinning the element.
- [x] `writeChart` covers absence / explicit-`false` / explicit-`true` emit on column / bar / stacked / percentStacked / clustered, the per-family scope gate (line / area / pie / doughnut / scatter all drop), non-boolean drop, OOXML element ordering (after `<c:overlap>` / `<c:gapWidth>`, before `<c:axId>`), the exactly-once emit guard, `parseChart` round-trip on bar and column, and end-to-end packaging via `writeXlsx`.
- [x] `cloneChart` covers chart-level inherit / null / wholesale-replace, the bar-only default coercion (read-side `bar` flattens to write-side `column` so the flag still threads through), the bar ↔ column coercion carry, the per-family drop on line / area / pie / doughnut / scatter, composition with other chart-level overrides (barGrouping / dataLabels), and end-to-end through `parseChart -> cloneChart -> writeChart` plus packaging via `writeXlsx`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -890,6 +890,16 @@ presence surfaces `true`, absence collapses to `undefined`. The
 element has no slot on `<c:areaChart>` / `<c:area3DChart>` per the
 OOXML schema, so the reader ignores it on those parents and on every
 non-line / non-stock family.
+`Chart.serLines` surfaces the chart-type-level `<c:serLines/>` flag on
+the first `<c:barChart>` / `<c:ofPieChart>` element — Excel's "Add
+Chart Element → Lines → Series Lines" toggle (connector lines drawn
+between paired data points across consecutive series in a stacked bar
+/ column chart). Same bare-element shape as `dropLines` / `hiLowLines`:
+presence surfaces `true`, absence collapses to `undefined`. The reader
+does not surface `serLines` for chart families whose OOXML schema
+rejects the element (line / pie / doughnut / area / scatter / stock /
+radar / surface / bubble) — a stray element on those parents is
+ignored.
 `ChartSeriesInfo.smooth` surfaces the per-series
 `<c:ser><c:smooth val=".."/>` flag — Excel's "Format Data Series →
 Line → Smoothed line" toggle — only on `line` / `line3D` / `scatter`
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
Reviewed the diff end-to-end. The implementation mirrors the existing connector-line resolvers (dropLines / hiLowLines) cleanly across all three layers:

- Reader scopes the lookup to <c:barChart> / <c:ofPieChart> per the OOXML schema, correctly ignoring stray elements on line / pie / area / scatter / stock parents. The bare-element on-presence shape matches dropLines / hiLowLines.
- Writer emits <c:serLines/> in the correct CT_BarChart sequence position (after overlap / gapWidth, before axId+). Honoring the toggle on a clustered chart is the correct call — matches Excel's own behavior where the element pins regardless of grouping but renders nothing when not stacked.
- Clone resolver carries the flag across the bar ↔ column coercion (both routing to <c:barChart>) and drops on every other resolved family. The decision to treat false ≡ null on the override side is consistent with the dropLines / hiLowLines grammar.

Test coverage is thorough — parseChart guards on every other family (line / pie / area / doughnut / scatter), ofPieChart surfacing for downstream introspection even though hucre's writer never authors it, OOXML schema-order placement on both clustered (gapWidth) and stacked (overlap) bar charts, exactly-once emit, and end-to-end through parseChart -> cloneChart -> writeXlsx packaging.

CI is green, no concerns. Merging.

productdevbook (commented):
Re-reviewed after the merge-main conflict resolution. The conflicts were purely additive — both features (serLines on bar/column, showLineMarkers on line) are scoped to disjoint chart families, so the resolution simply concatenates the two sections in chart-reader.ts (parser helper functions) and README.md (cloneChart connector-line documentation block). Local pnpm test (88 files / 3875 tests) and pnpm build pass cleanly with both features in place. Will merge once CI completes on the merged head.

PR #239, opened 2026-05-03
feat(xlsx): chart autoTitleDeleted — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:chart><c:autoTitleDeleted val=".."/>` (Excel's "user explicitly deleted the auto-generated title" flag) at the read, write, and clone layers so a template's auto-title state survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:autoTitleDeleted>` records whether the user explicitly deleted the auto-generated title that single-series charts synthesise from the series name. The element sits on `<c:chart>` directly (between `<c:title>` and `<c:plotArea>` per CT_Chart, ECMA-376 Part 1, §21.2.2.4), not nested inside `<c:title>`, and is independent of whether a literal `<c:title>` is emitted — a chart with no title may still pin `val="1"` to suppress Excel's synthesis or `val="0"` to let Excel synthesise one.

Until now hucre's writer hard-coded the value from the title presence (emitting `val="0"` for titled charts and `val="1"` for titleless ones) and the reader ignored the element entirely, so a templated single-series chart that pinned `val="0"` to keep the auto-title visible silently lost the synthesised title on every parse -> clone -> write loop. The new `autoTitleDeleted` field lets callers override the title-presence-derived default explicitly, and the parser / writer / clone layers preserve the override through the full template loop.

Refs #136

## API

```ts

// ── Read side ──
console.log(source.autoTitleDeleted);
// true       ← when the template pinned <c:autoTitleDeleted val="1"/>
// undefined  ← when the template emits the default <c:autoTitleDeleted val="0"/>
//              (or absence)

// ── Write side — suppress Excel's auto-title on a titleless chart ──
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [...],
    charts: [
      {
        type: "column",
        // no `title` — Excel would normally synthesise one from the
        // series name, but the flag stops it.
        autoTitleDeleted: true,   // emits <c:autoTitleDeleted val="1"/>
      },
    ],
  }],
});

// ── Write side — let Excel synthesise the auto-title (titleless) ──
autoTitleDeleted: false           // emits <c:autoTitleDeleted val="0"/>

// ── Write side — undefined / omitted ──
// (no field set)                  // writer derives from title presence:
//                                //   <c:autoTitleDeleted val="0"/> when
//                                //   `title` is set, val="1" otherwise.

// ── Clone-through ──
  // autoTitleDeleted: undefined   // inherit the parsed value
  // autoTitleDeleted: null        // drop, fall back to title-presence-derived default
  // autoTitleDeleted: true        // pin
  // autoTitleDeleted: false       // pin
});
```

## Behavior

- **Read** — collapses the OOXML default `val="0"` and absence to `undefined` so a fresh chart and a chart that pinned the default round-trip identically. Accepts `"1"` / `"true"` / `"0"` / `"false"`; unknown / missing values drop to `undefined`.
- **Write** — when the caller pins `autoTitleDeleted` explicitly the literal boolean wins. When the field is omitted the writer derives the value from the title presence so back-compat holds: titled charts emit `val="0"`, titleless charts emit `val="1"`. The element is always emitted (Excel's reference contract).
- **Clone** — follows the same `boolean | null | boolean` override grammar as `roundedCorners` / `plotVisOnly` / `titleOverlay`. The override is independent of the resolved title presence so a clone with no literal title can still pin the flag, and a clone with a literal title can override the derived default.

## Test plan

- [x] `pnpm test` (lint + typecheck + 3836 vitest tests) passes.
- [x] `pnpm build` produces the dist bundle.
- [x] New `parseChart — autoTitleDeleted` describe block (10 tests) covers the OOXML truthy / falsy spellings, default collapse, missing `val`, unknown tokens, every chart family, independence from the title element, and co-existence with sibling chart-level toggles.
- [x] New `writeChart — autoTitleDeleted` describe block (14 tests) covers the title-presence-derived defaults, explicit overrides, element ordering inside `<c:chart>`, single-emit guard, every chart family, composition with `titleOverlay`, non-boolean inputs, and a `writeXlsx` package round-trip.
- [x] New `cloneChart — autoTitleDeleted` describe block (12 tests) covers inherit / replace / drop semantics, type flatten through every family, decoupling from the resolved title, composition with `titleOverlay`, and three end-to-end `writeXlsx` re-parse roundtrips.

### DIFF
--- a/README.md
+++ b/README.md
@@ -871,6 +871,18 @@ truthy / falsy spellings (`"1"` / `"true"` / `"0"` / `"false"`);
 unknown values and missing `val` attributes drop to `undefined`. The
 flag is dropped whenever the chart omits the `<c:title>` element
 entirely — there is no overlay slot to surface in that case.
+`Chart.autoTitleDeleted` surfaces the chart-level
+`<c:chart><c:autoTitleDeleted val=".."/>` flag — Excel's record of
+whether the user explicitly deleted the auto-generated title that
+single-series charts synthesise from the series name. The element sits
+on `<c:chart>` directly (between `<c:title>` and `<c:plotArea>` per
+CT_Chart, ECMA-376 Part 1, §21.2.2.4), not nested inside `<c:title>`,
+so a chart with no `<c:title>` may still pin the flag. The OOXML
+default `false` collapses to `undefined` so absence and
+`<c:autoTitleDeleted val="0"/>` round-trip identically; only an
+explicit `val="1"` surfaces `true`. The reader accepts the OOXML
+truthy / falsy spellings (`"1"` / `"true"` / `"0"` / `"false"`);
+unknown values and missing `val` attributes drop to `undefined`.
 `Chart.dropLines` surfaces the chart-type-level `<c:dropLines/>` flag
 on the first `<c:lineChart>` / `<c:line3DChart>` / `<c:areaChart>` /
 `<c:area3DChart>` element — Excel's "Add Chart Element → Lines → Drop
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
Reviewed the changes — surfacing `<c:chart><c:autoTitleDeleted>` at the read / write / clone layers.

**Strengths**
- Symmetric round-trip contract: reader collapses both absence and `val="0"` to `undefined`, writer derives the same default from title presence, and the field round-trips through `parseChart` -> `cloneChart` -> `writeChart` for both the default and pinned states. The asymmetric collapse (only `val="1"` surfaces; `val="0"` collapses to `undefined`) matches the existing project convention used by other chart-level toggles.
- The override resolver (`resolveAutoTitleDeleted` in chart-writer.ts:262) carefully accepts only literal `true` / `false` and falls back to the title-presence-derived default for any other input, preventing stray non-boolean inputs from inverting the derived value.
- Independence from the title block is correctly modeled — the flag carries through clones with `title: null` or `showTitle: false`, and composes with `titleOverlay` on a different parent. Test coverage exercises this independence.
- Element ordering (after `<c:title>`, before `<c:plotArea>`) per CT_Chart §21.2.2.4 is asserted via dedicated tests and the writer always emits exactly one element (no double-emit on override).
- 36 new tests across reader / writer / clone — covering OOXML truthy/falsy spellings, defaults, missing `val`, unknown tokens, every chart family, composition with sibling chart-level toggles, and three end-to-end `writeXlsx` round-trips.

**Fixes during review**
- Resolved end-of-file conflicts in `test/charts.test.ts`, `test/charts-write.test.ts`, and `test/chart-clone.test.ts` against `main` — main's recently-landed `showLineMarkers` describe blocks were interleaved with this PR's `autoTitleDeleted` describe blocks. Both feature blocks are kept side-by-side. Source files (chart-reader / chart-writer / chart-clone) auto-merged cleanly; both features now coexist.
- Re-ran `pnpm test` (lint + typecheck + 3914 vitest tests) and `pnpm build` after the merge — both green.

**No remaining concerns.** Will merge once CI is green (currently pending).

PR #240, opened 2026-05-03
feat(xlsx): chart legend entries — read, write, clone-through

productdevbook:
## Summary

Surfaces the per-series `<c:legendEntry>` overrides at the read, write, and clone layers so a template's hide-by-index legend overrides survive `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:legendEntry>` mirrors Excel's "Format Legend Entries -> Hide" action — each entry pins a 0-based series `<c:idx>` selector and a `<c:delete val=".."/>` flag (the OOXML `CT_LegendEntry` shape, ECMA-376 Part 1, §21.2.2.115). The element sits inside `<c:legend>` between `<c:legendPos>` and `<c:overlay>`, and is the canonical way templates suppress a single series from the legend without removing it from the plot. Up until now hucre's writer had no slot for the element, and the reader ignored it entirely, so a templated dashboard whose helper series was hidden from the legend silently leaked the entry back into the rendered chart on every parse -> clone -> write loop. Bridges another legend-customization gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.legendEntries);
// [{ idx: 1, delete: true }]   ← when the template hid series #1 from the legend
// undefined                    ← when the template emits no <c:legendEntry> children

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [
      ["Region", "Sales", "Forecast"],
      ["North", 100, 120],
      ["South", 200, 180],
    ],
    charts: [{
      type: "column",
      title: "Sales vs Forecast",
      series: [
        { name: "Sales",    values: "B2:B3", categories: "A2:A3" },
        { name: "Forecast", values: "C2:C3", categories: "A2:A3" },
      ],
      // Hide the forecast helper series from the legend without
      // dropping it from the plot.
      legendEntries: [{ idx: 1, delete: true }],
    }],
  }],
});

// -- Clone-through --
//   → inherits the source's parsed legendEntries unchanged.
cloneChart(source, {
  legendEntries: null,
});
//   → drops the inherited list (writer emits no <c:legendEntry> children).
cloneChart(source, {
  legendEntries: [
    { idx: 0, delete: false },
    { idx: 2, delete: true },
  ],
});
//   → replaces the inherited list outright.
```

## Implementation notes

- **Reader**: `parseLegendEntries` walks `<c:legend>`'s `<c:legendEntry>` children, only admitting entries whose `<c:idx val=".."/>` selector parses to a non-negative integer (matches `xsd:unsignedInt`). Missing `<c:delete>` collapses to `delete: false` (the OOXML default). Duplicate `idx` values keep the first occurrence. The lookup is gated on a visible legend — a hidden / missing legend has no slot for entry overrides.
- **Writer**: `resolveLegendEntries` deduplicates with last-wins semantics (so a clone-through that appends an override naturally beats the source's parsed value) and emits in ascending `idx` order to match Excel's reference serialization. The writer always emits both `<c:idx>` and `<c:delete>` children explicitly so a re-parse sees the canonical CT_LegendEntry shape. The block slots between `<c:legendPos>` and `<c:overlay>` per the CT_Legend sequence. The list is silently ignored when `legend: false` suppresses the entire legend element.
- **Clone**: `resolveLegendEntries` follows the standard `undefined` (inherit) / `null` (drop) / `array` (replace) grammar. Empty-array overrides collapse to `undefined` to match the writer's omit-default serialization. Inherited entries are defensively copied so a downstream mutation to the cloned `SheetChart` never leaks back into the parsed `Chart`. The list is dropped from the cloned `SheetChart` whenever the resolved `legend` is `false`, mirroring the `legendOverlay` scoping rule.
- **Validation**: Entries whose `idx` is non-finite, negative, or non-integer drop at write time rather than emit a token Excel would reject. Non-boolean `delete` values collapse to `false`.

## Test plan

- [x] `parseChart — legend entries` (12 tests) — single / multiple entries, missing `<c:delete>` defaults, OOXML truthy/falsy spellings, hidden legend, malformed `<c:idx>`, deduplication.
- [x] `writeChart — legend entries` (15 tests) — empty/unset omission, single entry shape, ascending idx ordering, explicit `delete: false` round-trip, CT_Legend ordering, `legend: false` suppression, idx validation, dedup, every chart family, full `writeXlsx` package round-trip.
- [x] `cloneChart — legendEntries` (16 tests) — inherit / replace / drop / empty-array semantics, defensive copy, every flatten, hidden-legend scoping, end-to-end parse -> clone -> write -> reparse.
- [x] `pnpm lint && pnpm typecheck && pnpm vitest run --dir=test` — all 3917 tests pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -860,6 +860,18 @@ missing `val` attributes drop to `undefined`. The flag is dropped
 whenever `Chart.legend` is `false` or the chart omits the legend
 element entirely — there is no overlay slot on a hidden legend, so the
 parsed shape stays minimal.
+`Chart.legendEntries` surfaces the per-series legend-entry overrides
+pulled from `<c:legend><c:legendEntry>` children — Excel's "Format
+Legend Entries → Hide" action surfaces the same elements. Each entry
+exposes the 0-based series `idx` selector and the `delete` flag (the
+hide-bit pulled from `<c:delete val=".."/>`); per-entry text styling
+(`<c:txPr>`) is intentionally not modelled at this layer. Entries
+without a valid `<c:idx>` selector drop on parse rather than surface a
+fabricated index, and duplicate `<c:idx>` values keep only the first
+occurrence so a re-emit stays canonical. Absence collapses to
+`undefined` so a chart with no overrides round-trips minimally — the
+field is dropped whenever `Chart.legend` is `false` or the chart omits
+the legend element entirely.
 `Chart.titleOverlay` surfaces the title-overlay flag pulled from
 `<c:title><c:overlay val=".."/></c:title>` — Excel's "Format Chart
 Title → Show the title without overlapping the chart" toggle (the
@@ -1169,6 +1181,21 @@ on top of the plot area so the chart series get the full frame
 (`val="1"`). The flag is silently ignored when `legend: false`
 suppresses the entire legend element — there is no overlay slot on a
 hidden legend, so the writer skips emitting any orphaned `<c:overlay>`.
+The chart-level `legendEntries` field maps to `<c:legendEntry>` children
+inside `<c:legend>` — Excel's "Format Legend Entries → Hide" action
+surfaces the same elements. Each entry pins a 0-based series `idx` and
+a `delete` flag (defaults to `false` — the entry renders); the writer
+emits both `<c:idx>` and `<c:delete>` children explicitly so a re-parse
+sees the canonical CT_LegendEntry shape. Entries whose `idx` is
+non-finite, negative, or non-integer drop at write time rather than emit
+a token Excel rejects, and duplicate `idx` values are deduplicated with
+last-wins semantics so a clone-through that appends an override naturally
+beats the source's parsed value. The writer slots `<c:legendEntry>`
+between `<c:legendPos>` and `<c:overlay>` (CT_Legend sequence) and
+emits entries in ascending `idx` order so the file matches Excel's
+reference serialization. The list is silently ignored when `legend:
+false` suppresses the entire legend element — there is no slot for
+entries on a hidden legend, so the writer skips emission entirely.
 The chart-level `titleOverlay` field maps to `<c:overlay val=".."/>`
 inside `<c:title>` — Excel's "Format Chart Title → Show the title
 without overlapping the chart" toggle (the checkbox is the inverse of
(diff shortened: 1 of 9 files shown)


productdevbook (commented):
## Review summary

Reviewed the chart legend-entries feature end-to-end (reader / writer / clone) and the supporting tests. The implementation is clean and follows established conventions used by neighbouring legend-scoped fields (`legendOverlay`, `titleOverlay`, etc.).

### What I checked
- **Types** (`src/_types.ts`): `ChartLegendEntry` is well-documented, exported from the public surface (`src/index.ts`). `Chart.legendEntries` and `SheetChart.legendEntries` both carry thorough JSDoc explaining the absence-vs-empty-array semantics.
- **Reader** (`parseLegendEntries`): correctly gated on a visible legend, validates `<c:idx>` against `xsd:unsignedInt` semantics (non-finite / negative / non-integer drop), accepts the OOXML truthy/falsy `<c:delete>` spellings via `readBoolVal`, and dedupes with first-wins semantics so a re-emit stays canonical.
- **Writer** (`resolveLegendEntries` + `buildLegend`): dedupes with last-wins (so a clone-through that appends an override naturally beats the inherited value), emits in ascending `idx` order matching Excel's reference serialization, slots `<c:legendEntry>` between `<c:legendPos>` and `<c:overlay>` per CT_Legend, and always emits both `<c:idx>` and `<c:delete>` children explicitly so a re-parse sees the canonical shape.
- **Clone** (`resolveLegendEntries`): standard `undefined` (inherit) / `null` (drop) / `array` (replace) grammar; defensive copy of both the array and each entry object so downstream mutation cannot leak back into the parsed `Chart`; correctly scoped to drop the field when the resolved legend is `false`.
- **Tests**: 43 new tests across `test/charts.test.ts` (parser), `test/charts-write.test.ts` (writer), and `test/chart-clone.test.ts` (clone-through). Coverage includes single/multi entries, missing `<c:delete>` defaults, OOXML truthy/falsy spellings, hidden-legend scoping, malformed `<c:idx>`, dedup, ordering, every chart family, and a full `writeXlsx` round-trip.
- **README**: three new sections cover the parser, writer, and clone-through semantics consistently with neighbouring features.

### What I fixed
The PR was `CONFLICTING` against `main` because PR #239 (autoTitleDeleted) merged the same end-of-file regions in `test/chart-clone.test.ts` and `test/charts-write.test.ts`. I merged `main` in and resolved the append-conflicts by keeping both feature describe blocks side by side. Auto-merges in `README.md`, `src/_types.ts`, `src/xlsx/chart-clone.ts`, `src/xlsx/chart-reader.ts`, and `src/xlsx/chart-writer.ts` were correct (the two features touch disjoint code paths).

Locally verified after the merge:
- `pnpm test` (lint + typecheck + vitest) — **3956 / 3956 pass** (43 new + the existing autoTitleDeleted suite + the rest).
- `pnpm build` — obuild succeeds.

### Remaining concerns
None blocking. CI is currently running on the merge commit; will merge once the green tick lands.

PR #241, opened 2026-05-03
feat(xlsx): chart 3-D view configuration — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:chart><c:view3D>` (CT_View3D, ECMA-376 Part 1, §21.2.2.228) at the read, write, and clone layers so a template's 3-D rotation / perspective shape survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:view3D>` mirrors Excel's "3-D Rotation" pane on 3D chart families (`bar3D`, `line3D`, `pie3D`, `area3D`, `surface3D`). The element holds six independently optional CT_View3D children — `<c:rotX>` (X rotation, signed -90..90), `<c:hPercent>` (height percent, 5..500), `<c:rotY>` (Y rotation, 0..360), `<c:depthPercent>` (depth percent, 20..2000), `<c:rAngAx>` (right-angle-axes flag), and `<c:perspective>` (foreshortening factor, 0..240). Until now hucre's writer had no slot for the element, and the reader ignored it entirely, so a templated 3D dashboard chart silently lost its rotation and perspective settings on every parse -> clone -> write loop. Bridges another rotation-customization gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.view3D);
// { rotX: 15, hPercent: 100, rotY: 20, depthPercent: 100, rAngAx: true, perspective: 30 }
//   ← when the template authored the bar3D chart with Excel's defaults
// undefined
//   ← when the template emits no `<c:view3D>` element

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [
      ["Region", "Sales"],
      ["North", 100],
      ["South", 200],
    ],
    charts: [{
      type: "column",
      title: "Sales (3D look)",
      series: [{ name: "Sales", values: "B2:B3", categories: "A2:A3" }],
      // Pin a custom rotation / perspective. Each numeric field is
      // clamped against its OOXML simple-type range; out-of-range and
      // fractional inputs drop silently.
      view3D: { rotX: 20, rotY: 30, perspective: 45 },
    }],
  }],
});

// -- Clone-through --
//   → inherits the source's parsed view3D unchanged (defensively copied).
cloneChart(source, {
  view3D: null,
});
//   → drops the inherited block (writer skips `<c:view3D>` entirely).
cloneChart(source, {
  view3D: { rotX: 45 },
});
//   → replaces the inherited block wholesale (no per-field merge).
cloneChart(source, {
  view3D: {},
});
//   → empty object collapses to a bare `<c:view3D/>` shell on emit.
```

## Implementation notes

- **Reader**: `parseView3D` walks `<c:view3D>` and surfaces only the children whose `val` parses inside the matching OOXML simple-type range. Each numeric field is validated with a strict integer regex (rejects `15.5`, `15px`); out-of-range values drop rather than fabricate a clamped result. Boolean `<c:rAngAx>` accepts the OOXML truthy / falsy spellings; unknown tokens drop. The element itself is the gating signal — a `<c:view3D>` with no resolvable children surfaces as `{}`, mirroring how `dataTable` / `protection` handle a malformed inner block. The lookup runs on every chart family (the schema accepts the element on every CT_Chart) so a stray element on a 2D chart still surfaces for lossless round-trip.
- **Writer**: `buildView3D` returns `undefined` when the field is unset (writer skips emission), a bare `<c:view3D/>` shell when the input is `{}`, or the populated element with one self-closing child per pinned field. Children emit in CT_View3D schema order (`rotX`, `hPercent`, `rotY`, `depthPercent`, `rAngAx`, `perspective`). Each numeric child is clamped via `clampView3DInt` against its OOXML range; out-of-range, fractional, and non-finite inputs drop silently. The boolean child emits `val="0"` literally on `false` for round-trip parity with templates that author the explicit OOXML default. The block slots between `<c:autoTitleDeleted>` and `<c:plotArea>` per the CT_Chart sequence.
- **Clone**: `resolveView3D` follows the standard `undefined` (inherit) / `null` (drop) / `object` (replace) grammar. Inherited values are defensively shallow-copied so a downstream mutation to the cloned `SheetChart` never leaks back into the parsed `Chart`. Unlike `dataTable`, the field has no axis dependency — `<c:view3D>` lives on `<c:chart>`, so the resolver applies to every chart family (pie / doughnut included). The grammar mirrors `protection` / `dataTable` so the chart-level block toggles compose the same way at the call site.
- **Validation**: Numeric fields outside the matching OOXML range (`rotX` -90..90, `hPercent` 5..500, `rotY` 0..360, `depthPercent` 20..2000, `perspective` 0..240) drop at write time rather than emit a token Excel's strict validator would reject. Fractional and non-finite inputs (`NaN`, `Infinity`) drop at the same boundary.

## Test plan

- [x] `parseChart — view3D` (14 tests) — fully-pinned shape, partial shape, signed rotX, boundary values, out-of-range drop, fractional drop, OOXML truthy / falsy spellings, missing val drop, unknown rAngAx tokens, malformed children gating, bare element gating, absence, every chart family scope, sibling toggle co-existence.
- [x] `writeChart — view3D` (16 tests) — unset omission, empty-object bare shell, fully-pinned children, signed rotX, boundary values, out-of-range drop, fractional drop, non-finite drop, explicit `false` rAngAx, non-boolean rAngAx drop, CT_View3D schema order, CT_Chart placement, single-emit guard, every chart family scope, full round-trip via parse / write, full `writeXlsx` package round-trip.
- [x] `cloneChart — view3D` (12 tests) — inherit / replace / drop / empty-object semantics, defensive copy, every flatten (line → column / doughnut / pie), end-to-end parse -> clone -> write -> reparse for fully-pinned, empty-shell, and dropped overrides.
- [x] `pnpm test` — all 3959 tests pass (lint + typecheck + vitest).

### DIFF
--- a/README.md
+++ b/README.md
@@ -735,6 +735,27 @@ surfaces as an empty `{}`, and absence of the element collapses the
 field to `undefined`. Every chart family carries a slot — the element
 lives on `<c:chartSpace>` (a sibling of `<c:chart>`), not inside
 `<c:plotArea>`, so pie / doughnut still surface the block.
+`Chart.view3D` surfaces the chart-level `<c:chart><c:view3D>` block
+(CT_View3D, ECMA-376 Part 1, §21.2.2.228) — Excel's "3-D Rotation"
+pane on 3D chart families. The reader returns a `ChartView3D` object
+with up to six independently optional fields (`rotX`, `hPercent`,
+`rotY`, `depthPercent`, `rAngAx`, `perspective`); each child is
+independently optional on `CT_View3D`, so the reader only surfaces
+the fields the file actually pinned. Numeric fields are validated
+against the matching OOXML simple-type range
+(`ST_RotX` / `ST_HPercent` / `ST_RotY` / `ST_DepthPercent` /
+`ST_Perspective`) — out-of-range, fractional, and non-numeric `val`
+attributes drop rather than fabricate a clamped value. The boolean
+`<c:rAngAx>` accepts the OOXML truthy / falsy spellings (`"1"` /
+`"true"` / `"0"` / `"false"`); unknown tokens drop to `undefined`.
+The element itself is the gating signal — a `<c:view3D>` block with
+no resolvable children surfaces as an empty `{}`, and absence of the
+element collapses the field to `undefined`. The element lives on
+`<c:chart>` (between `<c:autoTitleDeleted>` and `<c:plotArea>`), so
+the OOXML schema accepts it on every chart family — though it is
+only meaningful on 3D families (`bar3D`, `line3D`, `pie3D`, `area3D`,
+`surface3D`); a stray element on a 2D chart still surfaces here so
+the round-trip through `cloneChart` stays lossless.
 `ChartAxisInfo.tickLblSkip` and `ChartAxisInfo.tickMarkSkip` surface
 the category-axis tick-thinning knobs (`<c:catAx><c:tickLblSkip val=".."/>`
 and `<c:catAx><c:tickMarkSkip val=".."/>`). Both elements live on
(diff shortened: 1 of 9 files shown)


productdevbook (commented):
Reviewed the diff and ran the full suite locally after merging main into the branch.

**Resolved**

- Conflict in `test/charts.test.ts` where main added a `parseChart — legend entries` describe block at the same trailing location where this branch added `parseChart — view3D`. Kept both suites by closing the view3D describe before the legend entries describe begins. Pushed as a merge commit.
- `pnpm test` (lint + typecheck + 4001 vitest cases) and `pnpm build` both pass on the merged tree.

**Review notes**

- `parseView3D` and `buildView3D` mirror each other on the reader/writer boundary cleanly: the reader's strict `^-?\d+$` regex rejects exactly the same shape (`15.5`, `15px`, `NaN`) that the writer's `Number.isInteger` + `Number.isFinite` guard rejects, so the round-trip is bidirectionally lossless.
- The CT_View3D child emission order (`rotX`, `hPercent`, `rotY`, `depthPercent`, `rAngAx`, `perspective`) matches ECMA-376 §21.2.2.228 — verified by the schema-order test.
- Element placement between `<c:autoTitleDeleted>` and `<c:plotArea>` is consistent with `CT_Chart` §21.2.2.4.
- `resolveView3D` follows the established `undefined` (inherit, with defensive shallow copy) / `null` (drop) / `object` (replace wholesale) grammar used by `protection` / `dataTable`, so it composes the same way at the call site.
- Per-family scope: the resolver applies to every chart kind including pie/doughnut, which is correct since `<c:view3D>` lives on `<c:chart>` directly (no axis dependency, unlike `dataTable`).
- Edge cases that are well-covered: bare `<c:view3D/>` shell surfacing as `{}`, fractional drop, signed `rotX`, OOXML `0`/`1`/`true`/`false` boolean spellings, malformed/missing `val` drop, every CT_View3D range boundary, double-emit guard, full parse → clone → write → reparse loops.
- README and TSDoc updates accurately describe the read/write/clone semantics and the per-family scope nuance (only meaningful on 3D families but accepted on every CT_Chart).

**Concerns**

None blocking. Approving and queuing for merge once CI is green on the merge commit.

PR #242, opened 2026-05-03
feat(xlsx): chart axis custom display unit divisor — read, write, clone-through

productdevbook:
## Summary

Surfaces the `<c:dispUnits><c:custUnit val=".."/></c:dispUnits>` variant on chart value axes at the read, write, and clone layers so a template authored with Excel's "Format Axis -> Display units -> Other" path survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch alongside the existing `<c:builtInUnit>` preset.

`<c:custUnit>` is the second arm of the OOXML `CT_DispUnits` `xsd:choice` (ECMA-376 Part 1, §21.2.2.32) — the alternative to the named `<c:builtInUnit>` preset. It carries an arbitrary `CT_Double` divisor that rescales every numeric tick label (e.g. `86400` to convert seconds to days; `1500` to scale by an arbitrary factor the named presets cannot express). Until now hucre's reader and writer only modelled the built-in path; the `custUnit` variant was explicitly noted as out-of-scope. A templated chart pinning a custom divisor therefore silently lost the value on every `parse -> clone -> write` loop. Bridges another axis-customization gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.axes?.y?.dispUnits);
// { custUnit: 86400 }                            ← when the template emits <c:custUnit val="86400"/>
// { custUnit: 500, showLabel: true }             ← with the automatic unit annotation
// { unit: "millions" }                           ← unchanged for the existing built-in path

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [
      ["Hour", "Seconds"],
      ["H1", 360_000],
      ["H2", 720_000],
    ],
    charts: [{
      type: "column",
      title: "Elapsed time",
      series: [{ name: "Seconds", values: "B2:B3", categories: "A2:A3" }],
      // 86_400 seconds per day — convert tick labels to days.
      axes: { y: { dispUnits: { custUnit: 86_400, showLabel: true } } },
    }],
  }],
});

// -- Clone-through --
//   → inherits the source's parsed dispUnits unchanged.
cloneChart(source, {
  axes: { y: { dispUnits: { custUnit: 1500 } } },
});
//   → replaces the inherited preset with a custom divisor.
cloneChart(source, {
  // Layer a custom divisor on top of an inherited preset; the writer
  // emits the more specific <c:custUnit> per the OOXML xsd:choice.
  axes: { y: { dispUnits: { unit: "millions", custUnit: 1500 } } },
});
```

## Implementation notes

- **Type model**: `ChartAxisDispUnits.unit` becomes optional and a new optional `custUnit?: number` joins it. The two are mutually exclusive in the OOXML schema's `xsd:choice`; the writer enforces the choice on emit by preferring `custUnit` when both are pinned. Keeping both fields on the cloned object lets a caller append a custom divisor to a parsed source without manually pruning the inherited preset.
- **Reader** (`parseAxisDispUnits`): walks `<c:dispUnits>` and surfaces both children in one pass — `<c:custUnit>` first (it wins on a malformed template that declares both), falling back to `<c:builtInUnit>` if the cust value is malformed. The numeric `<c:custUnit val>` parses via `Number.parseFloat` and only surfaces when the result is finite and strictly positive (`0`, negative, `NaN`, `Infinity` drop). A `<c:dispUnits>` block that resolves neither child surfaces as `undefined` (Excel's strict validator rejects an empty shell).
- **Writer** (`buildAxisDispUnits` + `normalizeAxisDispUnits`): the normalizer keeps both fields when both are valid; the builder emits exactly one element per the OOXML `xsd:choice`, preferring `<c:custUnit>` when both are pinned. A bare `ChartAxisDispUnits` object with neither field valid drops the entire `<c:dispUnits>` element rather than ship a shell Excel rejects. Same value-axis-only scope rule as the existing built-in path — `<c:dispUnits>` only emits on `<c:valAx>` (Y axis on bar / column / line / area; both axes on scatter); category axes and pie / doughnut silently drop the field.
- **Clone** (`normalizeDispUnits` + `applyDispUnitsOverride`): same `undefined` (inherit) / `null` (drop) / value (replace) grammar the existing fields use. The normalizer keeps both fields when valid so a clone-through that layers a custom divisor on top of an inherited preset survives without manual pruning. Invalid `custUnit` values drop at the clone boundary so the writer never sees a token Excel would reject.
- **Validation**: `custUnit` must be a finite positive number; `0`, negative, `NaN`, `Infinity`, and non-number inputs drop silently rather than emit a token Excel rejects. Non-string `unit` tokens and tokens outside the OOXML `ST_BuiltInUnit` enum drop at the same boundary (existing behavior). When both fields drop, the entire `<c:dispUnits>` element is omitted.

## Test plan

- [x] `parseChart — axis dispUnits` (8 new tests) — custUnit surfaces, fractional values per CT_Double, showLabel alongside custUnit, non-positive / non-finite drop, custUnit wins when both pinned, fallback to builtInUnit when cust malformed, catAx scope guard, plus an updated drop-when-malformed case.
- [x] `writeChart — axis dispUnits` (10 new tests) — custUnit emit shape, fractional emit, dispUnitsLbl alongside custUnit, custUnit wins when both pinned, fallback to builtInUnit on invalid custUnit, full element drop when neither field resolves, catAx scope guard, both scatter axes, parse round-trip, full `writeXlsx` package round-trip.
- [x] `cloneChart — axis dispUnits` (8 new tests) — inherit / replace / mixed-pin / drop semantics, invalid custUnit drop at clone boundary, showLabel pass-through, pie flatten guard, end-to-end parse -> clone -> write -> reparse.
- [x] `pnpm test` — all 3981 tests pass (lint + typecheck + vitest).

### DIFF
--- a/README.md
+++ b/README.md
@@ -803,25 +803,32 @@ distinct from the `"autoZero"` default which defers to Excel's
 auto-placement). Unknown semantic tokens and non-numeric `val`
 attributes drop rather than fabricate values the writer would reject.
 `ChartAxisInfo.dispUnits` surfaces the per-value-axis display-unit
-preset pulled from `<c:valAx><c:dispUnits><c:builtInUnit val=".."/></c:dispUnits></c:valAx>`
+configuration pulled from `<c:valAx><c:dispUnits>...</c:dispUnits></c:valAx>`
 — Excel's "Format Axis -> Display units" dropdown. The element rescales
-the numeric tick labels by the chosen preset (e.g. `"millions"` divides
+the numeric tick labels by the chosen divisor (e.g. `"millions"` divides
 every label by 1e6) so a chart whose source range stores raw amounts
 can show compact tick labels without modifying the underlying cells.
 The OOXML schema places the element exclusively on `CT_ValAx`, so
 `dispUnits` only surfaces on the value-axis side of bar / column / line
 / area charts (the Y axis) and on both axes of scatter charts (both are
 value axes); category axes (`<c:catAx>`) and pie / doughnut never
-surface the field. The reader keeps the parsed `unit` token and the
-presence of `<c:dispUnitsLbl>` (`showLabel: true` when Excel paints its
-automatic annotation alongside the axis); the alternative
-`<c:custUnit val=".."/>` (custom numeric divisor) and any rich-text
-`<c:dispUnitsLbl>` body are intentionally not surfaced. The OOXML
+surface the field. The reader keeps both choices the OOXML schema
+allows: the named `unit` token from `<c:builtInUnit val=".."/>` and the
+arbitrary numeric divisor from `<c:custUnit val=".."/>` (Excel's
+"Display units → Other" path — e.g. `86400` to convert seconds to
+days). The OOXML schema places the two children in an `xsd:choice`
+(exactly one may appear); on emit hucre prefers `custUnit` when both
+fields are pinned so a clone can layer a custom divisor on top of an
+inherited preset without manually pruning the original. The
+`showLabel: true` flag tracks the presence of `<c:dispUnitsLbl>`
+(Excel paints its automatic annotation alongside the axis); the rich-
+text `<c:dispUnitsLbl>` body is intentionally not surfaced. The OOXML
 schema accepts the nine `ST_BuiltInUnit` tokens (`"hundreds"`,
 `"thousands"`, `"tenThousands"`, `"hundredThousands"`, `"millions"`,
 `"tenMillions"`, `"hundredMillions"`, `"billions"`, `"trillions"`);
 unknown tokens drop to `undefined` rather than fabricate a value the
-writer would never emit.
+writer would never emit. Non-positive / non-finite `custUnit` values
+drop at the same boundary.
 `ChartAxisInfo.crossBetween` surfaces the per-value-axis cross-between
 mode pulled from `<c:valAx><c:crossBetween val=".."/></c:valAx>` — the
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
Reviewed the diff. The implementation is symmetrical across reader, writer, and clone layers, the OOXML xsd:choice precedence (custUnit wins when both are pinned) is consistently enforced on both ends so round-trips are stable, and the validation gate (finite, strictly positive) correctly rejects 0/negative/NaN/Infinity at every boundary. Test coverage is thorough — 26 new tests exercise the read, write, and clone paths plus the parse → clone → write → reparse round-trip and a full writeXlsx packaging.

Notes I checked but did not gate on:
- `ChartAxisDispUnits.unit` widening from required to optional is a minor type change; acceptable on the 0.x line and necessary for the xsd:choice modeling.
- `String(custUnit)` via `xmlSelfClose` will switch to scientific notation for divisors >= 1e21; xs:double permits this and the round-trip test confirms `Number.parseFloat` recovers the same value, so it is fine in practice.
- `Number.parseFloat` is lenient about trailing junk on the `val` attribute, but that mirrors typical OOXML reader behavior and is consistent with the existing reader style.
- README updates land alongside the code; web docs do not reference dispUnits today so no other doc changes are needed.

CI is green and the change reads cleanly. Merging via squash.

PR #243, opened 2026-05-03
feat(xlsx): chart category-axis auto type detection — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:catAx><c:auto val=\"..\"/>` (CT_CatAx, ECMA-376 Part 1, §21.2.2.7) at the read, write, and clone layers so a template's \"Text axis\" pin survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:auto>` mirrors Excel's \"Format Axis -> Axis Options -> Axis Type\" radio. The OOXML default `true` lets Excel inspect the axis labels and decide at render time whether to treat the axis as a discrete category axis or a chronological date axis. `false` pins the axis as a literal category axis — Excel keeps every label as-is regardless of whether the cells parse as dates / numerics. Until now hucre's writer hardcoded `<c:auto val=\"1\"/>` and the reader ignored the element entirely, so a templated chart whose user pinned the \"Text axis\" radio silently lost the choice on every parse -> clone -> write loop. Bridges another category-axis customization gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.axes?.x?.auto);
// false                                ← when the template pinned <c:auto val=\"0\"/>
// undefined                            ← when the template pinned <c:auto val=\"1\"/> or omits the element

// -- Write side --
await writeXlsx({
  sheets: [{
    name: \"Dashboard\",
    rows: [
      [\"Period\", \"Sales\"],
      [\"Q1 2024\", 100],
      [\"Q2 2024\", 200],
    ],
    charts: [{
      type: \"column\",
      title: \"Quarterly sales\",
      series: [{ name: \"Sales\", values: \"B2:B3\", categories: \"A2:A3\" }],
      // Pin the axis as a literal category axis so Excel does not
      // collapse the period strings into a chronological date axis.
      axes: { x: { auto: false } },
    }],
  }],
});

// -- Clone-through --
//   → inherits the source's parsed `auto: false` unchanged.
cloneChart(source, {
  axes: { x: { auto: null } },
});
//   → drops the inherited flag (the writer falls back to the OOXML default `true`).
```

## Implementation notes

- **Type model**: `SheetChart.axes.x.auto?: boolean` (writer side, X-axis only) and `ChartAxisInfo.auto?: boolean` (reader side). The clone option `axes.x.auto?: boolean | null` follows the standard `undefined` (inherit) / `null` (drop) / boolean (replace) grammar.
- **Reader** (`parseAxisAuto`): walks `<c:catAx>` for `<c:auto val=\"..\"/>`. Only `val=\"0\"` / `val=\"false\"` surfaces `false`; the OOXML default `val=\"1\"` (and `val=\"true\"`, missing `val`, unknown tokens, absence) collapses to `undefined`. The element lives on `CT_CatAx` exclusively per the OOXML schema — `<c:dateAx>`, `<c:valAx>`, and `<c:serAx>` reject it — so the parser ignores it on every other axis flavour, mirroring the writer's catAx-only emit path.
- **Writer**: previously hardcoded `<c:auto val=\"1\"/>` in the catAx builder. Now resolves through `xAuto: chart.axes?.x?.auto !== false` so only an explicit `axes.x.auto === false` flips the toggle off; absence (and any non-boolean) falls back to the OOXML default `true`, matching Excel's reference serialization. The element slots between `<c:crosses>` and `<c:lblAlgn>` per the CT_CatAx schema sequence. The flag threads through bar / column / line / area but is silently dropped on scatter (both axes are value axes) and pie / doughnut (no axes at all).
- **Clone** (`applyAutoOverride`): inverse of `applyNoMultiLvlLblOverride` — `<c:auto>` defaults to `true` in the OOXML schema, so the helper collapses `true` rather than `false` (symmetric with the parser-side default-collapse). The catAx-only scope is enforced in the resolver — scatter / pie / doughnut clones drop the inherited flag silently so the cloned `SheetChart` accurately reflects what the chart will paint.
- **Validation**: only literal `boolean` survives the type guard; non-boolean inputs collapse to the default. This matches how `noMultiLvlLbl` / `legendOverlay` / `roundedCorners` / axis `hidden` treat their inputs at the write layer.

## Test plan

- [x] `parseChart — axis auto` (10 new tests) — surfaces `false` on `val=\"0\"`, collapses `val=\"1\"` and absence to `undefined`, OOXML truthy / falsy spellings, missing `val`, unknown tokens, valAx scope guard, scatter scope guard, co-surfaces alongside other catAx fields.
- [x] `writeChart — axis auto` (12 new tests) — emits `val=\"1\"` by default, emits `val=\"0\"` on override, emits default on `auto: true` and non-boolean, exactly-one-element guard, threads through bar / column / line / area, scatter scope guard, pie / doughnut scope guard, valAx scope guard, OOXML schema position, parse round-trip, full `writeXlsx` package round-trip.
- [x] `cloneChart — axis auto` (12 new tests) — inherit / replace / null drop / `true` drop / non-boolean drop semantics, source-without-flag pickup, pie / doughnut / scatter scope guards, chart-type coercion, composition with other axis overrides, end-to-end parse -> clone -> write -> reparse, full `writeXlsx` cloned-chart round-trip.
- [x] `pnpm test` — all 4037 tests pass (lint + typecheck + vitest).

### DIFF
--- a/README.md
+++ b/README.md
@@ -801,6 +801,20 @@ ignores the element on every other axis flavour. Only an explicit
 unknown tokens all collapse to `undefined` so absence and the default
 round-trip identically. The reader accepts the OOXML truthy / falsy
 spellings (`"1"` / `"true"` / `"0"` / `"false"`).
+`ChartAxisInfo.auto` surfaces the per-axis `<c:auto val=".."/>` flag
+— Excel's "Format Axis -> Axis Options -> Axis Type" radio (the
+"Text axis" choice flips the value to `0`). The OOXML default `true`
+lets Excel inspect the axis labels and decide at render time whether
+to treat the axis as a discrete category axis or a chronological
+date axis; `false` pins the axis as a literal category axis so a
+date-shaped category range still renders as a flat category axis
+without any chronological grouping. The OOXML schema places the
+element on `CT_CatAx` exclusively, so the parser ignores it on every
+other axis flavour. Only an explicit `val="0"` surfaces `false`; the
+OOXML default `val="1"`, absence, missing `val`, and unknown tokens
+all collapse to `undefined` so absence and the default round-trip
+identically. The reader accepts the OOXML truthy / falsy spellings
+(`"1"` / `"true"` / `"0"` / `"false"`).
 `ChartAxisInfo.reverse` surfaces the per-axis
 `<c:scaling><c:orientation val="maxMin"/></c:scaling>` flag — Excel's
 "Categories / Values in reverse order" toggle. Only `"maxMin"` surfaces
@@ -1444,6 +1458,22 @@ only an explicit `true` emits `val="1"`. The element lives on
 `<c:serAx>` reject it), so the flag threads through bar / column /
 line / area but is silently dropped on scatter (both axes are value
 axes) and pie / doughnut (no axes at all).
+The `axes.x.auto` flag maps to `<c:catAx><c:auto val=".."/>` — Excel's
+"Format Axis -> Axis Options -> Axis Type" radio (the "Text axis"
+choice flips the value to `0`). Excel's default `true` lets the
+renderer inspect the axis labels and decide at render time whether to
+treat the axis as a discrete category axis or a chronological date
+axis; `false` pins the axis as a literal category axis so a
+date-shaped category range still renders as a flat category axis
+without any chronological grouping. The writer always emits the
+element because Excel's reference serialization includes
+`<c:auto val="1"/>` on every category axis; `true`, absence, and
+non-boolean inputs all collapse to the default `val="1"`, while only
+an explicit `false` emits `val="0"`. The element lives on `CT_CatAx`
+exclusively (even `<c:dateAx>`, `<c:valAx>`, and `<c:serAx>` reject
+it), so the flag threads through bar / column / line / area but is
+silently dropped on scatter (both axes are value axes) and pie /
+doughnut (no axes at all).
 The `axes.x.reverse` and `axes.y.reverse` flags map to
 `<c:scaling><c:orientation val="maxMin"/></c:scaling>` — Excel's
 "Categories / Values in reverse order" toggle. On a category axis,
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
Reviewed end-to-end. The change cleanly bridges the read/write/clone gap for `<c:catAx><c:auto val=".."/>` (CT_CatAx, ECMA-376 Part 1, §21.2.2.7), which previously round-tripped lossily because the writer hardcoded `val="1"` and the reader ignored the element entirely.

Highlights:
- **Symmetric default-collapse**: `parseAxisAuto` collapses both `val="1"` and absence to `undefined`, while `applyAutoOverride` is the inverse of `applyNoMultiLvlLblOverride` (collapses `true` rather than `false`) — matches Excel's reference behavior where the schema default is `true`.
- **Scope guard**: catAx-only enforcement is consistent across reader (`axis.local === "catAx"`), writer (gated in the catAx builder), and clone resolver (`isCatAxisX` check before `applyAutoOverride`). Scatter / pie / doughnut all drop the flag silently.
- **OOXML schema position**: `<c:auto>` slots between `<c:crosses>` and `<c:lblAlgn>` per CT_CatAx, and the test pins this ordering explicitly.
- **Test coverage**: 34 new tests (10 reader, 12 writer, 12 clone) covering OOXML truthy/falsy spellings, missing `val`, unknown tokens, value-axis scope guard, scatter/pie/doughnut scope guards, chart-type coercion, schema position, and full `writeXlsx` round-trip.
- **API consistency**: `axes.x.auto?: boolean | null` follows the same `undefined` (inherit) / `null` (drop) / boolean (replace) grammar as the surrounding axis overrides.
- **Docs**: README updates for both reader (`ChartAxisInfo.auto`) and writer (`axes.x.auto`) sides.

CI green. Merging.

PR #244, opened 2026-05-04
feat(xlsx): chart data labels showLeaderLines flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:showLeaderLines>` (CT_DLbls, ECMA-376 Part 1, §21.2.2.196) at the read, write, and clone layers so a templated pie / doughnut chart's leader-lines pin survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<c:showLeaderLines>` mirrors Excel's "Format Data Labels -> Label Options -> Show Leader Lines" checkbox on pie / doughnut data labels. The OOXML default is `true` (Excel paints the connecting lines on every label that gets pushed outside its slice); pinning `false` opts out and renders labels without the connectors. Until now hucre's writer had no slot for the element and the reader ignored it entirely, so a templated pie/doughnut dashboard chart whose user unchecked "Show Leader Lines" silently lost the choice on every parse -> clone -> write loop. Bridges another data-labels customization gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.dataLabels?.showLeaderLines);
// false      ← when the template pinned <c:showLeaderLines val="0"/>
// undefined  ← when the template pinned <c:showLeaderLines val="1"/> or omits the element

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [
      ["Slice", "Value"],
      ["A", 30],
      ["B", 70],
    ],
    charts: [{
      type: "pie",
      title: "Distribution",
      series: [{ name: "Distribution", values: "B2:B3", categories: "A2:A3" }],
      // Suppress the leader lines — labels render without the
      // connecting lines back to their slice.
      dataLabels: { showValue: true, showLeaderLines: false },
    }],
  }],
});

// -- Clone-through --
//   → inherits the source's parsed `showLeaderLines: false` unchanged.
cloneChart(source, {
  dataLabels: null,
});
//   → drops the inherited block (no `<c:dLbls>` emitted).
```

## Implementation notes

- **Type model**: `ChartDataLabels.showLeaderLines?: boolean` (writer side) and `ChartDataLabelsInfo.showLeaderLines?: boolean` (reader side). Same shape as the other `show*` toggles so a parsed value can flow straight from the read side back into the write side without transformation.
- **Reader** (`parseDataLabels`): walks `<c:dLbls>` for `<c:showLeaderLines val=".."/>`. Only the literal OOXML falsy spellings (`val="0"` / `val="false"`) surface `false`; the OOXML default `val="1"` (and `val="true"`, missing `val`, unknown tokens, absence) collapses to `undefined`. Permissive on chart family — surfaces the flag on any chart's data labels so a templated chart whose chart-type element ends up coerced still surfaces the source's intent.
- **Writer**: emits `<c:showLeaderLines val="0"/>` only when `dl.showLeaderLines === false` AND the chart type is `pie` or `doughnut`. The OOXML schema scopes the element to `EG_DLbls` (pie / doughnut only — every other family routes through `EG_DLblsShared` which omits it). Bar / column / line / area / scatter silently drop the override to mirror Excel's reference serialization; emitting it on those families would break Excel's strict validator. The element slots between `<c:separator>` and `<c:extLst>` per the `EG_DLbls` sequence.
  - Threads `chartType: chart.type` through `SeriesOptions` so per-series `<c:dLbls>` blocks honour the same scope guard as the chart-level block.
- **Clone**: the existing `{ ...sourceLabels }` spread in `resolveChartDataLabels` and `resolveSeriesDataLabels` carries the new field forward without changes — the standard `undefined` (inherit) / `null` (drop) / object (replace) grammar applies wholesale just as it does for `showLegendKey` / `numberFormat`. The cloned model preserves `showLeaderLines: false` even after a chart-type coercion to a non-pie family; the writer is the layer that drops it on emit.
- **Validation**: only literal `boolean` survives the type guard at the write layer; non-boolean inputs collapse to the default. This matches how `showLegendKey` / `legendOverlay` / `roundedCorners` / axis `hidden` treat their inputs.

## Test plan

- [x] `parseChart — data labels showLeaderLines` (11 new tests) — surfaces `false` on `val="0"`, collapses `val="1"` and absence to `undefined`, OOXML truthy / falsy spellings, missing `val`, unknown tokens, makes the flag alone enough to surface a record, surfaces on per-series `<c:dLbls>`, co-surfaces alongside numberFormat and other show* toggles, permissive on non-pie families.
- [x] `writeChart — data labels showLeaderLines` (15 new tests) — omits the element by default on a pie chart, emits `val="0"` on pie / doughnut overrides, treats `true` like absence, collapses non-boolean inputs, scope-guards bar / column / line / area / scatter, OOXML schema position (after `<c:separator>` / `<c:showBubbleSize>`), exactly-one-element guard, threads through series-level dLbls, scope-guards series-level overrides on column, composes with showLegendKey / numberFormat, parse round-trip, full `writeXlsx` package round-trip.
- [x] `cloneChart — data labels showLeaderLines` (11 new tests) — inherit / null drop / wholesale replace / per-series inherit / per-series null drop semantics, source-without-flag pickup, composition with other data-labels fields, chart-type coercion (pie → doughnut preserves; pie → bar carries through but writer drops), end-to-end parse → clone → write → reparse, full `writeXlsx` cloned-chart round-trip.
- [x] `pnpm test` — all 4099 tests pass (lint + typecheck + vitest).

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -634,6 +634,26 @@ export interface ChartDataLabels {
    * from the source cell formatting.
    */
   numberFormat?: ChartAxisNumberFormat;
+  /**
+   * Render the leader lines that connect each data label back to its
+   * pie / doughnut slice when Excel pushes a label outside the slice it
+   * belongs to. Mirrors Excel's "Format Data Labels -> Label Options ->
+   * Show Leader Lines" checkbox.
+   *
+   * Maps to `<c:showLeaderLines val=".."/>` inside `<c:dLbls>`, which
+   * sits at the tail of the `EG_DLbls` group (after `<c:separator>`,
+   * before `<c:extLst>`). The OOXML schema scopes the element to
+   * pie / doughnut chart families exclusively (`EG_DLbls` for
+   * `CT_PieChart` / `CT_DoughnutChart` only — bar / column / line /
+   * area / scatter route through `EG_DLblsShared` which omits it). The
+   * writer drops the field silently on every non-pie / non-doughnut
+   * family to mirror Excel's reference serialization.
+   *
+   * The OOXML default is `true` (Excel paints leader lines on every
+   * label that gets pushed outside its slice). Set to `false` to opt
+   * out and render labels without the connecting lines.
+   */
+  showLeaderLines?: boolean;
 }
 
 /**
@@ -2720,6 +2740,25 @@ export interface ChartDataLabelsInfo {
    * empty.
    */
   numberFormat?: ChartAxisNumberFormat;
+  /**
+   * Mirror of {@link ChartDataLabels.showLeaderLines}. Surfaces the
+   * `<c:showLeaderLines val=".."/>` flag parsed from the source
+   * `<c:dLbls>` block — Excel's "Format Data Labels -> Show Leader
+   * Lines" checkbox.
+   *
+   * The OOXML default is `true`. Surfaces `false` only when the source
+   * pinned `<c:showLeaderLines val="0"/>`; absence and the default
+   * `val="1"` collapse to `undefined` so a re-parse of a writer that
+   * omits the element matches a re-parse of one that pins the default
+   * explicitly.
+   *
+   * The element only appears on pie / doughnut data-labels per the
+   * OOXML schema (`EG_DLbls` is scoped to `CT_PieChart` /
+   * `CT_DoughnutChart`); the parser surfaces it on every chart family
+   * the source emits it for so a templated chart can round-trip
+   * cleanly even when the chart-type element ends up coerced.
+   */
+  showLeaderLines?: boolean;
 }
 
 /**
--- a/src/xlsx/chart-reader.ts
+++ b/src/xlsx/chart-reader.ts
@@ -1439,6 +1439,29 @@ function parseDataLabels(el: XmlElement): ChartDataLabelsInfo | undefined {
   const numFmt = parseDataLabelsNumberFormat(el);
   if (numFmt) out.numberFormat = numFmt;
 
+  // `<c:showLeaderLines val=".."/>` mirrors Excel's "Format Data
+  // Labels -> Show Leader Lines" checkbox. The OOXML default is
+  // `true` (Excel paints leader lines on every label that gets pushed
+  // outside its slice), so absence and `val="1"` collapse to
+  // `undefined` — only an explicit `val="0"` (or `"false"`) surfaces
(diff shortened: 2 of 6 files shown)


productdevbook (commented):
Reviewed the diff end-to-end (types, reader, writer, three test suites — 37 new tests).

Looks great:
- Type model is symmetric across `ChartDataLabels` (writer) and `ChartDataLabelsInfo` (reader); the `?: boolean` shape lines up with the existing `show*` toggles so a parsed value flows straight back through the writer with no transformation.
- Reader is appropriately strict on input — only literal OOXML falsy spellings (`"0"` / `"false"`) surface `false`; `val="1"`, missing `val`, unknown tokens, and absence all collapse to `undefined`. Empty-record collapse is updated to include the new field, and `showLeaderLines: false` alone is enough to surface a record (correct — Excel honors that pin even on an otherwise empty `<c:dLbls>`).
- Reader is permissive on chart family (surfaces the flag on a coerced bar template) while the writer enforces the OOXML schema scope (pie / doughnut only via `EG_DLbls`). The split is the right one: the cloned model preserves intent, the writer drops on emit. The test that pies-to-bar coerces and round-trips through `writeChart` confirms the wire-level scope guard.
- `SeriesOptions.chartType` was promoted to required (and `options` to required on `buildSeries`) so per-family callers cannot forget to thread it through — every chart builder (`buildBarChart` / `buildLineChart` / `buildAreaChart` / `buildPieChart` / `buildDoughnutChart` / `buildScatterChart`) was updated. The series-level scope guard test on a column chart locks in the pattern.
- Element placement after `<c:separator>` / `<c:showBubbleSize>` matches the `EG_DLbls` sequence, and the exactly-one-emission guard test prevents accidental double-emit regressions.
- Clone semantics fall out of the existing `{ ...sourceLabels }` spread for free; the `null` drop / wholesale replace tests cover all three grammar paths.

Non-blocking follow-up: the README has dedicated prose for `showLegendKey` under both the reader (~line 638) and writer (~line 1112) chart-data-labels sections. A short paragraph for `showLeaderLines` (mirroring the `showLegendKey` shape — "pie / doughnut only, OOXML default `true`, only `false` flips it") would be a nice touch in a follow-up docs PR for symmetry, but is not required to ship this.

CI green, merging.

PR #245, opened 2026-05-04
feat(xlsx): chart axis tick-label rotation — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:catAx><c:txPr><a:bodyPr rot="N"/></c:txPr></c:catAx>` (and the matching `<c:valAx>` slot) at the read, write, and clone layers so a template's axis-label rotation pin survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<a:bodyPr rot="N"/>` mirrors Excel's "Format Axis -> Alignment -> Custom angle" knob — `N` is in 60000ths of a degree, so 45° serializes as `rot="2700000"` and -90° as `rot="-5400000"`. Until now hucre's writer had no slot for the element and the reader ignored it entirely, so a templated dashboard chart whose user rotated the category labels diagonally (a common pattern for fitting long category strings under a column chart's bars) silently lost the choice on every parse -> clone -> write loop. Bridges another category-axis customization gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.axes?.x?.labelRotation);
// 45         ← when the template pinned <a:bodyPr rot="2700000"/> on the catAx
// undefined  ← when the template emits no rotation (the OOXML default 0)

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [
      ["Quarter", "Revenue"],
      ["Q1 2026 (Americas)", 100],
      ["Q2 2026 (Americas)", 200],
    ],
    charts: [{
      type: "column",
      title: "Quarterly Revenue",
      // Tilt long category labels diagonally so they fit under the
      // column bars without overlapping their neighbours.
      axes: { x: { labelRotation: -45 } },
    }],
  }],
});

// -- Clone-through --
//   → inherits the source's parsed labelRotation unchanged.
cloneChart(source, {
  axes: { x: { labelRotation: null } },
});
//   → drops the inherited rotation (writer skips <c:txPr> entirely).
cloneChart(source, {
  axes: { x: { labelRotation: 30 } },
});
//   → replaces the inherited rotation with 30 degrees.
```

## Implementation notes

- **Type model**: `axes.x.labelRotation?: number` and `axes.y.labelRotation?: number` on `SheetChart` (writer side) plus the matching `labelRotation?: number` on `ChartAxisInfo` (reader side). Same shape as the other numeric axis fields (`tickLblSkip`, `lblOffset`) so a parsed value flows straight from the read side back into the write side without transformation. Callers pin the value in degrees; the writer converts to 60000ths at emit time.
- **Reader** (`parseAxisLabelRotation`): walks `<c:txPr><a:bodyPr rot="N"/></c:txPr>` for the `rot` attribute. `Number.parseInt` rejects non-numeric tokens; `Math.round(parsed / 60000)` converts to whole degrees. The OOXML default `0` and absence both collapse to `undefined`. Out-of-range values clamp to the `-90..90` band Excel's UI exposes so a corrupt template cannot surface a value the writer would never emit. Surfaces on every axis flavour — the OOXML schema places `<c:txPr>` on `CT_CatAx` / `CT_ValAx` / `CT_DateAx` / `CT_SerAx` alike.
- **Writer**: `buildAxisTxPr(rotationDeg)` returns `undefined` when the rotation is unset / 0 / non-finite (writer skips emission), or a minimal `<c:txPr>` shell — `<a:bodyPr rot="N"/>` carries the rotation, `<a:lstStyle/>` is the empty list-style placeholder the schema requires, and `<a:p><a:pPr><a:defRPr/></a:pPr><a:endParaRPr/></a:p>` is the empty paragraph stub Excel always emits even when no per-run styling is pinned. Additional `<a:bodyPr>` attributes Excel writes in its full reference (`spcFirstLastPara` / `vertOverflow` / `wrap` / `anchor` / `anchorCtr`) are intentionally omitted — the OOXML schema marks them all optional, and dropping them keeps the writer's footprint minimal while preserving the rotation intent. The block slots between `<c:tickLblPos>` and `<c:crossAx>` per the CT_CatAx / CT_ValAx sequence. Threads through bar / column / line / area / scatter via `buildBarAxes` / `buildScatterAxes`; pie / doughnut have no axes at all and the field is silently dropped on those families.
  - `normalizeAxisLabelRotation(value)` clamps inputs to `-90..90`, rounds fractional inputs to the nearest whole degree, and collapses `0` / `NaN` / `Infinity` / non-numeric inputs to `undefined`.
- **Clone**: `applyLabelRotationOverride` follows the standard `number | null` override grammar — `undefined` inherits, `null` drops, a literal number replaces. Out-of-range / non-numeric overrides clamp / collapse the same way as the writer's normalizer, so the cloned `SheetChart` always carries a value the writer will accept. The catAx / valAx scope rule applies via the same `pie / doughnut` short-circuit the surrounding axis helpers use — those families have no axes at all and the field is silently dropped.
- **Validation**: only literal finite numbers survive the type guard at every layer; non-numeric inputs collapse to the default. This matches how `tickLblSkip` / `tickMarkSkip` / `lblOffset` treat their inputs.

## Test plan

- [x] `parseChart — axis labelRotation` (14 new tests) — surfaces the rotation in whole degrees from the `rot` attribute, surfaces negative rotations literally, collapses the OOXML default `rot="0"` to undefined, returns undefined when `<c:txPr>` is absent / `<a:bodyPr>` is absent / `rot` is missing, clamps out-of-range values to `-90..90`, drops non-numeric tokens, rounds non-integer 60000ths, surfaces on the value axis, surfaces independently on both axes, surfaces on scatter Y axis, co-surfaces alongside other axis fields.
- [x] `writeChart — axis labelRotation` (20 new tests) — omits `<c:txPr>` by default, emits the X-axis rotation, emits negative rotations as negative 60000ths, emits the Y-axis rotation, emits independently on both axes, collapses the OOXML default `0` to absence, clamps above-90 / below-`-90` to the band endpoints, rounds non-integer degrees, drops non-finite / non-numeric inputs, places `<c:txPr>` between `<c:tickLblPos>` and `<c:crossAx>` per the OOXML schema, threads through bar / column / line / area, threads through scatter (both axes), drops on pie / doughnut, single-emit guard, minimal CT_TextBody shape, parse round-trip, default round-trip collapse, full `writeXlsx` package round-trip.
- [x] `cloneChart — axis labelRotation` (14 new tests) — inherit / null drop / replace semantics, clamp out-of-range overrides, round fractional overrides, collapse 0 to undefined, collapse non-finite overrides, add to a source axis without one, drop on pie / doughnut chart-type coercion, carry through bar -> column coercion, compose with other axis overrides, end-to-end parse -> clone -> write -> reparse, full `writeXlsx` cloned-chart round-trip.
- [x] `pnpm test` — all 4110 tests pass (lint + typecheck + vitest).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1077,7 +1077,7 @@ charts; `lineGrouping` and `areaGrouping` accept
 `top` / `bottom` / `left` / `right` / `topRight` / `false`, and
 `altText` / `frameTitle` flow through to the drawing's `xdr:cNvPr`
 attributes for screen readers.
-`axes: { x: { title, gridlines, scale, numberFormat, majorTickMark, minorTickMark, tickLblPos, reverse, crosses, crossesAt, dispUnits, crossBetween }, y: { title, gridlines, scale, numberFormat, majorTickMark, minorTickMark, tickLblPos, reverse, crosses, crossesAt, dispUnits, crossBetween } }`
+`axes: { x: { title, gridlines, scale, numberFormat, majorTickMark, minorTickMark, tickLblPos, labelRotation, reverse, crosses, crossesAt, dispUnits, crossBetween }, y: { title, gridlines, scale, numberFormat, majorTickMark, minorTickMark, tickLblPos, labelRotation, reverse, crosses, crossesAt, dispUnits, crossBetween } }`
 attaches per-axis labels, gridlines, numeric scaling, the tick-label
 number format and the tick-rendering trio — `x` lands inside
 `<c:catAx>` (or the X value axis for scatter), `y` inside the value
@@ -1488,6 +1488,25 @@ exclusively (even `<c:dateAx>`, `<c:valAx>`, and `<c:serAx>` reject
 it), so the flag threads through bar / column / line / area but is
 silently dropped on scatter (both axes are value axes) and pie /
 doughnut (no axes at all).
+The `axes.x.labelRotation` and `axes.y.labelRotation` fields pin the
+tick-label rotation in whole degrees, mapping to
+`<c:txPr><a:bodyPr rot="N"/></c:txPr>` on the matching axis — Excel's
+"Format Axis -> Alignment -> Custom angle" knob. The OOXML `rot`
+attribute is in 60000ths of a degree; the writer converts at emit
+time so callers pin the value in degrees. Useful for rotating long
+category labels diagonally so they fit underneath a column chart's
+bars without overlapping their neighbours, or for tilting a value
+axis's number labels when a tight chart frame would otherwise crowd
+them. Range: `-90..90` (Excel's UI band — out-of-range values clamp
+to the nearest endpoint, fractional inputs round to the nearest whole
+degree, non-finite / non-numeric inputs drop). The writer omits the
+entire `<c:txPr>` block on the OOXML default `0` (and on absence) so a
+fresh chart matches Excel's minimal serialization. The element sits on
+every axis flavour per the OOXML schema (`<c:catAx>` / `<c:valAx>` /
+`<c:dateAx>` / `<c:serAx>` all carry an optional `<c:txPr>`), so the
+field threads through every chart family that has axes (bar / column /
+line / area / scatter); pie / doughnut have no axes at all and the
+field is silently dropped there.
 The `axes.x.reverse` and `axes.y.reverse` flags map to
 `<c:scaling><c:orientation val="maxMin"/></c:scaling>` — Excel's
 "Categories / Values in reverse order" toggle. On a category axis,
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1717,6 +1717,37 @@ export interface SheetChart {
        */
       tickLblPos?: ChartAxisTickLabelPosition;
       /**
+       * Tick-label rotation in degrees. Maps to
+       * `<c:catAx><c:txPr><a:bodyPr rot="N"/></c:txPr></c:catAx>` (or
+       * `<c:valAx>` for scatter / value axes) — Excel's "Format Axis ->
+       * Alignment -> Custom angle" knob. Useful for rotating long
+       * category labels diagonally so they fit underneath a column
+       * chart's bars without overlapping their neighbours.
+       *
+       * Accepted range: `-90..90` (the band Excel's UI exposes; values
+       * outside the band clamp to the nearest endpoint). The OOXML
+       * `rot` attribute is in 60000ths of a degree — the writer
+       * converts at emit time so callers pin the value in degrees.
+       *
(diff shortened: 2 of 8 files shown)


productdevbook (commented):
## Review

Surfaces axis tick-label rotation (`<c:txPr><a:bodyPr rot="N"/></c:txPr>`) end-to-end across reader, writer, and clone. The design is consistent with the surrounding axis fields (`tickLblSkip`, `lblOffset`, `tickLblPos`) — degrees on the public API, OOXML's 60000ths-of-a-degree conversion at the XML boundary, default-collapse to undefined for round-trip stability, and the standard `undefined`/`null`/value override grammar on the clone side.

### Correctness
- Conversion factor (60000) and -90..90 clamp band match Excel's UI semantics and OOXML's `ST_TextRotation`.
- Reader/writer/clone all collapse `0` to `undefined` so absence and the OOXML default round-trip identically.
- `<c:txPr>` is correctly placed between `<c:tickLblPos>` and `<c:crossAx>` per CT_CatAx / CT_ValAx, and threaded through bar / column / line / area / scatter while pie / doughnut short-circuit.
- Type guards reject non-finite and non-numeric inputs at every layer (parser, writer, cloner), so a stray runtime value cannot leak past the writer's strict-validator contract.

### Test coverage
48 new tests (14 parser + 20 writer + 14 clone) cover all the edge cases — defaults, clamps, fractional rounding, axis-flavour scoping, schema position, end-to-end round-trips through `writeXlsx`. After the merge with main the suite stands at 4147 tests passing.

### Fixes applied this round
- Resolved append-conflicts in `test/charts.test.ts`, `test/charts-write.test.ts`, and `test/chart-clone.test.ts` where the new `labelRotation` describe blocks landed alongside the data-labels `showLeaderLines` blocks merged into main. Both suites now sit side by side.
- Verified locally: `pnpm test` (lint + format + typecheck + 4147 vitest tests) and `pnpm build` both green.

### Minor nits (non-blocking)
- The PR body documents the empty-paragraph stub as `<a:p><a:pPr><a:defRPr/></a:pPr><a:endParaRPr/></a:p>` but `buildAxisTxPr` actually emits `<a:endParaRPr lang="en-US"/>`. Excel itself emits the `lang` attribute, so the implementation matches Excel — only the doc snippet trails the code. Not worth a re-push.
- `LABEL_ROTATION_MIN_DEG` / `LABEL_ROTATION_MAX_DEG` are duplicated between the reader and the writer. Acceptable for now since the two layers stay decoupled and the constants are tiny, mirroring how `tickLblSkip` and friends were structured in earlier PRs.

LGTM — auto-merging once CI finishes.

PR #246, opened 2026-05-04
feat(xlsx): chart upDownBars gap width — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:upDownBars><c:gapWidth val=\"N\"/></c:upDownBars>` (CT_UpDownBars, ECMA-376 Part 1, §21.2.2.218) at the read, write, and clone layers so a templated line chart's up/down-bars spacing pin survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

Until now hucre's writer hard-coded `<c:gapWidth val=\"150\"/>` (the OOXML default Excel itself emits) and the reader ignored the element entirely, so a template whose author thinned or widened the up/down bars from the default lost the choice on every parse -> clone -> write loop. The `ST_GapAmount` schema (`0..500`, percentage of the bar width) is shared with the bar-chart `<c:gapWidth>`, but the two are independently scoped — the bar-family value spaces category groups while this one spaces the up/down bars themselves on a line chart. Bridges another line-chart gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.upDownBars);            // true
console.log(source.upDownBarsGapWidth);    // 200 (when the template pinned val=\"200\")
                                           // undefined (when the template emits val=\"150\" or omits the child)

// -- Write side --
await writeXlsx({
  sheets: [{
    name: \"Stock\",
    rows: [
      [\"Open\", \"Close\"],
      [10, 12],
      [12, 11],
    ],
    charts: [{
      type: \"line\",
      series: [
        { name: \"Open\",  values: \"A2:A3\" },
        { name: \"Close\", values: \"B2:B3\" },
      ],
      // Pack the up / down bars tighter than Excel's reference 150% gap.
      upDownBars: true,
      upDownBarsGapWidth: 50,
    }],
  }],
});

// -- Clone-through --
//   → inherits the source's parsed upDownBarsGapWidth unchanged.
cloneChart(source, {
  upDownBarsGapWidth: null,
});
//   → drops the inherited gap width (writer falls back to the OOXML default 150).
cloneChart(source, {
  upDownBarsGapWidth: 75,
});
//   → replaces the inherited gap width with 75%.
```

## Implementation notes

- **Type model**: `upDownBarsGapWidth?: number` on `SheetChart` (writer side) and the matching `upDownBarsGapWidth?: number` on `Chart` (reader side). Same shape as the existing bar-family `gapWidth` so a parsed value flows straight from the read side back into the write side without transformation.
- **Reader** (`parseUpDownBarsGapWidth`): walks `<c:upDownBars><c:gapWidth val=\"N\"/></c:upDownBars>` only when the parent `<c:upDownBars>` element is present on a `CT_LineChart` / `CT_Line3DChart` / `CT_StockChart` body. The OOXML default `150` and absence both collapse to `undefined`. Out-of-range values (outside the `ST_GapAmount` `0..500` band) are dropped rather than clamped so a corrupt template does not silently surface a value the writer would never emit.
- **Writer** (`buildUpDownBars` + `clampUpDownBarsGapWidth`): rounds non-integer inputs, drops out-of-range / non-finite inputs, then falls back to the OOXML default `150` when the resolved value is `undefined`. The `<c:gapWidth>` child is always emitted alongside `<c:upDownBars>` to mirror Excel's reference serialization. Silently dropped on bar / column / pie / doughnut / area / scatter chart kinds (no parent element to host the value) and on a chart that pins `upDownBars: false` (parent element collapses to absence).
- **Clone**: `resolveUpDownBarsGapWidth` follows the standard `number | null` override grammar — `undefined` inherits, `null` drops, a literal number replaces. The override is silently dropped when the resolved `upDownBars` is anything other than `true` (the OOXML schema scopes `<c:gapWidth>` exclusively to `<c:upDownBars>`, so leaking the value through to the cloned `SheetChart` would surface a setting the writer would never emit anyway). The clone is also gated on the resolved chart type being `line` for the same reason `upDownBars` itself is.
- **Validation**: only literal finite numbers survive the type guard at every layer; non-numeric inputs collapse to the default. This matches how `gapWidth` / `holeSize` / `firstSliceAng` treat their inputs.

## Test plan

- [x] `parseChart — upDownBars gap width` (13 new tests) — surfaces a custom gap width, the boundary values 0 and 500, collapses the OOXML default 150 to undefined, returns undefined when `<c:gapWidth>` is absent / `<c:upDownBars>` is absent, drops out-of-range / negative / non-numeric / missing-val inputs, ignores the field on bar charts, surfaces on stock and 3D line charts, does not collide with the bar-chart `gapWidth` field.
- [x] `writeChart — upDownBars gap width` (16 new tests) — emits a custom gap width, the boundary values 0 and 500, falls back to 150 when unset, rounds non-integer values, falls back to 150 when out-of-range / non-finite, drops the value when the parent toggle is off / the chart kind is non-line, single-emit guard, no collision with bar-chart `gapWidth`, parse round-trip, default round-trip collapse, full `writeXlsx` package round-trip.
- [x] `cloneChart — upDownBars gap width` (14 new tests) — inherit / null drop / replace semantics, drops on resolved `upDownBars: false / null`, drops a stray gap width on a source with no parent toggle, drops on a flatten to non-line / area, composes with an `upDownBars` override that adds the parent toggle, end-to-end parse -> clone -> write -> reparse, full `writeXlsx` cloned-chart round-trip with custom and null-drop overrides.
- [x] `pnpm test` — all 4143 tests pass (lint + typecheck + vitest).

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1472,11 +1472,39 @@ export interface SheetChart {
    *
    * The writer emits a default `<c:gapWidth val="150"/>` child to
    * mirror Excel's reference serialization — `150` is the OOXML
-   * default for `CT_UpDownBars/gapWidth`. Custom gap widths are not
-   * exposed at this layer; pass a richer model in a follow-up if a
-   * caller needs to thin or widen the bars.
+   * default for `CT_UpDownBars/gapWidth`. Pin
+   * {@link upDownBarsGapWidth} to thin or widen the bars.
    */
   upDownBars?: boolean;
+  /**
+   * Width of the gap between up / down bars as a percentage of the bar
+   * width. Maps to `<c:lineChart><c:upDownBars><c:gapWidth val=".."/>
+   * </c:upDownBars></c:lineChart>` — Excel's "Format Up Bars / Down
+   * Bars -> Gap Width" slider. Accepted range: `0` – `500` (the OOXML
+   * `ST_GapAmount` schema). Excel's default is `150` (each up/down
+   * bar group's gap equals 1.5× the bar width); smaller values pack
+   * the bars tighter, `0` removes the gap entirely.
+   *
+   * Only meaningful when {@link upDownBars} is `true` — the writer
+   * silently drops the value on every other line chart configuration
+   * (the OOXML schema places `<c:gapWidth>` on `CT_UpDownBars`, so
+   * there is no slot for the value when the parent element is not
+   * emitted). The writer also drops the field on bar / column / pie /
+   * doughnut / area / scatter chart kinds for the same reason
+   * {@link upDownBars} is line-only.
+   *
+   * Out-of-range or non-finite values fall back to the OOXML default
+   * `150` so a fresh chart with a corrupt input still matches Excel's
+   * reference serialization. Non-integer values round to the nearest
+   * whole percent (Excel's UI accepts integer percentages only).
+   *
+   * Distinct from {@link gapWidth}: the bar-chart gap width controls
+   * spacing between category groups on a `<c:barChart>`, while this
+   * field controls the spacing between the up / down bars themselves
+   * on a line chart. Both share the same `ST_GapAmount` schema range
+   * but are independently scoped.
+   */
+  upDownBarsGapWidth?: number;
   /**
    * Whether the line chart paints markers at each data point. Maps to
    * `<c:lineChart><c:marker val=".."/></c:lineChart>` — Excel's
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed the diff end to end — types, reader, writer, and clone all line up cleanly. A few observations:

- Symmetric defaults across layers: the reader collapses `<c:gapWidth val="150"/>` to `undefined` and the writer falls back to `150` on absence, so absence and the OOXML default round-trip identically — same shape the existing `gapWidth` / `holeSize` / `firstSliceAng` knobs use.
- Strict drop-on-out-of-range policy on both sides (reader rejects `<0` / `>500`, writer's `clampUpDownBarsGapWidth` drops non-finite or out-of-band inputs to `undefined` so the writer's `?? 150` fallback kicks in). That matches the rationale in the doc comment — silently clamping `800` to `500` would mislead the caller about what Excel renders.
- Clone gating is correct: `upDownBarsGapWidth` only flows through when `resolvedUpDownBars === true`, so a flatten to column / area / pie / scatter or a `null` / `false` parent override drops the gap width silently. The OOXML schema scopes `<c:gapWidth>` exclusively to `<c:upDownBars>`, so this is the right policy.
- Test coverage is thorough (43 new tests across read / write / clone, including boundary 0 / 500, default-collapse, parent-toggle gating, bar-chart non-collision, stock + 3D line read paths, full `writeXlsx` round-trip).
- The single-emit guard test (`/<c:gapWidth\b/g` count === 1) is a nice regression hedge.

CI is SUCCESS, branch is mergeable, no concerns. Merging via squash.

One follow-up nit (non-blocking): the README sections that describe `Chart.upDownBars` / `SheetChart.upDownBars` / the cloner's `upDownBars` override do not yet mention `upDownBarsGapWidth`. Worth a quick docs pass in a follow-up commit alongside the next batch of chart features (matches the cadence of the recent `docs+web` commit).

PR #247, opened 2026-05-04
feat(xlsx): chart title rotation — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:title><c:tx><c:rich><a:bodyPr rot=\"N\"/></c:rich></c:tx></c:title>` (CT_Title body-properties, ECMA-376 Part 1, §21.2.2.210) at the read, write, and clone layers so a template's rotated chart-title pin survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<a:bodyPr rot=\"N\"/>` mirrors Excel's \"Format Chart Title -> Size & Properties -> Alignment -> Custom angle\" knob — `N` is in 60000ths of a degree, so 45° serializes as `rot=\"2700000\"` and -90° as `rot=\"-5400000\"`. Until now hucre's writer hardcoded `rot=\"0\"` on the title's rich-text body and the reader ignored the attribute entirely, so a templated dashboard chart whose user rotated the title vertically (a common pattern for fitting a long chart title into a narrow sidebar tile, or for mirroring the rotation of a tilted axis label set) silently lost the choice on every parse -> clone -> write loop. Bridges another title-customization gap for the dashboard composition flow tracked in #136.

Mirrors the axis-side `labelRotation` field added in #245 — same `-90..90` band, same conversion factor, same normalization grammar — so a single rotation value threads cleanly through both the chart title and its axis labels without bookkeeping the units.

Refs #136

## API

\`\`\`ts

// -- Read side --
console.log(source.titleRotation);
// 45         ← when the template pinned <a:bodyPr rot=\"2700000\"/> on the title body
// undefined  ← when the template emits no rotation (the OOXML default 0) or no <c:title>

// -- Write side --
await writeXlsx({
  sheets: [{
    name: \"Dashboard\",
    rows: [
      [\"Quarter\", \"Revenue\"],
      [\"Q1 2026\", 100],
      [\"Q2 2026\", 200],
    ],
    charts: [{
      type: \"column\",
      title: \"Quarterly Revenue\",
      // Stand the title on its side so it fits a narrow sidebar tile.
      titleRotation: -90,
    }],
  }],
});

// -- Clone-through --
//   → inherits the source's parsed titleRotation unchanged.
cloneChart(source, {
  titleRotation: null,
});
//   → drops the inherited rotation (writer falls back to the OOXML default rot=\"0\").
cloneChart(source, {
  titleRotation: 30,
});
//   → replaces the inherited rotation with 30 degrees.
\`\`\`

## Implementation notes

- **Type model**: `SheetChart.titleRotation?: number` (writer side) and `Chart.titleRotation?: number` (reader side). The clone option `titleRotation?: number | null` follows the standard `undefined` (inherit) / `null` (drop) / `number` (replace) grammar. Same shape as the axis-side `labelRotation` so a parsed value flows straight from the read side back into the write side without transformation. Callers pin the value in degrees; the writer converts to 60000ths at emit time.
- **Reader** (`parseTitleRotation`): walks `<c:title><c:tx><c:rich><a:bodyPr rot=\"N\"/></c:rich></c:tx></c:title>` for the `rot` attribute. The lookup is scoped to the title body (`<c:tx><c:rich>`) so a stray `<a:bodyPr>` elsewhere in the chart cannot leak in. `Number.parseInt` rejects non-numeric tokens; `Math.round(parsed / 60000)` converts to whole degrees. The OOXML default `0` and absence both collapse to `undefined`. Out-of-range values clamp to the `-90..90` band Excel's UI exposes so a corrupt template cannot surface a value the writer would never emit. Returns `undefined` when the chart omits the `<c:title>` element entirely or when the title is a `<c:strRef>` (formula reference) with no `<c:rich>` body.
- **Writer**: previously hard-coded `rot=\"0\"` in the title body's `<a:bodyPr>` attributes. Now resolves through `resolveTitleRotation(chart)` — `normalizeTitleRotation` clamps inputs to `-90..90`, rounds fractional inputs to the nearest whole degree, and collapses `0` / `NaN` / `Infinity` / non-numeric inputs to `undefined`. The writer threads the resolved value into the existing `buildTitle` helper so the rotation slots into the same `<a:bodyPr>` attribute that already carries the title body's other layout metadata (`spcFirstLastPara` / `vertOverflow` / `wrap` / `anchor` / `anchorCtr` are preserved exactly as before — only the `rot` attribute changes). Threads through every chart family that emits a literal title (bar / column / line / pie / doughnut / area / scatter); silently dropped when no title is rendered (`title` absent or `showTitle === false`) — there is no `<c:title>` block to host the rotation in either case.
- **Clone** (`resolveTitleRotation` + `normalizeTitleRotation`): follows the standard `number | null` override grammar — `undefined` inherits, `null` drops, a literal number replaces. Out-of-range / non-numeric overrides clamp / collapse the same way as the writer's normalizer, so the cloned `SheetChart` always carries a value the writer will accept. The title-presence scope rule applies via the same `titleRendered` short-circuit that gates `titleOverlay` — when the resolved chart renders no title, the inherited rotation drops silently so the cloned `SheetChart` accurately reflects what the chart will paint.
- **Validation**: only literal finite numbers survive the type guard at every layer; non-numeric inputs collapse to the default. This matches how the axis-side `labelRotation` field treats its inputs.

## Test plan

- [x] `parseChart — title rotation` (12 new tests) — surfaces the rotation in whole degrees from the `rot` attribute, surfaces negative rotations literally, collapses the OOXML default `rot=\"0\"` to undefined, returns undefined when `<a:bodyPr>` omits `rot` / `<c:title>` is absent / the title is a `<c:strRef>` formula reference, clamps out-of-range values to `-90..90`, drops non-numeric / empty tokens, rounds non-integer 60000ths, co-surfaces alongside `titleOverlay` and the title text.
- [x] `writeChart — title rotation` (18 new tests) — emits `rot=\"0\"` by default, emits the rotation as 60000ths on positive / negative / endpoint inputs, collapses the OOXML default `0` to absence, clamps above-90 / below-`-90` to the band endpoints, rounds non-integer degrees, drops non-finite / non-numeric inputs back to the default, ignores the rotation when the chart renders no title (`showTitle: false` or `title` absent), composes with `titleOverlay`, threads through every chart family (bar / column / line / pie / doughnut / area / scatter), exactly-one-`<a:bodyPr>` guard, no axis-style `<c:txPr>` emitted on the title, parse round-trip, default round-trip collapse, full `writeXlsx` package round-trip.
- [x] `cloneChart — title rotation` (18 new tests) — inherit / replace / null drop / `0` drop / non-finite drop semantics, clamp out-of-range overrides, round fractional overrides, defensive clamp on a parsed source rotation, source-without-rotation pickup, line -> column / line -> doughnut chart-type coercion, drop on `title: null`, drop on `showTitle: false`, drop on override-`title: null`, composition with `titleOverlay`, end-to-end parse -> clone -> write -> reparse, full `writeXlsx` cloned-chart round-trip.
- [x] `pnpm test` — all 4196 tests pass (lint + typecheck + vitest).

### DIFF
--- a/README.md
+++ b/README.md
@@ -925,6 +925,21 @@ truthy / falsy spellings (`"1"` / `"true"` / `"0"` / `"false"`);
 unknown values and missing `val` attributes drop to `undefined`. The
 flag is dropped whenever the chart omits the `<c:title>` element
 entirely — there is no overlay slot to surface in that case.
+`Chart.titleRotation` surfaces the chart-title rotation pulled from
+`<c:title><c:tx><c:rich><a:bodyPr rot="N"/></c:rich></c:tx></c:title>`
+— Excel's "Format Chart Title → Size & Properties → Alignment → Custom
+angle" knob. The OOXML attribute is in 60000ths of a degree; the reader
+divides by 60000 (rounding) to surface a whole-degree value in the
+`-90..90` band Excel's UI exposes. The OOXML default `0` (and absence
+of the `<a:bodyPr>` element / `rot` attribute) collapses to `undefined`
+so absence and the default round-trip identically through `cloneChart`.
+Out-of-range values clamp to the nearest endpoint of the `-90..90` band;
+non-numeric tokens drop back to `undefined`. The field is dropped
+whenever the chart omits the `<c:title>` element entirely — there is no
+rotation to surface in that case. Mirrors the axis-side
+`ChartAxisInfo.labelRotation` field — same range, same conversion
+factor — so a parsed value threads straight back into the writer-side
+`SheetChart.titleRotation`.
 `Chart.autoTitleDeleted` surfaces the chart-level
 `<c:chart><c:autoTitleDeleted val=".."/>` flag — Excel's record of
 whether the user explicitly deleted the auto-generated title that
@@ -1284,6 +1299,22 @@ emitted (no `title` set or `showTitle: false`) — there is no `<c:title>`
 block to host the overlay child in either case. Independent of
 `legendOverlay`: the legend and title `<c:overlay>` elements live on
 different parents, so the two flags compose freely.
+The chart-level `titleRotation` field maps to `<a:bodyPr rot="N"/>`
+inside `<c:title><c:tx><c:rich>` — Excel's "Format Chart Title → Size &
+Properties → Alignment → Custom angle" knob. The OOXML attribute is in
+60000ths of a degree, so `titleRotation: 45` serializes as
+`rot="2700000"` and `titleRotation: -90` as `rot="-5400000"`; the writer
+performs the conversion at emit time. Accepted range: `-90..90` (Excel's
+UI band) — out-of-range inputs clamp to the nearest endpoint, non-integer
+inputs round to the nearest whole degree, and `0` / `NaN` / `Infinity` /
+non-numeric inputs collapse to absence so the writer falls back to the
+OOXML default `rot="0"` (horizontal). The flag is silently ignored when
+no title is emitted (no `title` set or `showTitle: false`) — there is no
+`<c:title>` block to host the rotation in either case. Mirrors the axis-
+side `axes.x.labelRotation` / `axes.y.labelRotation` fields — same
+range, same conversion factor — so a caller can thread a single rotation
+value through both the chart title and an axis label set without
+bookkeeping the units.
 The chart-level `autoTitleDeleted` field maps to
 `<c:autoTitleDeleted val=".."/>` inside `<c:chart>` — Excel's record of
 whether the user explicitly deleted the auto-generated title that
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
Reviewed the chart title rotation feature. The implementation cleanly mirrors the established axis-side `labelRotation` pattern (#245) — same -90..90 band, same 60000ths conversion, same normalize-and-clamp grammar — so the read/write/clone surfaces feel cohesive.

Checks:
- Read side scopes the lookup to `<c:title><c:tx><c:rich><a:bodyPr>` so a stray `bodyPr` elsewhere cannot leak in. Strict numeric guard (`Number.parseInt` + `Number.isFinite`) drops malformed tokens, and the OOXML default 0 collapses to undefined for symmetric round-trip with absence.
- Writer threads through `buildTitle` so all chart families (bar/column/line/pie/doughnut/area/scatter) inherit the rotation. Title-presence gating on `showTitle && chart.title` correctly drops the field when there is no `<c:title>` slot to host it.
- Clone follows the standard `undefined`/`null`/`number` override grammar, gates on `titleRendered`, and re-runs the writer's normalizer so the cloned SheetChart is guaranteed to round-trip cleanly.
- 48 new tests cover positive/negative/endpoint/clamp/round/non-finite/non-numeric paths, every chart family, the formula-reference (`<c:strRef>`) escape hatch, composition with `titleOverlay`, and full parse -> clone -> write -> reparse loops including `writeXlsx` packaging.
- README updates explain the field on the read, write, and clone sides with the same depth as adjacent fields.

CI is green. Merging.

PR #248, opened 2026-05-04
feat(xlsx): chart axis title rotation — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:catAx><c:title><c:tx><c:rich><a:bodyPr rot=\"N\"/></c:rich></c:tx></c:title></c:catAx>` (and the matching `<c:valAx>` shape — CT_Title body-properties on every axis flavour, ECMA-376 Part 1, §21.2.2.7 / §21.2.2.32) at the read, write, and clone layers so a template's rotated axis-title pin survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<a:bodyPr rot=\"N\"/>` mirrors Excel's "Format Axis Title -> Size & Properties -> Alignment -> Custom angle" knob — `N` is in 60000ths of a degree, so 45° serializes as `rot=\"2700000\"` and -90° as `rot=\"-5400000\"`. Until now hucre's writer hardcoded `rot=\"0\"` on the axis title's rich-text body and the reader ignored the attribute entirely, so a templated dashboard chart whose user stood the Y-axis title vertically (a common pattern that reclaims horizontal real estate next to the value labels) silently lost the choice on every parse -> clone -> write loop. Bridges another title-customization gap for the dashboard composition flow tracked in #136.

Mirrors the chart-level `titleRotation` field added in #247 — same `-90..90` band, same conversion factor (60000ths-of-a-degree), same `0` / absence / non-finite collapse to the OOXML default `rot=\"0\"`, same normalization grammar — so a single rotation value threads cleanly through both the chart title and either axis title without bookkeeping the units.

Refs #136

## API

```ts
import { readXlsx, writeXlsx, cloneChart, parseChart } from \"hucre\";

// -- Read side --
const wb = await readXlsx(templateBytes);
const source = wb.sheets[0].charts![0];
console.log(source.axes?.x?.axisTitleRotation);
// 45         ← when the template pinned <a:bodyPr rot=\"2700000\"/> on the X-axis title body
// undefined  ← when the template emits no rotation (the OOXML default 0) or no <c:title>

// -- Write side --
await writeXlsx({
  sheets: [{
    name: \"Dashboard\",
    rows: [
      [\"Quarter\", \"Revenue\"],
      [\"Q1 2026\", 100],
      [\"Q2 2026\", 200],
    ],
    charts: [{
      type: \"column\",
      title: \"Quarterly Revenue\",
[...]

PR #251, opened 2026-05-06
feat(xlsx): chart title font size — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr sz=\"N\"/></a:pPr><a:r><a:rPr sz=\"N\"/></a:r></a:p></c:rich></c:tx></c:title>` (CT_Title's default-paragraph and run text-run-properties — ECMA-376 Part 1, §21.2.2.4 / §21.2.2.114) at the read, write, and clone layers so a template's custom title font size survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<a:defRPr sz=\"N\"/>` mirrors Excel's "Format Chart Title -> Font -> Size" knob — `N` is in 100ths of a point, so 18pt serializes as `sz=\"1800\"` and 24pt as `sz=\"2400\"`. Until now hucre's writer hardcoded `sz=\"1400\"` (14pt) on both the title's default-paragraph `<a:defRPr>` and the literal run's `<a:rPr>` and the reader ignored the attribute entirely, so a templated dashboard chart whose user scaled the title up for a hero tile or down to fit a tight header slot silently lost the choice on every parse -> clone -> write loop. Bridges another title-customization gap for the dashboard composition flow tracked in #136.

Mirrors the chart-level `titleRotation` field added in #247 / #248 — same `1..400`pt band the OOXML `ST_TextFontSize` schema exposes, same 0.5pt half-step granularity Excel's UI exposes, same out-of-range / non-finite drop semantics, same title-presence scope rule (the field is silently ignored when `showTitle === false` or the chart has no literal title) — so a single configuration call threads cleanly through both the chart title's rotation and its size without bookkeeping the units.

Refs #136

## API

```ts

// -- Read side --
console.log(source.titleFontSize);
// 24         ← when the template pinned <a:defRPr sz=\"2400\"/> on the title's <a:p><a:pPr>
// 14         ← when the template emits sz=\"1400\" (Excel's reference 14pt)
// undefined  ← when the chart has no <c:title> or the title is a <c:strRef> formula reference

// -- Write side --
await writeXlsx({
  sheets: [{
    name: \"Dashboard\",
    rows: [
      [\"Region\", \"Revenue\"],
      [\"North\", 100],
      [\"South\", 200],
    ],
    charts: [{
      type: \"column\",
      title: \"Quarterly Revenue\",
      // Scale the title up for a hero dashboard tile.
      titleFontSize: 24,
    }],
  }],
});

// -- Clone-through --
//   → inherits the source's parsed titleFontSize unchanged.
cloneChart(source, {
  titleFontSize: null,
});
//   → drops the inherited size (writer falls back to Excel's default 14pt).
cloneChart(source, {
  titleFontSize: 18,
});
//   → replaces the inherited size with 18pt.
```

## Implementation notes

- **Type model**: `SheetChart.titleFontSize?: number` (writer side); `Chart.titleFontSize?: number` (reader side). The clone option `titleFontSize?: number | null` follows the standard `undefined` (inherit) / `null` (drop) / `number` (replace) grammar. Same shape as the chart-level `titleRotation` so a parsed value flows straight from the read side back into the write side without transformation. Callers pin the value in points; the writer converts to 100ths at emit time.
- **Reader** (`parseTitleFontSize`): walks `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr sz=\"N\"/></a:pPr></a:p></c:rich></c:tx></c:title>` for the `sz` attribute. The lookup is scoped to the title's default-paragraph slot so a stray `<a:defRPr>` elsewhere in the chart (e.g. on an axis title or a data-labels block) cannot leak in. `Number.parseInt` rejects non-numeric tokens; `Math.round((parsed / 100) * 2) / 2` converts to whole / half points and rounds to the nearest 0.5pt (matching Excel's UI granularity). Out-of-range values (outside the `1..400`pt band the OOXML `ST_TextFontSize` schema exposes) drop to `undefined` rather than fabricate a value the writer would never emit. Returns `undefined` when the chart has no `<c:title>` element, or when the title is a `<c:strRef>` (formula reference) with no `<c:rich>` body — there is no `<a:p>` slot to host the size in either case.
- **Writer**: `buildTitle` previously hard-coded `sz=\"1400\"` on both the default-paragraph `<a:defRPr>` and the literal run's `<a:rPr>`. Now resolves through a `fontSizePt?: number` parameter — `normalizeTitleFontSize` clamps inputs through the `1..400`pt band, rounds fractional inputs to the nearest 0.5pt, and collapses `NaN` / `Infinity` / non-numeric / out-of-range inputs to `undefined`. The size lands on both the `<a:defRPr>` and the `<a:rPr>` so a re-parse picks the value up off either canonical slot — Excel keeps both attributes in sync. `<a:bodyPr>` and the title's other layout attributes (`spcFirstLastPara` / `vertOverflow` / `wrap` / `anchor` / `anchorCtr` / `b=\"0\"` / `lang=\"en-US\"`) are preserved exactly as before; only the `sz` attribute changes. Absence (`undefined`) collapses to the application-default `1400` (14pt) so a fresh chart matches Excel's reference serialization byte-for-byte.
- **Clone** (`resolveTitleFontSize`): follows the standard `number | null` override grammar — `undefined` inherits, `null` drops, a literal number replaces. Out-of-range / non-numeric overrides clamp / collapse the same way as the writer's normalizer, so the cloned `SheetChart` always carries a value the writer will accept. The title-presence scope rule applies via the same `titleRendered` gate the writer uses — when the resolved title is dropped (`title: null` or `showTitle: false`), the inherited size drops silently so the cloned `SheetChart` accurately reflects what the chart will paint.
- **Validation**: only literal finite numbers within the `1..400`pt band survive the type guard at every layer; non-numeric inputs collapse to the default. This matches how the chart-level `titleRotation` field treats its inputs.

## Test plan

- [x] `parseChart — title font size` (13 new tests) — surfaces the size in points from the `sz` attribute, surfaces the default 14pt size literally, rounds non-integer 100ths to the nearest 0.5pt, returns undefined when `<a:defRPr>` omits `sz` / `<c:title>` is absent / the title is a `<c:strRef>` formula reference / `<a:p>` has no `<a:pPr>` element, drops below-1pt and above-400pt values, surfaces the band endpoints (1pt / 400pt) literally, drops non-numeric / empty tokens, co-surfaces alongside `titleRotation` / `titleOverlay` and other chart fields.
- [x] `writeChart — title font size` (15 new tests) — emits `sz=\"1400\"` by default on both the `<a:defRPr>` and the `<a:rPr>`, emits the size as 100ths on positive / fractional / endpoint inputs, supports half-point granularity, rounds fractional inputs to the nearest 0.5pt, emits the band endpoints (`sz=\"100\"` / `sz=\"40000\"`), drops out-of-range / non-finite / non-numeric inputs back to the 14pt default, ignores the size when the chart renders no title (`showTitle=false` or absent title), composes independently with `titleRotation` / `titleOverlay`, preserves the title body's other attributes (`b=\"0\"` / `lang=\"en-US\"`), threads through every chart family (bar / column / line / pie / scatter / area), parse round-trip, default round-trip, full `writeXlsx` package round-trip.
- [x] `cloneChart — title font size` (16 new tests) — inherit, replace, null drop, drop on out-of-range / non-finite overrides, half-step rounding on fractional overrides, defensive clamp on a parsed source size, drop on `title: null`, drop on `showTitle: false`, preserve when override pins a new title, composition with `titleRotation`, threads through line → column / line → doughnut flatten, end-to-end parse -> clone -> write -> reparse, full `writeXlsx` cloned-chart round-trip.
- [x] `pnpm test` — all 4339 tests pass (lint + typecheck + vitest).

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1342,6 +1342,33 @@ export interface SheetChart {
    */
   titleRotation?: number;
   /**
+   * Chart title font size in whole or half points. Maps to
+   * `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr sz="N"/></a:pPr>
+   * <a:r><a:rPr sz="N"/></a:r></a:p></c:rich></c:tx></c:title>` —
+   * Excel's "Format Chart Title -> Font -> Size" knob. The OOXML
+   * attribute is in 100ths of a point, so 18pt serializes as
+   * `sz="1800"` and 14pt (Excel's reference default) as `sz="1400"`;
+   * the writer performs the conversion at emit time and lands the
+   * value on both the default-paragraph `<a:defRPr>` and the literal
+   * run's `<a:rPr>` so a re-parse picks the size up off either
+   * canonical slot.
+   *
+   * Accepted range: `1..400`pt (the band the OOXML `ST_TextFontSize`
+   * schema exposes — `100..400000` in 100ths of a point). Fractional
+   * inputs round to the nearest 0.5pt (Excel's UI granularity); inputs
+   * outside the band, `NaN`, `Infinity`, and non-numeric inputs all
+   * collapse to `undefined` so the writer falls back to the default
+   * `14`pt size Excel itself emits on a fresh chart.
+   *
+   * Default: omitted — the title renders at Excel's default 14pt. Set
+   * an explicit value to scale the title up for a hero dashboard tile
+   * (e.g. `24`) or down to fit a tight header slot (e.g. `10`).
+   * Silently ignored when no title is rendered (`showTitle === false`
+   * or `title` is absent) — there is no `<c:title>` block to host the
+   * size in either case.
+   */
+  titleFontSize?: number;
+  /**
    * Auto-title-deleted flag. Maps to `<c:chart><c:autoTitleDeleted
    * val=".."/>` — Excel's record of whether the user explicitly deleted
    * the auto-generated title that single-series charts synthesise from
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed end-to-end. The implementation mirrors the chart-level `titleRotation` field added in #247/#248 cleanly: same `1..400`pt clamp band (matching the OOXML `ST_TextFontSize` `100..400000` schema), same 0.5pt half-step granularity from `Math.round(value * 2) / 2`, same drop semantics for `NaN`/`Infinity`/non-numeric/out-of-range, and the same title-presence scope rule via the `showTitle && chart.title` gate. The reader's `parseTitleFontSize` walks the canonical `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr sz>` chain and bails on the first missing link so a `<c:strRef>` formula-reference title cannot leak a stray `sz` from elsewhere in the chart. The writer lands the size on both `<a:defRPr>` and `<a:rPr>` so re-parse picks the value off either canonical slot, and the `b="0"` / `lang="en-US"` surrounding attrs are preserved byte-for-byte. The clone layer's `resolveTitleFontSize(undefined|null|number)` grammar matches the existing `titleRotation` / `titleOverlay` convention. Test coverage (44 new tests across reader/writer/clone) covers band endpoints, half-step rounding, scope drops, family threading, and a full template -> clone -> write -> reparse loop. CI is green. Merging.

PR #252, opened 2026-05-06
feat(xlsx): chart title bold flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr b=\".."/></a:pPr><a:r><a:rPr b=\".."/></a:r></a:p></c:rich></c:tx></c:title>` (CT_Title's default-paragraph and run text-character-properties bold flag — ECMA-376 Part 1, §21.1.2.3.7) at the read, write, and clone layers so a template's bold chart title survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<a:defRPr b=\"..\"/>` mirrors Excel's "Format Chart Title -> Font -> Bold" toggle. Until now hucre's writer hardcoded `b=\"0\"` (non-bold) on both the title's default-paragraph `<a:defRPr>` and the literal run's `<a:rPr>` and the reader ignored the attribute entirely, so a templated dashboard chart whose user bolded the title for emphasis silently lost the choice on every parse -> clone -> write loop. Bridges another title-customization gap for the dashboard composition flow tracked in #136.

Mirrors the chart-level `titleFontSize` field added in #251 — same canonical-slot pair (`<a:defRPr>` + `<a:rPr>`), same title-presence scope rule (silently ignored when `showTitle === false` or no literal title), same `boolean | null` clone grammar (`undefined` inherits, `null` drops, a literal boolean replaces) — so a single configuration call threads cleanly through every chart-title knob Excel exposes (bold, size, rotation, overlay) without bookkeeping the canonical OOXML slots.

Refs #136

## API

```ts

// -- Read side --
console.log(source.titleBold);
// true       ← when the template pinned <a:defRPr b="1"/> on the title's <a:p><a:pPr>
// undefined  ← when the template emits b="0" (Excel's reference non-bold)
// undefined  ← when the chart has no <c:title> or the title is a <c:strRef> formula reference

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [
      ["Region", "Revenue"],
      ["North", 100],
      ["South", 200],
    ],
    charts: [{
      type: "column",
      title: "Quarterly Revenue",
      // Bold the title for a hero dashboard tile.
      titleBold: true,
    }],
  }],
});

// -- Clone-through --
//   → inherits the source's parsed titleBold unchanged.
cloneChart(source, {
  titleBold: null,
});
//   → drops the inherited bold flag (writer falls back to OOXML default b="0").
cloneChart(source, {
  titleBold: false,
});
//   → replaces the inherited bold flag with explicit non-bold (b="0").
```

## Implementation notes

- **Type model**: `SheetChart.titleBold?: boolean` (writer side); `Chart.titleBold?: boolean` (reader side). The clone option `titleBold?: boolean | null` follows the standard `undefined` (inherit) / `null` (drop) / `boolean` (replace) grammar. Same shape as `titleFontSize` so a parsed value flows straight from the read side back into the write side without transformation.
- **Reader** (`parseTitleBold`): walks `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr b=\".."/></a:pPr></a:p></c:rich></c:tx></c:title>` for the `b` attribute. The lookup is scoped to the title's default-paragraph slot so a stray `<a:defRPr>` elsewhere in the chart (e.g. on an axis title or a data-labels block) cannot leak in. Uses the existing `parseBoolAttr` helper to accept the OOXML truthy / falsy spellings (`\"1\"` / `\"true\"` / `\"0\"` / `\"false\"`); unknown values and missing `b` attributes drop to `undefined`. The OOXML default `false` collapses to `undefined` so absence and `b=\"0\"` round-trip identically — only an explicit `b=\"1\"` surfaces `true`. Returns `undefined` when the chart has no `<c:title>` element, or when the title is a `<c:strRef>` (formula reference) with no `<c:rich>` body — there is no `<a:p>` slot to host the flag in either case.
- **Writer**: `buildTitle` previously hard-coded `b=\"0\"` (non-bold) on both the default-paragraph `<a:defRPr>` and the literal run's `<a:rPr>`. Now resolves through a `bold?: boolean` parameter — `normalizeTitleBold` accepts only literal booleans and collapses every other token (typed escapes from an untyped caller) to `undefined`. The flag lands on both the `<a:defRPr>` and the `<a:rPr>` so a re-parse picks the value up off either canonical slot — Excel keeps both attributes in sync. Other title attributes (`<a:bodyPr>` rotation, `sz`, `lang=\"en-US\"`) are preserved exactly; only the `b` attribute changes. Absence (`undefined`) collapses to the application-default `0` (non-bold) so a fresh chart matches Excel's reference serialization.
- **Clone** (`resolveTitleBold`): follows the standard `boolean | null` override grammar — `undefined` inherits, `null` drops, a literal boolean replaces. Non-boolean overrides collapse the same way as the writer's normalizer, so the cloned `SheetChart` always carries a value the writer will accept. The title-presence scope rule applies via the same `titleRendered` gate the writer uses — when the resolved title is dropped (`title: null` or `showTitle: false`), the inherited flag drops silently so the cloned `SheetChart` accurately reflects what the chart will paint.
- **Validation**: only literal `true` / `false` survive the type guard at every layer; non-boolean inputs collapse to the default. This matches how `titleOverlay` / `roundedCorners` / `plotVisOnly` treat their inputs.

## Test plan

- [x] `parseChart — title bold` (10 new tests) — surfaces `true` from `b=\"1\"`, collapses `b=\"0\"` and absence to `undefined`, accepts truthy / falsy spellings (`\"true\"` / `\"false\"`), returns undefined when `<c:title>` is absent / the title is a `<c:strRef>` formula reference / `<a:p>` has no `<a:pPr>`, drops unknown / empty tokens, does not leak from a stray axis-title `<a:defRPr>`, co-surfaces alongside `titleFontSize` / `titleRotation` / `titleOverlay`.
- [x] `writeChart — title bold` (12 new tests) — emits `b=\"0\"` by default on both slots, emits `b=\"1\"` on `true`, emits `b=\"0\"` on `false` (functionally equivalent to omission), drops non-boolean inputs back to the default, ignores the flag when the chart renders no title (`showTitle=false` or absent title), composes independently with `titleFontSize` / `titleRotation` / `titleOverlay`, preserves the title body's other attributes (`lang=\"en-US\"`, `sz`), threads through every chart family (bar / column / line / pie / scatter / area), parse round-trip, default round-trip collapses to `undefined`, explicit `false` round-trip also collapses to `undefined`, full `writeXlsx` package round-trip.
- [x] `cloneChart — title bold` (13 new tests) — inherit, replace (true to false), null drop, replace null source with true override, drop on non-boolean override, drop on `title: null`, drop on `showTitle: false`, preserve when an override pins a fresh title, composition with `titleFontSize` / `titleRotation` / `titleOverlay`, threads through line -> column / line -> doughnut flatten, end-to-end parse -> clone -> write -> reparse round-trip.
- [x] `pnpm test` — all 4395 tests pass (lint + typecheck + vitest).

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1369,6 +1369,33 @@ export interface SheetChart {
    */
   titleFontSize?: number;
   /**
+   * Chart title bold flag. Maps to
+   * `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr b=".."/></a:pPr>
+   * <a:r><a:rPr b=".."/></a:r></a:p></c:rich></c:tx></c:title>` —
+   * Excel's "Format Chart Title -> Font -> Bold" toggle. The OOXML
+   * attribute is the `xsd:boolean` `b` on `CT_TextCharacterProperties`
+   * (ECMA-376 Part 1, §21.1.2.3.7); the writer lands the value on both
+   * the default-paragraph `<a:defRPr>` and the literal run's `<a:rPr>`
+   * so a re-parse picks the flag up off either canonical slot — Excel
+   * keeps the two attributes in sync.
+   *
+   * Default: omitted — the title renders non-bold (`b="0"`, Excel's
+   * reference serialization for a fresh chart title). Set `true` to
+   * emit `b="1"` on both slots so the title renders bold; set `false`
+   * explicitly to pin the non-default `b="0"` (functionally identical
+   * to omission, but useful when overriding a templated title that
+   * had bold pinned upstream).
+   *
+   * Silently ignored when no title is rendered (`showTitle === false`
+   * or `title` is absent) — there is no `<c:title>` block to host the
+   * flag in either case. Composes independently with
+   * {@link titleFontSize} / {@link titleRotation} / {@link titleOverlay}:
+   * all four fields land on the same `<c:title>` element so a single
+   * configuration call threads cleanly through every chart-title knob
+   * Excel exposes.
+   */
+  titleBold?: boolean;
+  /**
    * Auto-title-deleted flag. Maps to `<c:chart><c:autoTitleDeleted
    * val=".."/>` — Excel's record of whether the user explicitly deleted
    * the auto-generated title that single-series charts synthesise from
@@ -3664,6 +3691,27 @@ export interface Chart {
    * {@link SheetChart.titleFontSize} without transformation.
    */
   titleFontSize?: number;
+  /**
+   * Chart title bold flag pulled from
+   * `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr b=".."/></a:pPr>
+   * </a:p></c:rich></c:tx></c:title>`. Reflects Excel's "Format Chart
+   * Title -> Font -> Bold" toggle.
+   *
+   * The OOXML default `false` collapses to `undefined` so absence and
+   * `<a:defRPr b="0"/>` round-trip identically through
+   * {@link cloneChart} — only an explicit `<a:defRPr b="1"/>` surfaces
+   * `true`. The reader accepts the OOXML truthy / falsy spellings
+   * (`"1"` / `"true"` / `"0"` / `"false"`); unknown values and missing
+   * `b` attributes drop to `undefined`.
+   *
+   * Reported as `undefined` whenever the source chart has no
+   * `<c:title>` element at all, or when the title is a `<c:strRef>`
+   * (formula reference) with no `<c:rich>` body — there is no `<a:p>`
+   * to host the flag in either case. The parsed value threads
+   * straight back into the writer-side {@link SheetChart.titleBold}
+   * without transformation.
+   */
+  titleBold?: boolean;
   /**
    * Auto-title-deleted flag pulled from `<c:chart><c:autoTitleDeleted
    * val=".."/>`. Reflects Excel's "the user explicitly deleted the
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed. The implementation cleanly mirrors the existing `titleFontSize` pattern (#251) — same canonical-slot pair (`<a:defRPr>` + `<a:rPr>`), same title-presence scope rule, same `boolean | null` clone grammar — which keeps the chart-title knobs composable at the call site.

Strengths:
- Reader scopes the `<a:defRPr>` lookup tightly through `title -> tx -> rich -> p -> pPr -> defRPr`, so a stray axis-title `<a:defRPr b="1"/>` cannot leak in (covered by the `does not leak from a stray axis-title` test).
- Writer lands the flag on both `<a:defRPr>` and `<a:rPr>` matching Excel's reference serialization, with `undefined` collapsing to the OOXML default `b="0"` so a fresh chart stays byte-identical.
- Symmetric `normalizeTitleBold` helpers in writer and clone defend against typed escapes from untyped callers (covered by `drops a non-boolean override` and `drops non-boolean inputs` tests).
- 35 new tests cover read / write / clone, all chart families, scope-rule gates (`showTitle=false`, `title: null`, `<c:strRef>`, missing `<a:pPr>`), composition with sibling title knobs, and full `writeXlsx` round-trips.

CI green, mergeable, no remaining concerns. Merging.

PR #253, opened 2026-05-06
feat(xlsx): chart title italic flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr i=".."/></a:pPr><a:r><a:rPr i=".."/></a:r></a:p></c:rich></c:tx></c:title>` (CT_TextCharacterProperties' italic flag — ECMA-376 Part 1, §21.1.2.3.7) at the read, write, and clone layers so a template's italic chart title survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

`<a:defRPr i=".."/>` mirrors Excel's "Format Chart Title -> Font -> Italic" toggle. Until now hucre's writer never emitted the attribute (the title always rendered non-italic) and the reader ignored the attribute entirely, so a templated dashboard chart whose user italicized the title for emphasis silently lost the choice on every parse -> clone -> write loop. Bridges another title-customization gap for the dashboard composition flow tracked in #136.

Mirrors the chart-level `titleBold` field added in #252 — same canonical-slot pair (`<a:defRPr>` + `<a:rPr>`), same title-presence scope rule (silently ignored when `showTitle === false` or no literal title), same `boolean | null` clone grammar (`undefined` inherits, `null` drops, a literal boolean replaces) — so a single configuration call threads cleanly through every chart-title knob Excel exposes (bold, italic, size, rotation, overlay) without bookkeeping the canonical OOXML slots.

Refs #136

## API

```ts

// -- Read side --
console.log(source.titleItalic);
// true       <- when the template pinned <a:defRPr i="1"/> on the title's <a:p><a:pPr>
// undefined  <- when the template emits i="0" or omits the attribute (Excel's reference non-italic)
// undefined  <- when the chart has no <c:title> or the title is a <c:strRef> formula reference

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [
      ["Region", "Revenue"],
      ["North", 100],
      ["South", 200],
    ],
    charts: [{
      type: "column",
      title: "Quarterly Revenue",
      // Italicize the title for a hero dashboard tile.
      titleItalic: true,
    }],
  }],
});

// -- Clone-through --
//   -> inherits the source's parsed titleItalic unchanged.
cloneChart(source, {
  titleItalic: null,
});
//   -> drops the inherited italic flag (writer falls back to the OOXML default — no `i` attribute).
cloneChart(source, {
  titleItalic: false,
});
//   -> replaces the inherited flag with explicit non-italic (omits `i` on emit).
```

## Implementation notes

- **Type model**: `SheetChart.titleItalic?: boolean` (writer side); `Chart.titleItalic?: boolean` (reader side). The clone option `titleItalic?: boolean | null` follows the standard `undefined` (inherit) / `null` (drop) / `boolean` (replace) grammar. Same shape as `titleBold` so a parsed value flows straight from the read side back into the write side without transformation.
- **Reader** (`parseTitleItalic`): walks `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr i=".."/></a:pPr></a:p></c:rich></c:tx></c:title>` for the `i` attribute. The lookup is scoped to the title's default-paragraph slot so a stray `<a:defRPr>` elsewhere in the chart (e.g. on an axis title or a data-labels block) cannot leak in. Uses the existing `parseBoolAttr` helper to accept the OOXML truthy / falsy spellings (`"1"` / `"true"` / `"0"` / `"false"`); unknown values and missing `i` attributes drop to `undefined`. The OOXML default `false` collapses to `undefined` so absence and `i="0"` round-trip identically — only an explicit `i="1"` surfaces `true`. Returns `undefined` when the chart has no `<c:title>` element, or when the title is a `<c:strRef>` (formula reference) with no `<c:rich>` body — there is no `<a:p>` slot to host the flag in either case.
- **Writer**: `buildTitle` now resolves through an `italic?: boolean` parameter — `normalizeTitleItalic` accepts only literal booleans and collapses every other token (typed escapes from an untyped caller) to `undefined`. Unlike `titleBold` (which Excel always emits), Excel itself omits the `i` attribute on a non-italic title, so the writer follows suit: only an explicit `true` produces `i="1"` on both the `<a:defRPr>` and the `<a:rPr>`; absence and `false` both omit the attribute. Other title attributes (`<a:bodyPr>` rotation, `sz`, `b`, `lang="en-US"`) are preserved exactly; only the `i` attribute changes.
- **Clone** (`resolveTitleItalic`): follows the standard `boolean | null` override grammar — `undefined` inherits, `null` drops, a literal boolean replaces. Non-boolean overrides collapse the same way as the writer's normalizer, so the cloned `SheetChart` always carries a value the writer will accept. The title-presence scope rule applies via the same `titleRendered` gate the writer uses — when the resolved title is dropped (`title: null` or `showTitle: false`), the inherited flag drops silently so the cloned `SheetChart` accurately reflects what the chart will paint.
- **Validation**: only literal `true` / `false` survive the type guard at every layer; non-boolean inputs collapse to the default. This matches how `titleBold` / `titleOverlay` / `roundedCorners` / `plotVisOnly` treat their inputs.

## Test plan

- [x] `parseChart — title italic` (10 new tests) — surfaces `true` from `i="1"`, collapses `i="0"` and absence to `undefined`, accepts truthy / falsy spellings (`"true"` / `"false"`), returns undefined when `<c:title>` is absent / the title is a `<c:strRef>` formula reference / `<a:p>` has no `<a:pPr>`, drops unknown / empty tokens, does not leak from a stray axis-title `<a:defRPr>`, co-surfaces alongside `titleBold` / `titleFontSize` / `titleRotation` / `titleOverlay`.
- [x] `writeChart — title italic` (12 new tests) — omits `i` by default, emits `i="1"` on `true`, omits `i` on `false` (functionally equivalent to omission), drops non-boolean inputs back to the default, ignores the flag when the chart renders no title (`showTitle=false` or absent title), composes independently with `titleBold` / `titleFontSize` / `titleRotation` / `titleOverlay`, preserves the title body's other attributes (`lang="en-US"`, `sz`, `b`), threads through every chart family (bar / column / line / pie / scatter / area), parse round-trip, default round-trip collapses to `undefined`, explicit `false` round-trip also collapses to `undefined`, full `writeXlsx` package round-trip.
- [x] `cloneChart — title italic` (13 new tests) — inherit, replace (true to false), null drop, replace null source with true override, drop on non-boolean override, drop on `title: null`, drop on `showTitle: false`, preserve when an override pins a fresh title, composition with `titleBold` / `titleFontSize` / `titleRotation` / `titleOverlay`, threads through line -> column / line -> doughnut flatten, end-to-end parse -> clone -> write -> reparse round-trip.
- [x] `pnpm test` — all 4432 tests pass (lint + typecheck + vitest).

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1396,6 +1396,34 @@ export interface SheetChart {
    */
   titleBold?: boolean;
   /**
+   * Chart title italic flag. Maps to
+   * `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr i=".."/></a:pPr>
+   * <a:r><a:rPr i=".."/></a:r></a:p></c:rich></c:tx></c:title>` —
+   * Excel's "Format Chart Title -> Font -> Italic" toggle. The OOXML
+   * attribute is the `xsd:boolean` `i` on `CT_TextCharacterProperties`
+   * (ECMA-376 Part 1, §21.1.2.3.7); the writer lands the value on both
+   * the default-paragraph `<a:defRPr>` and the literal run's `<a:rPr>`
+   * so a re-parse picks the flag up off either canonical slot — Excel
+   * keeps the two attributes in sync.
+   *
+   * Default: omitted — the title renders non-italic (no `i` attribute,
+   * Excel's reference serialization for a fresh chart title; the
+   * application-default `false` collapses to absence). Set `true` to
+   * emit `i="1"` on both slots so the title renders italic; set `false`
+   * explicitly to pin the non-default `i="0"` (functionally identical
+   * to omission, but useful when overriding a templated title that had
+   * italic pinned upstream).
+   *
+   * Silently ignored when no title is rendered (`showTitle === false`
+   * or `title` is absent) — there is no `<c:title>` block to host the
+   * flag in either case. Composes independently with {@link titleBold}
+   * / {@link titleFontSize} / {@link titleRotation} /
+   * {@link titleOverlay}: all five fields land on the same
+   * `<c:title>` element so a single configuration call threads cleanly
+   * through every chart-title knob Excel exposes.
+   */
+  titleItalic?: boolean;
+  /**
    * Auto-title-deleted flag. Maps to `<c:chart><c:autoTitleDeleted
    * val=".."/>` — Excel's record of whether the user explicitly deleted
    * the auto-generated title that single-series charts synthesise from
@@ -3712,6 +3740,27 @@ export interface Chart {
    * without transformation.
    */
   titleBold?: boolean;
+  /**
+   * Chart title italic flag pulled from
+   * `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr i=".."/></a:pPr>
+   * </a:p></c:rich></c:tx></c:title>`. Reflects Excel's "Format Chart
+   * Title -> Font -> Italic" toggle.
+   *
+   * The OOXML default `false` collapses to `undefined` so absence and
+   * `<a:defRPr i="0"/>` round-trip identically through
+   * {@link cloneChart} — only an explicit `<a:defRPr i="1"/>` surfaces
+   * `true`. The reader accepts the OOXML truthy / falsy spellings
+   * (`"1"` / `"true"` / `"0"` / `"false"`); unknown values and missing
+   * `i` attributes drop to `undefined`.
+   *
+   * Reported as `undefined` whenever the source chart has no
+   * `<c:title>` element at all, or when the title is a `<c:strRef>`
+   * (formula reference) with no `<c:rich>` body — there is no `<a:p>`
+   * to host the flag in either case. The parsed value threads
+   * straight back into the writer-side {@link SheetChart.titleItalic}
+   * without transformation.
+   */
+  titleItalic?: boolean;
   /**
    * Auto-title-deleted flag pulled from `<c:chart><c:autoTitleDeleted
    * val=".."/>`. Reflects Excel's "the user explicitly deleted the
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
LGTM. Reviewed the read/write/clone layers and the test coverage:

- `parseTitleItalic` correctly walks the canonical `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr>` chain and bails on every missing link, so a stray axis-title `<a:defRPr>` cannot leak in. The `i="0"` and absence collapse to `undefined` per OOXML default — a sound symmetry with how `titleBold` was wired in #252.
- Writer mirrors Excel's reference serialization: `i` is omitted entirely for non-italic titles (only `true` produces `i="1"`). This is the right asymmetry vs. `titleBold` (Excel always emits `b`) and matches the PR description.
- Clone resolver follows the standard `undefined` (inherit) / `null` (drop) / `boolean` (replace) grammar, with non-boolean overrides collapsing via `normalizeTitleItalic`. Title-presence gating (under `titleRendered`) drops the flag silently when the title is suppressed — consistent with the rest of the title knobs.
- Test coverage is thorough (35 new tests): truthy/falsy spellings, `<c:strRef>`-only title, missing `<a:pPr>`, stray axis-title leak, composition with bold/size/rotation/overlay, cross-family (bar/column/line/pie/scatter/area), and full `parse -> clone -> write -> reparse` round-trip.

CI is green. Merging.

PR #254, opened 2026-05-06
feat(xlsx): chart title font color — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr><a:solidFill><a:srgbClr val="RRGGBB"/></a:solidFill></a:defRPr></a:pPr><a:r><a:rPr><a:solidFill><a:srgbClr val="RRGGBB"/></a:solidFill></a:rPr></a:r></a:p></c:rich></c:tx></c:title>` (CT_SRgbColor inside CT_TextCharacterProperties' fill choice — ECMA-376 Part 1, §20.1.2.3.32 / §21.1.2.3.7) at the read, write, and clone layers so a template's custom title font color survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

Mirrors the chart-level `titleBold` / `titleItalic` fields added in #252 / #253 — same canonical-slot pair (`<a:defRPr>` + `<a:rPr>`), same title-presence scope rule (silently ignored when `showTitle === false` or no literal title), same clone grammar (`undefined` inherits, `null` drops, a literal value replaces) — so a single configuration call threads cleanly through every chart-title knob Excel exposes (color, bold, italic, size, rotation, overlay) without bookkeeping the canonical OOXML slots.

Bridges another title-customization gap for the dashboard composition flow tracked in #136 — the issue's example explicitly calls for "axis labels, colors, and titles", and the title font color is the most visible knob a dashboard hero tile reaches for.

Refs #136

## API

```ts

// -- Read side --
console.log(source.titleColor);
// "1070CA"   <- when the template pinned <a:defRPr><a:solidFill><a:srgbClr val="1070CA"/></a:solidFill></a:defRPr>
// undefined  <- when the template emits no <a:solidFill> (theme text color inherited)
// undefined  <- when the chart has no <c:title> or the title is a <c:strRef> formula reference
// undefined  <- when the fill is a <a:schemeClr> theme reference (only literal sRGB triples round-trip)

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [
      ["Region", "Revenue"],
      ["North", 100],
      ["South", 200],
    ],
    charts: [{
      type: "column",
      title: "Quarterly Revenue",
      // Render the title in the dashboard's hero blue.
      titleColor: "1070CA",
    }],
  }],
});

// -- Clone-through --
//   -> inherits the source's parsed titleColor unchanged.
cloneChart(source, {
  titleColor: null,
});
//   -> drops the inherited fill (writer falls back to the theme text color — no <a:solidFill>).
cloneChart(source, {
  titleColor: "#FF0000",
});
//   -> replaces the inherited fill with the dashboard's red accent (the leading '#' is stripped).
```

## Implementation notes

- **Type model**: `SheetChart.titleColor?: string` (writer side); `Chart.titleColor?: string` (reader side). The clone option `titleColor?: string | null` follows the standard `undefined` (inherit) / `null` (drop) / `string` (replace) grammar. The string format is the 6-character uppercase hex sRGB triple; the writer / reader / clone all accept a leading `#` and a lowercase form on input but normalize to the canonical uppercase form on output so a parsed value flows straight from the read side back into the write side without transformation.
- **Reader** (`parseTitleColor`): walks `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr><a:solidFill><a:srgbClr val="..."/></a:solidFill></a:defRPr></a:pPr></a:p></c:rich></c:tx></c:title>` for the `val` attribute. The lookup is scoped to the title's default-paragraph slot so a stray `<a:solidFill>` elsewhere in the chart (e.g. on an axis title or a data-labels block) cannot leak in. Theme references (`<a:schemeClr>`), `<a:hslClr>`, `<a:sysClr>`, and `<a:prstClr>` all collapse to `undefined` since only the literal RGB triple round-trips losslessly through the writer. Malformed `val` tokens (wrong length, non-hex characters, alpha-channel forms) likewise drop to `undefined` rather than fabricate a value the writer would round-trip into a malformed `<a:srgbClr>`. Returns `undefined` when the chart has no `<c:title>` element, or when the title is a `<c:strRef>` (formula reference) with no `<c:rich>` body — there is no `<a:p>` slot to host the fill in either case.
- **Writer**: `buildTitle` now resolves through an `rgbHex?: string` parameter — `normalizeTitleColor` strips a leading `#`, validates the 6-character hex form, and uppercases the canonical output; every other token (typed escape from an untyped caller — wrong length, non-hex, alpha-channel form, non-string) collapses to `undefined`. When the color is set, the `<a:defRPr>` / `<a:rPr>` slots expand from self-closing to wrapping a single `<a:solidFill><a:srgbClr val="..."/></a:solidFill>` child; otherwise the writer keeps the existing self-closing form so a fresh chart with no custom color matches Excel's reference serialization byte-for-byte. The fill lands on both the default-paragraph `<a:defRPr>` and the literal run's `<a:rPr>` so a re-parse picks the value up off either canonical slot — Excel keeps the two slots in sync.
- **Clone** (`resolveTitleColor`): follows the standard `string | null` override grammar — `undefined` inherits, `null` drops, a literal hex string replaces. Malformed overrides collapse the same way as the writer's normalizer, so the cloned `SheetChart` always carries a value the writer will accept. The title-presence scope rule applies via the same `titleRendered` gate the writer uses — when the resolved title is dropped (`title: null` or `showTitle: false`), the inherited fill drops silently so the cloned `SheetChart` accurately reflects what the chart will paint.
- **Validation**: only valid 6-character hex strings (with or without leading `#`, any case) survive the type guard at every layer; non-string inputs and malformed hex collapse to the default. This matches how `titleBold` / `titleItalic` / `titleOverlay` / `roundedCorners` / `plotVisOnly` treat their inputs (a literal valid value is the only path to a non-default emission).

## Test plan

- [x] `parseChart — title color` (10 new tests) — surfaces uppercase canonical hex from `<a:srgbClr val="FF0000"/>`, normalizes lowercase input to uppercase, strips leading `#`, collapses absence of `<a:solidFill>` to `undefined`, returns undefined when `<c:title>` is absent / the title is a `<c:strRef>` formula reference / `<a:p>` has no `<a:pPr>`, collapses theme-color references (`<a:schemeClr>`) to undefined, drops malformed val tokens (wrong length / non-hex / 8-char alpha form), does not leak from a stray axis-title `<a:solidFill>`, co-surfaces alongside `titleBold` / `titleItalic` / `titleFontSize` / `titleRotation` / `titleOverlay`.
- [x] `writeChart — title color` (12 new tests) — omits `<a:solidFill>` by default (theme inherit), emits the fill on a hex input, normalizes lowercase to uppercase, strips leading `#`, drops malformed inputs back to the default, ignores the field when the chart renders no title (`showTitle=false` or absent title), composes independently with `titleBold` / `titleItalic` / `titleFontSize` / `titleRotation` / `titleOverlay`, preserves the title body's other attributes (`lang="en-US"`, `sz`, `b`), threads through every chart family (bar / column / line / pie / scatter / area), parse round-trip, default round-trip collapses to `undefined`, malformed-input round-trip also collapses to `undefined`, full `writeXlsx` package round-trip.
- [x] `cloneChart — title color` (16 new tests) — inherit, replace, lowercase normalization, leading-`#` strip, null drop, replace null source with hex override, drop on malformed override, drop on non-string override, drop on `title: null`, drop on `showTitle: false`, preserve when an override pins a fresh title, composition with `titleBold` / `titleItalic` / `titleFontSize` / `titleRotation` / `titleOverlay`, threads through line -> column / line -> doughnut flatten, end-to-end parse -> clone -> write -> reparse round-trip.
- [x] `pnpm test` — all 4473 tests pass (lint + typecheck + vitest).

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1424,6 +1424,45 @@ export interface SheetChart {
    */
   titleItalic?: boolean;
   /**
+   * Chart title font color. Maps to
+   * `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr><a:solidFill>
+   * <a:srgbClr val="RRGGBB"/></a:solidFill></a:defRPr></a:pPr>
+   * <a:r><a:rPr><a:solidFill><a:srgbClr val="RRGGBB"/></a:solidFill>
+   * </a:rPr></a:r></a:p></c:rich></c:tx></c:title>` — Excel's
+   * "Format Chart Title -> Font -> Font Color" picker. The OOXML
+   * `<a:srgbClr val=".."/>` carries the 6-character uppercase hex
+   * sRGB color (`CT_SRgbColor`, ECMA-376 Part 1, §20.1.2.3.32); the
+   * writer lands the fill on both the default-paragraph `<a:defRPr>`
+   * and the literal run's `<a:rPr>` so a re-parse picks the color up
+   * off either canonical slot — Excel keeps the two values in sync.
+   *
+   * Accepts the standard 6-character hex string with or without a
+   * leading `#` (`"FF0000"` / `"#FF0000"` / `"ff0000"`); the writer
+   * normalizes to the OOXML uppercase canonical form
+   * (`<a:srgbClr val="FF0000"/>`). The 8-character `#RRGGBBAA` form
+   * is *not* accepted — alpha lives on `<a:srgbClr><a:alpha val=".."/>`
+   * which is a separate runs-level knob; pinning `titleColor` carries
+   * the RGB triple only.
+   *
+   * Default: omitted — the title renders in Excel's reference
+   * inherited theme color (no `<a:solidFill>` element, the writer
+   * skips the fill block entirely). Pin a hex value to render the
+   * title in that color (e.g. `"1070CA"` for the dashboard hero blue
+   * the issue-#136 example reaches for). Malformed inputs (wrong
+   * length, non-hex characters, alpha-channel form) collapse to
+   * `undefined` so a stray non-hex token never produces a malformed
+   * `<a:srgbClr>`.
+   *
+   * Silently ignored when no title is rendered (`showTitle === false`
+   * or `title` is absent) — there is no `<c:title>` block to host the
+   * fill in either case. Composes independently with {@link titleBold}
+   * / {@link titleItalic} / {@link titleFontSize} /
+   * {@link titleRotation} / {@link titleOverlay}: all six fields land
+   * on the same `<c:title>` element so a single configuration call
+   * threads cleanly through every chart-title knob Excel exposes.
+   */
+  titleColor?: string;
+  /**
    * Auto-title-deleted flag. Maps to `<c:chart><c:autoTitleDeleted
    * val=".."/>` — Excel's record of whether the user explicitly deleted
    * the auto-generated title that single-series charts synthesise from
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed. The change is consistent with the chart-title knob pattern established by #252 (titleBold) and #253 (titleItalic): a per-file normalizer/resolver triple, the same `undefined`/`null`/value clone grammar, and the same `titleRendered` gate. Surface-level checks I verified:

- The reader walks the canonical `<c:title>/<c:tx>/<c:rich>/<a:p>/<a:pPr>/<a:defRPr>/<a:solidFill>/<a:srgbClr>` chain and bails on the first missing link, matching the schema scope so a stray axis-title fill cannot leak in (covered by the dedicated test).
- The writer lands the fill on both `<a:defRPr>` and `<a:rPr>` so re-parse picks it up off either canonical slot, and self-closing form is preserved when no color is set (byte-for-byte parity with Excel's reference for default).
- All four normalizers (reader `normalizeRgbHex`, writer `normalizeTitleColor`, clone `normalizeTitleColor`) share the same `#`-strip, length-check, hex regex, uppercase pipeline. Slight DRY duplication, but it matches the codebase convention of one normalizer per layer.
- Edge cases covered: `<c:strRef>` formula title, `<a:schemeClr>`/`<a:hslClr>`/`<a:sysClr>`/`<a:prstClr>` theme refs, 8-char alpha form, missing `<a:pPr>`, malformed val, non-string input via untyped escape, `title: null` / `showTitle: false` drop-through.

CI is green (4473 tests passing, build succeeds). Merging.

PR #255, opened 2026-05-06
feat(xlsx): chart axis title font size — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:catAx><c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr sz=\"N\"/></a:pPr><a:r><a:rPr sz=\"N\"/></a:r></a:p></c:rich></c:tx></c:title></c:catAx>` (and the matching `<c:valAx>` shape — CT_Title's default-paragraph and run text-run-properties on every axis flavour, ECMA-376 Part 1, §21.2.2.4 / §21.1.2.3.7) at the read, write, and clone layers so a template's custom axis-title font size survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

Mirrors the chart-level `titleFontSize` field added in #251 — same `1..400`pt band the OOXML `ST_TextFontSize` schema exposes, same 0.5pt half-step granularity Excel's UI exposes, same out-of-range / non-finite drop semantics, same title-presence scope rule (silently ignored when the axis renders no title) — so a single configuration call threads cleanly through both the chart title and either axis title without bookkeeping the units. Mirrors `axisTitleRotation` (#248) for the same OOXML slot-pair, so `axisTitleFontSize` and `axisTitleRotation` compose freely on the same `<c:title>` body.

Until now hucre's writer hardcoded `sz=\"1000\"` (10pt) on both axis titles' `<a:defRPr>` and `<a:rPr>` and the reader ignored the attribute entirely, so a templated dashboard chart whose user scaled an axis title up for a hero tile or down to fit a tight axis-label slot silently lost the choice on every parse -> clone -> write loop. Bridges another title-customization gap for the dashboard composition flow tracked in #136 — the issue's example explicitly calls for axis-label customization and the title font size is the last knob the chart-title family had that the axis-title family didn't.

Refs #136

## API

\`\`\`ts

// -- Read side --
console.log(source.axes?.x?.axisTitleFontSize);
// 14         <- when the template pinned <a:defRPr sz=\"1400\"/>
// 10         <- the writer's reference 10pt default round-trips literally
// undefined  <- when the chart has no <c:title> on the X axis or the title is a <c:strRef>
// undefined  <- when the size is out-of-range / non-numeric (corrupt template)

// -- Write side --
await writeXlsx({
  sheets: [{
    name: \"Dashboard\",
    rows: [/* ... */],
    charts: [{
      type: \"column\",
      title: \"Q4 Revenue\",
      series: [/* ... */],
      axes: {
        x: { title: \"Period\", axisTitleFontSize: 12 },
        y: { title: \"USD\", axisTitleFontSize: 16 },
      },
    }],
  }],
});

// -- Clone-through --
  axes: {
    x: { axisTitleFontSize: 14 },     // replaces the inherited size
    y: { axisTitleFontSize: null },   // drops it (writer falls back to 10pt)
  },
});
\`\`\`

## Test plan

- [x] \`pnpm test\` (lint + format + typecheck + 4527 vitest tests; 54 new tests covering parser / writer / clone for the new field) — all pass
- [x] \`pnpm build\` — completes cleanly

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1923,6 +1923,39 @@ export interface SheetChart {
        * dropped on those families.
        */
       axisTitleRotation?: number;
+      /**
+       * Axis title font size in whole or half points. Maps to
+       * `<c:catAx><c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr sz="N"/></a:pPr>
+       * <a:r><a:rPr sz="N"/></a:r></a:p></c:rich></c:tx></c:title></c:catAx>`
+       * (or `<c:valAx>` for scatter / value axes) — Excel's "Format Axis
+       * Title -> Font -> Size" knob. The OOXML attribute is in 100ths
+       * of a point, so 12pt serializes as `sz="1200"` and 10pt (Excel's
+       * reference default for axis titles) as `sz="1000"`; the writer
+       * performs the conversion at emit time and lands the value on
+       * both the default-paragraph `<a:defRPr>` and the literal run's
+       * `<a:rPr>` so a re-parse picks the size up off either canonical
+       * slot.
+       *
+       * Mirrors {@link SheetChart.titleFontSize} for axis titles — same
+       * `1..400`pt band the OOXML `ST_TextFontSize` schema exposes,
+       * same 0.5pt half-step granularity Excel's UI exposes, same
+       * out-of-range / non-finite drop semantics. Useful for shrinking
+       * a long Y-axis unit label so it fits a tight chart frame, or
+       * bumping the X-axis title up to match a presentation slide's
+       * typography.
+       *
+       * Default: omitted — the axis title renders at Excel's reference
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed end-to-end. The implementation cleanly mirrors the existing `axisTitleRotation` and the chart-level `titleFontSize` patterns: reader scopes the lookup to the title's `<c:rich>` body (no leakage from `<c:txPr>` tick-label `<a:defRPr>`), writer lands the `sz` on both `<a:defRPr>` and `<a:rPr>` and falls back to the 1000 (10pt) Excel reference when absent, clone helper follows the standard `undefined` (inherit) / `null` (drop) / value (replace) grammar with title-presence gating. Out-of-range, non-finite, non-numeric, half-step rounding, and `<c:strRef>` formula-title cases all have explicit tests. CI green; merging.

PR #256, opened 2026-05-06
feat(xlsx): chart axis title bold flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:catAx><c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr b=\"..\"/></a:pPr><a:r><a:rPr b=\"..\"/></a:r></a:p></c:rich></c:tx></c:title></c:catAx>` (and the matching `<c:valAx>` shape — CT_Title's default-paragraph and run text-character-properties bold flag on every axis flavour, ECMA-376 Part 1, §21.1.2.3.7) at the read, write, and clone layers so a template's bold axis-title pin survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

Mirrors the chart-level `titleBold` field added in #252 — same canonical-slot pair (`<a:defRPr>` + `<a:rPr>`), same drop-on-default-`false` semantics, same `boolean | null` clone grammar (`undefined` inherits, `null` drops, a literal boolean replaces) — so a single configuration call threads cleanly through both the chart title and either axis title without bookkeeping the canonical OOXML slots. Mirrors `axisTitleRotation` (#248) and `axisTitleFontSize` (#255) for the same OOXML slot-pair, so all three axis-title knobs compose freely on the same `<c:title>` body.

Until now hucre's writer hardcoded `b=\"0\"` (non-bold) on both axis titles' `<a:defRPr>` and `<a:rPr>` and the reader ignored the attribute entirely, so a templated dashboard chart whose user bolded an axis title for emphasis silently lost the choice on every parse -> clone -> write loop. Bridges another title-customization gap for the dashboard composition flow tracked in #136 — the issue's example explicitly calls for axis-label customization and the bold flag is one of the most visible knobs the chart-title family had that the axis-title family didn't.

Refs #136

## API

```ts

// -- Read side --
console.log(source.axes?.x?.axisTitleBold);
// true       <- when the template pinned <a:defRPr b="1"/>
// undefined  <- when the template emits <a:defRPr b="0"/> (OOXML default round-trips identically)
// undefined  <- when the chart has no <c:title> on the X axis or the title is a <c:strRef>

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [/* ... */],
    charts: [{
      type: "column",
      title: "Q4 Revenue",
      series: [/* ... */],
      axes: {
        x: { title: "Period", axisTitleBold: false },
        y: { title: "USD", axisTitleBold: true },
      },
    }],
  }],
});

// -- Clone-through --
  axes: {
    x: { axisTitleBold: true },     // replaces the inherited flag
    y: { axisTitleBold: null },     // drops it (writer falls back to b="0")
  },
});
```

## Implementation notes

- **Type model**: `SheetChart.axes.{x,y}.axisTitleBold?: boolean` (writer side); `ChartAxisInfo.axisTitleBold?: boolean` (reader side). The clone option `axisTitleBold?: boolean | null` follows the standard `undefined` (inherit) / `null` (drop) / `boolean` (replace) grammar.
- **Reader** (`parseAxisTitleBold`): walks `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr b=\"..\"/></a:pPr></a:p></c:rich></c:tx></c:title>` for the `b` attribute. The lookup is scoped to the axis title's `<c:rich>` body so a stray `<a:defRPr>` elsewhere on the axis (e.g. on the tick-label `<c:txPr>`) cannot leak in. The OOXML default `false` collapses to `undefined` so absence and `b=\"0\"` round-trip identically through `cloneChart` — only an explicit `b=\"1\"` surfaces `true`. Returns `undefined` when the axis omits `<c:title>` or when the title is a `<c:strRef>` (formula reference) with no `<c:rich>` body.
- **Writer**: `buildAxisTitle` now resolves through a `bold?: boolean` parameter — `normalizeAxisTitleBold` accepts `true` / `false` literally and collapses every other token (typed escape from an untyped caller) to `undefined`. Absence falls back to the OOXML default `b=\"0\"` (non-bold) so a fresh chart matches Excel's reference serialization byte-for-byte. The flag lands on both the default-paragraph `<a:defRPr>` and the literal run's `<a:rPr>` so a re-parse picks the value up off either canonical slot — Excel keeps the two attributes in sync.
- **Clone** (`applyAxisTitleBoldOverride`): follows the standard `boolean | null` override grammar — `undefined` inherits, `null` drops, a literal boolean replaces. Non-boolean overrides collapse the same way as the writer's normalizer, so the cloned `SheetChart` always carries a value the writer will accept. The title-presence scope rule applies via the same gate the writer uses — when the resolved title is dropped (`title: null` or `showTitle: false`), the inherited flag drops silently so the cloned `SheetChart` accurately reflects what the chart will paint.

## Test plan

- [x] `parseChart — axis title bold` (12 new tests) — surfaces `true` from `b=\"1\"`, collapses `b=\"0\"` and absence to `undefined`, accepts the OOXML truthy / falsy spellings, drops unknown tokens, returns undefined when `<c:title>` is absent / the title is a `<c:strRef>` formula reference / `<a:p>` has no `<a:pPr>`, surfaces on both X and Y axes including scatter's value-axis pair, does not leak from a stray chart-level `<a:defRPr>`, co-surfaces alongside `axisTitleRotation` on the same axis title.
- [x] `writeChart — axis title bold` (10 new tests) — emits `b=\"0\"` on both `<a:defRPr>` and `<a:rPr>` by default, emits `b=\"1\"` on `axisTitleBold=true`, emits `b=\"0\"` on `axisTitleBold=false` (functionally identical to omission), emits independently per axis, drops non-boolean inputs to the default, silently ignores when no axis title renders, composes with `axisTitleRotation`, threads through every chart family that has axes (bar / column / line / area / scatter), full `writeXlsx` package round-trip.
- [x] `cloneChart — axis title bold` (10 new tests) — inherits the source's value, inherits independently per axis, override replaces, `null` drops, replace null source with override, returns undefined with no source / no override, drops non-boolean override, drops on `title: null`, composes with `axisTitleRotation`, end-to-end parse -> clone -> write -> reparse round-trip.
- [x] `pnpm test` — all 4506 tests pass (lint + format + typecheck + vitest).

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1923,6 +1923,47 @@ export interface SheetChart {
        * dropped on those families.
        */
       axisTitleRotation?: number;
+      /**
+       * Axis title bold flag. Maps to
+       * `<c:catAx><c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr b=".."/></a:pPr>
+       * <a:r><a:rPr b=".."/></a:r></a:p></c:rich></c:tx></c:title></c:catAx>`
+       * (or `<c:valAx>` for scatter / value axes) — Excel's "Format
+       * Axis Title -> Font -> Bold" toggle. The OOXML attribute is the
+       * `xsd:boolean` `b` on `CT_TextCharacterProperties` (ECMA-376
+       * Part 1, §21.1.2.3.7); the writer lands the value on both the
+       * default-paragraph `<a:defRPr>` and the literal run's `<a:rPr>`
+       * so a re-parse picks the flag up off either canonical slot —
+       * Excel keeps the two attributes in sync.
+       *
+       * Mirrors {@link SheetChart.titleBold} for axis titles — same
+       * canonical-slot pair, same `boolean | null` clone grammar, same
+       * silent drop when the axis renders no title. Useful for matching
+       * the chart-level title's emphasis on the axis labels (e.g.
+       * bolding the Y-axis "Revenue (USD)" caption to align with a
+       * dashboard's heavy-weight typography).
+       *
+       * Default: omitted — the axis title renders non-bold (`b="0"`,
+       * the writer's reference serialization for a fresh axis title).
+       * Set `true` to emit `b="1"` on both slots so the axis title
+       * renders bold; set `false` explicitly to pin the non-default
+       * `b="0"` (functionally identical to omission, but useful when
+       * overriding a templated axis title that had bold pinned
+       * upstream).
+       *
+       * Silently dropped when the axis renders no title (the
+       * `<c:title>` element is absent in either case) and on `pie` /
+       * `doughnut` charts (no axes at all). Composes independently
+       * with {@link axisTitleRotation} / {@link axisTitleFontSize}: all
+       * three fields land on the same `<c:title>` body so a single
+       * configuration call threads cleanly through every axis-title
+       * knob Excel exposes.
+       *
+       * Sits on every axis flavour — `<c:catAx>` / `<c:valAx>` /
+       * `<c:dateAx>` / `<c:serAx>` all carry the same `<c:title>` shape
+       * per the OOXML schema, so the field round-trips on every chart
+       * family that has axes (bar / column / line / area / scatter).
+       */
+      axisTitleBold?: boolean;
       gridlines?: ChartAxisGridlines;
       scale?: ChartAxisScale;
       numberFormat?: ChartAxisNumberFormat;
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed end-to-end. The implementation cleanly mirrors the chart-level `titleBold` and the existing `axisTitleRotation` patterns: reader scopes the lookup to the title's `<c:rich>` body, writer lands the `b` on both `<a:defRPr>` and `<a:rPr>` and falls back to the OOXML default `b="0"` when absent, OOXML default `false` collapses to `undefined` so absence and `b="0"` round-trip identically, and the clone helper follows the standard `undefined` (inherit) / `null` (drop) / boolean (replace) grammar with title-presence gating. `<c:strRef>` formula-title and stray chart-level `<a:defRPr>` cases are explicitly tested.

Resolved a merge conflict with #255 (axisTitleFontSize) — both fields land on the same `<c:title>` body and `buildAxisTitle` had to absorb both the `fontSizePt` and `bold` parameters. After the merge `pnpm test` (4560 tests) and `pnpm build` pass locally; pending CI before squash-merge.

PR #257, opened 2026-05-06
feat(xlsx): chart axis title italic flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:catAx><c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr i=".."/></a:pPr><a:r><a:rPr i=".."/></a:r></a:p></c:rich></c:tx></c:title></c:catAx>` (and the matching `<c:valAx>` shape — CT_TextCharacterProperties italic flag on every axis flavour, ECMA-376 Part 1, §21.1.2.3.7) at the read, write, and clone layers so a template's italic axis-title pin survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

Mirrors the chart-level `titleItalic` field added in #253 — same canonical-slot pair (`<a:defRPr>` + `<a:rPr>`), same drop-on-default-`false` / drop-on-non-boolean semantics, same `boolean | null` clone grammar (`undefined` inherits, `null` drops, a literal boolean replaces) — so a single configuration call threads cleanly through both the chart title and either axis title without bookkeeping the canonical OOXML slots. Mirrors `axisTitleRotation` (#248) and `axisTitleFontSize` (#255) for the same OOXML slot-pair, so all four axis-title knobs (rotation, size, bold (#256), italic) compose freely on the same `<c:title>` body.

Until now hucre's writer hardcoded a non-italic axis title (omitted `i` attribute) and the reader ignored the attribute entirely, so a templated dashboard chart whose user italicised an axis title for emphasis silently lost the choice on every parse -> clone -> write loop. Bridges another title-customization gap for the dashboard composition flow tracked in #136.

Refs #136

## API

```ts

// -- Read side --
console.log(source.axes?.x?.axisTitleItalic);
// true       <- when the template pinned <a:defRPr i="1"/>
// undefined  <- when the template emits <a:defRPr i="0"/> or omits the attribute

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [/* ... */],
    charts: [{
      type: "column",
      title: "Q4 Revenue",
      series: [/* ... */],
      axes: {
        x: { title: "Period", axisTitleItalic: false },
        y: { title: "USD", axisTitleItalic: true },
      },
    }],
  }],
});

// -- Clone-through --
  axes: {
    x: { axisTitleItalic: true },     // replaces the inherited flag
    y: { axisTitleItalic: null },     // drops it (writer omits the attribute)
  },
});
```

## Implementation notes

- **Type model**: `SheetChart.axes.{x,y}.axisTitleItalic?: boolean` (writer side); `ChartAxisInfo.axisTitleItalic?: boolean` (reader side). The clone option `axisTitleItalic?: boolean | null` follows the standard `undefined` (inherit) / `null` (drop) / `boolean` (replace) grammar.
- **Reader** (`parseAxisTitleItalic`): walks `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr i=".."/></a:pPr></a:p></c:rich></c:tx></c:title>` for the `i` attribute. The lookup is scoped to the axis title's `<c:rich>` body so a stray `<a:defRPr>` elsewhere on the axis (e.g. on the tick-label `<c:txPr>`) cannot leak in. The OOXML default `false` collapses to `undefined` so absence and `i="0"` round-trip identically through `cloneChart` — only an explicit `i="1"` surfaces `true`. Returns `undefined` when the axis omits `<c:title>` or when the title is a `<c:strRef>` (formula reference) with no `<c:rich>` body.
- **Writer**: `buildAxisTitle` now resolves through an `italic?: boolean` parameter — `normalizeAxisTitleItalic` accepts `true` / `false` literally and collapses every other token (typed escape from an untyped caller) to `undefined`. Absence and explicit `false` both collapse to omitting the `i` attribute (Excel's reference serialization for a non-italic axis title — only the bold flag is always emitted). Only `true` emits `i="1"` on both the default-paragraph `<a:defRPr>` and the literal run's `<a:rPr>` so a re-parse picks the value up off either canonical slot — Excel keeps the two attributes in sync.
- **Clone** (`applyAxisTitleItalicOverride`): follows the standard `boolean | null` override grammar — `undefined` inherits, `null` drops, a literal boolean replaces. Non-boolean overrides collapse the same way as the writer's normalizer, so the cloned `SheetChart` always carries a value the writer will accept. The title-presence scope rule applies via the same gate the writer uses — when the resolved title is dropped (`title: null` or override), the inherited flag drops silently so the cloned `SheetChart` accurately reflects what the chart will paint.

## Test plan

- [x] `parseChart — axis title italic` (12 new tests) — surfaces `true` from `i="1"`, collapses `i="0"` and absence to `undefined`, accepts the OOXML truthy / falsy spellings, drops unknown tokens, returns undefined when `<c:title>` is absent / the title is a `<c:strRef>` formula reference / `<a:p>` has no `<a:pPr>`, surfaces on both X and Y axes, does not leak from a stray tick-label `<c:txPr>` italic flag, co-surfaces alongside `axisTitleRotation` / `axisTitleFontSize` on the same axis title.
- [x] `writeChart — axis title italic` (12 new tests) — omits the `i` attribute by default, emits `i="1"` on both `<a:defRPr>` and `<a:rPr>` on `axisTitleItalic=true`, omits the attribute on `axisTitleItalic=false` (matches OOXML default), emits independently per axis, drops non-boolean inputs to omission, silently ignores when no axis title renders, composes with `axisTitleRotation` / `axisTitleFontSize`, threads through every chart family that has axes (line / area / scatter), preserves the surrounding `b="0"` / `lang="en-US"` / `sz="1000"` defaults, full `writeXlsx` package round-trip.
- [x] `cloneChart — axis title italic` (15 new tests) — inherits the source's value, inherits independently per axis, override replaces, `null` drops, replace null source with override, returns undefined with no source / no override, drops non-boolean override, normalises a parsed source flag that is somehow non-boolean, drops on `title: null`, drops an override when no axis title resolves, preserves override when override pins a title, composes with chart-level `titleItalic` / `axisTitleRotation` / `axisTitleFontSize`, end-to-end parse -> clone -> write -> reparse round-trip, full `writeXlsx` propagation.
- [x] `pnpm test` — all 4568 tests pass (lint + format + typecheck + vitest).

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1956,6 +1956,42 @@ export interface SheetChart {
        * family that has axes (bar / column / line / area / scatter).
        */
       axisTitleFontSize?: number;
+      /**
+       * Axis title italic flag. Maps to
+       * `<c:catAx><c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr i=".."/></a:pPr>
+       * <a:r><a:rPr i=".."/></a:r></a:p></c:rich></c:tx></c:title></c:catAx>`
+       * (or `<c:valAx>` for scatter / value axes) — Excel's "Format Axis
+       * Title -> Font -> Italic" toggle. The OOXML attribute is the
+       * `xsd:boolean` `i` on `CT_TextCharacterProperties` (ECMA-376
+       * Part 1, §21.1.2.3.7); the writer lands the value on both the
+       * default-paragraph `<a:defRPr>` and the literal run's `<a:rPr>`
+       * so a re-parse picks the flag up off either canonical slot —
+       * Excel keeps the two attributes in sync.
+       *
+       * Mirrors {@link SheetChart.titleItalic} for axis titles — same
+       * canonical-slot pair, same drop-on-default semantics, same
+       * `boolean | null` clone grammar so a single configuration call
+       * threads cleanly through both the chart title and either axis
+       * title without bookkeeping the canonical OOXML slots.
+       *
+       * Default: omitted — the axis title renders non-italic (no `i`
+       * attribute, Excel's reference serialization for a fresh axis
+       * title; the application-default `false` collapses to absence).
+       * Set `true` to emit `i="1"` on both slots so the title renders
+       * italic; set `false` explicitly to pin the non-default `i="0"`
+       * (functionally identical to omission, but useful when overriding
+       * a templated axis title that had italic pinned upstream).
+       *
+       * Silently dropped when the axis renders no title (the
+       * `<c:title>` element is absent in either case) and on `pie` /
+       * `doughnut` charts (no axes at all).
+       *
+       * Sits on every axis flavour — `<c:catAx>` / `<c:valAx>` /
+       * `<c:dateAx>` / `<c:serAx>` all carry the same `<c:title>` shape
+       * per the OOXML schema, so the field round-trips on every chart
+       * family that has axes (bar / column / line / area / scatter).
+       */
+      axisTitleItalic?: boolean;
       gridlines?: ChartAxisGridlines;
       scale?: ChartAxisScale;
       numberFormat?: ChartAxisNumberFormat;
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
## Review

The italic flag mirrors the freshly merged axisTitleBold (#256) shape exactly: same canonical-slot pair (`<a:defRPr i=".."/>` and `<a:rPr i=".."/>`), same boolean | null clone grammar, same drop-on-default-`false` semantics, and the same `<c:title>` / `<c:rich>` body scope so the tick-label `<c:txPr>` italic flag cannot leak in. Read / write / clone are symmetric; out-of-band tokens collapse to undefined on every layer; the title-presence gate (`xTitle === undefined`) drops a stray inherited flag when the resolved axis carries no title. Test coverage is solid (39 new tests across reader/writer/clone) and exercises both X and Y, both catAx / valAx, the strRef formula path, the missing pPr branch, the tick-label leak guard, composition with rotation/fontSize/chart-titleItalic, and a full writeXlsx round-trip.

One nit worth flagging: `applyAxisTitleItalicOverride` routes both source and override through `normalizeTitleItalic`, which collapses a literal `false` to `undefined`. The matching bold helper `applyAxisTitleBoldOverride` (#256) preserves a literal `false`, so an inherited `axisTitleItalic: false` does not round-trip identically across `cloneChart` even though a literal `false` is a legal serialization (`i="0"`). Functionally harmless because the writer also drops `false` to absence (Excel's reference is `i` omitted), but the asymmetry with bold is a small surprise — easy follow-up if desired, not blocking.

## Fixes applied in this run

- Resolved CONFLICTING with main: merged #256 (axisTitleBold) cleanly so both flags coexist.
- Threaded both `bold` and `italic` parameters through `buildAxisTitle` and the four call sites (bar/column, line/area routed through column, scatter X, scatter Y); both flags now land on `<a:defRPr>` and `<a:rPr>` in the same emit path.
- Added `axisTitleBold` siblings everywhere `axisTitleItalic` was added (resolver helpers, render-options interface, type definitions for `SheetChart.axes.{x,y}`, `ChartAxisInfo`, `CloneChartOptions.axes.{x,y}`).
- Concatenated the new italic test suites after the bold suites in `chart-clone.test.ts`, `charts-write.test.ts`, and `charts.test.ts` (both blocks share identical helpers).

## Verified locally

- `pnpm test` (lint + format + typecheck + vitest): 4601 tests pass on 89 files.
- `pnpm build`: obuild succeeds, dist clean.

CI pending — will merge once green.

PR #258, opened 2026-05-06
feat(xlsx): chart axis title font color — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:catAx><c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr><a:solidFill><a:srgbClr val=\"RRGGBB\"/></a:solidFill></a:defRPr></a:pPr><a:r><a:rPr><a:solidFill><a:srgbClr val=\"RRGGBB\"/></a:solidFill></a:rPr></a:r></a:p></c:rich></c:tx></c:title></c:catAx>` (and the matching `<c:valAx>` shape — CT_SRgbColor inside CT_TextCharacterProperties' fill choice on every axis flavour, ECMA-376 Part 1, §20.1.2.3.32 / §21.1.2.3.7) at the read, write, and clone layers so a template's custom axis-title font color survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

Mirrors the chart-level `titleColor` field added in #254 — same canonical-slot pair (`<a:defRPr>` + `<a:rPr>`), same accept-with-or-without-`#` grammar (`"FF0000"` / `"#FF0000"` / `"ff0000"` all collapse to the OOXML uppercase canonical form), same `string | null` clone grammar (`undefined` inherits, `null` drops, a hex string replaces) — so a single configuration call threads cleanly through both the chart title and either axis title without bookkeeping the canonical OOXML slots. Mirrors `axisTitleRotation` (#248), `axisTitleFontSize` (#255), `axisTitleBold` (#256), and `axisTitleItalic` (#257) for the same `<c:title>` body, so all five axis-title knobs (rotation, size, bold, italic, color) compose freely on the same axis.

Until now hucre's writer hardcoded a theme-color axis title (omitted the `<a:solidFill>` block) and the reader ignored the fill entirely, so a templated dashboard chart whose user tinted an axis title to match an accent color silently lost the choice on every parse -> clone -> write loop. Bridges another title-customization gap for the dashboard composition flow tracked in #136 — the issue's example explicitly calls for \"axis labels, colors, and titles\".

Refs #136

## API

\`\`\`ts

// -- Read side --
console.log(source.axes?.x?.axisTitleColor);
// \"1070CA\"   <- when the template pinned <a:srgbClr val=\"1070CA\"/>
// undefined  <- when the fill is absent (theme text color inherited)
// undefined  <- when the fill is a <a:schemeClr> theme reference
// undefined  <- when the axis omits <c:title> or the title is a <c:strRef>

// -- Write side --
await writeXlsx({
  sheets: [{
    name: \"Dashboard\",
    rows: [/* ... */],
    charts: [{
      type: \"column\",
      title: \"Q4 Revenue\",
      series: [/* ... */],
      axes: {
        // Render the X-axis title in the dashboard's hero blue.
        x: { title: \"Period\", axisTitleColor: \"1070CA\" },
        // Render the Y-axis title in the revenue accent green.
        y: { title: \"USD\", axisTitleColor: \"#00C586\" },
      },
    }],
  }],
});

// -- Clone-through --
  axes: {
    x: { axisTitleColor: \"FF0000\" }, // replaces the inherited fill
    y: { axisTitleColor: null },     // drops it (writer omits <a:solidFill>)
  },
});
\`\`\`

## Implementation notes

- **Type model**: `SheetChart.axes.{x,y}.axisTitleColor?: string` (writer side); `ChartAxisInfo.axisTitleColor?: string` (reader side). The clone option `axisTitleColor?: string | null` follows the standard `undefined` (inherit) / `null` (drop) / `string` (replace) grammar. The string format is the 6-character uppercase hex sRGB triple; the writer / reader / clone all accept a leading `#` and a lowercase form on input but normalize to the canonical uppercase form on output so a parsed value flows straight from the read side back into the write side without transformation.
- **Reader** (`parseAxisTitleColor`): walks `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr><a:solidFill><a:srgbClr val=\"...\"/></a:solidFill></a:defRPr></a:pPr></a:p></c:rich></c:tx></c:title>` for the `val` attribute. The lookup is scoped to the axis title's `<c:rich>` body so a stray `<a:solidFill>` elsewhere on the axis (e.g. on the tick-label `<c:txPr>` or a `<c:spPr>` series fill) cannot leak in. Theme references (`<a:schemeClr>`), `<a:hslClr>`, `<a:sysClr>`, and `<a:prstClr>` all collapse to `undefined` since only the literal RGB triple round-trips losslessly through the writer. Malformed `val` tokens (wrong length, non-hex characters, alpha-channel forms) likewise drop to `undefined` rather than fabricate a value the writer would round-trip into a malformed `<a:srgbClr>`. Returns `undefined` when the axis omits `<c:title>` or when the title is a `<c:strRef>` (formula reference) with no `<c:rich>` body. Reuses the `normalizeRgbHex` helper introduced for the chart-level `titleColor` (#254).
- **Writer**: `buildAxisTitle` now resolves through an `rgbHex?: string` parameter — `normalizeAxisTitleColor` strips a leading `#`, validates the 6-character hex form, and uppercases the canonical output via the chart-level `normalizeTitleColor`; every other token (typed escape from an untyped caller — wrong length, non-hex, alpha-channel form, non-string) collapses to `undefined`. When the color is set, the `<a:defRPr>` / `<a:rPr>` slots expand from self-closing to wrapping a single `<a:solidFill><a:srgbClr val=\"...\"/></a:solidFill>` child; otherwise the writer keeps the existing self-closing form so a fresh axis title with no custom color matches Excel's reference serialization byte-for-byte. The fill lands on both the default-paragraph `<a:defRPr>` and the literal run's `<a:rPr>` so a re-parse picks the value up off either canonical slot.
- **Clone** (`applyAxisTitleColorOverride`): follows the standard `string | null` override grammar — `undefined` inherits, `null` drops, a literal hex string replaces. Malformed overrides collapse the same way as the writer's normalizer, so the cloned `SheetChart` always carries a value the writer will accept. The title-presence scope rule applies via the same gate the writer uses — when the resolved title is dropped (`title: null` or override), the inherited fill drops silently so the cloned `SheetChart` accurately reflects what the chart will paint.

## Test plan

- [x] `parseChart — axis title color` (12 new tests) — surfaces the uppercase canonical hex from `<a:srgbClr val=\"FF0000\"/>`, normalizes lowercase input to uppercase, strips leading `#`, collapses absence of `<a:solidFill>` to `undefined`, returns undefined when `<c:title>` is absent / the title is a `<c:strRef>` formula reference / `<a:p>` has no `<a:pPr>`, collapses theme-color references (`<a:schemeClr>`) to undefined, drops malformed val tokens (wrong length / non-hex / 8-char alpha form), does not leak from a stray chart-level title `<a:solidFill>`, does not leak from a stray tick-label `<c:txPr>` `<a:solidFill>`, surfaces on both X and Y axes, surfaces independently on both axes, co-surfaces alongside `axisTitleRotation` / `axisTitleFontSize` / `axisTitleBold` / `axisTitleItalic` on the same axis title.
- [x] `writeChart — axis title color` (14 new tests) — omits `<a:solidFill>` by default (theme inherit), emits the fill on a hex input, normalizes lowercase to uppercase, strips leading `#`, drops malformed inputs back to the default, drops non-string inputs, threads through both axes independently, ignores when no axis title renders, composes with `axisTitleRotation` / `axisTitleFontSize` / `axisTitleBold` / `axisTitleItalic`, threads through every chart family that has axes (line / area / scatter), preserves the surrounding `b=\"0\"` / `lang=\"en-US\"` / `sz=\"1000\"` defaults, parse round-trip, default round-trip collapses to `undefined`, full `writeXlsx` package round-trip.
- [x] `cloneChart — axis title color` (17 new tests) — inherits the source's value (X / Y), inherits independently per axis, override replaces, lowercase normalization, leading-`#` strip, `null` drops, replace unset source with override, drop on malformed override, drop on non-string override, normalises a parsed source value that is somehow malformed, drops on `title: null`, drops an override when no axis title resolves, preserves override when override pins a title, composes with chart-level `titleColor`, composes with `axisTitleRotation` / `axisTitleFontSize` / `axisTitleBold` / `axisTitleItalic`, threads through both axes independently on the clone, end-to-end parse -> clone -> write -> reparse round-trip, full `writeXlsx` propagation.
- [x] `pnpm test` — all 4648 tests pass (lint + format + typecheck + vitest).

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -2033,6 +2033,56 @@ export interface SheetChart {
        * family that has axes (bar / column / line / area / scatter).
        */
       axisTitleItalic?: boolean;
+      /**
+       * Axis title font color. Maps to
+       * `<c:catAx><c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr>
+       * <a:solidFill><a:srgbClr val="RRGGBB"/></a:solidFill></a:defRPr>
+       * </a:pPr><a:r><a:rPr><a:solidFill><a:srgbClr val="RRGGBB"/>
+       * </a:solidFill></a:rPr></a:r></a:p></c:rich></c:tx></c:title>
+       * </c:catAx>` (or `<c:valAx>` for scatter / value axes) — Excel's
+       * "Format Axis Title -> Font -> Font Color" picker. The OOXML
+       * `<a:srgbClr val=".."/>` carries the 6-character uppercase hex
+       * sRGB color (`CT_SRgbColor`, ECMA-376 Part 1, §20.1.2.3.32); the
+       * writer lands the fill on both the default-paragraph
+       * `<a:defRPr>` and the literal run's `<a:rPr>` so a re-parse
+       * picks the color up off either canonical slot — Excel keeps the
+       * two values in sync.
+       *
+       * Mirrors {@link SheetChart.titleColor} for axis titles — same
+       * canonical-slot pair, same accept-with-or-without-`#` grammar
+       * (`"FF0000"` / `"#FF0000"` / `"ff0000"` all collapse to the
+       * uppercase canonical form), same `string | null` clone grammar
+       * (`undefined` inherits, `null` drops, a hex string replaces) so
+       * a single configuration call threads cleanly through both the
+       * chart title and either axis title without bookkeeping the
+       * canonical OOXML slots. Mirrors {@link axisTitleRotation} /
+       * {@link axisTitleFontSize} / {@link axisTitleBold} /
+       * {@link axisTitleItalic} for the same `<c:title>` body, so all
+       * five axis-title knobs (rotation, size, bold, italic, color)
+       * compose freely on the same axis.
+       *
+       * Default: omitted — the axis title renders in Excel's reference
+       * inherited theme color (no `<a:solidFill>` element, the writer
+       * skips the fill block entirely). Pin a hex value to render the
+       * axis title in that color (e.g. `"1070CA"` for the dashboard
+       * hero blue the issue-#136 example reaches for). The 8-character
+       * `#RRGGBBAA` form is *not* accepted — alpha lives on
+       * `<a:srgbClr><a:alpha val=".."/>` which is a separate runs-level
+       * knob; pinning `axisTitleColor` carries the RGB triple only.
+       * Malformed inputs (wrong length, non-hex characters, alpha-
+       * channel form) collapse to `undefined` so a stray non-hex token
+       * never produces a malformed `<a:srgbClr>`.
+       *
+       * Silently dropped when the axis renders no title (the
+       * `<c:title>` element is absent in either case) and on `pie` /
+       * `doughnut` charts (no axes at all).
+       *
+       * Sits on every axis flavour — `<c:catAx>` / `<c:valAx>` /
+       * `<c:dateAx>` / `<c:serAx>` all carry the same `<c:title>` shape
+       * per the OOXML schema, so the field round-trips on every chart
+       * family that has axes (bar / column / line / area / scatter).
+       */
+      axisTitleColor?: string;
       gridlines?: ChartAxisGridlines;
       scale?: ChartAxisScale;
       numberFormat?: ChartAxisNumberFormat;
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed the diff end-to-end. The implementation cleanly mirrors the chart-level `titleColor` pattern from #254 and the four prior axis-title knobs (`axisTitleRotation` #248, `axisTitleFontSize` #255, `axisTitleBold` #256, `axisTitleItalic` #257), so the surface stays consistent for callers.

Highlights worth calling out:
- Reader scopes the lookup strictly to `<c:title><c:tx><c:rich>` and bails on each missing link in the canonical chain — the dedicated test for the tick-label `<c:txPr>` sibling and the chart-level title leak both confirm the scoping.
- Writer reuses `normalizeTitleColor` from #254 (accept-with-or-without-`#`, drop on malformed/non-string/8-char alpha forms, uppercase canonical output) and only expands `<a:defRPr>` / `<a:rPr>` from self-closing to wrapping when a fill is actually pinned, keeping byte-for-byte parity with Excel's reference serialization on the default path.
- Clone resolver follows the standard `undefined` (inherit) / `null` (drop) / value (replace) grammar and gates on the resolved title's presence (`xTitle === undefined ? undefined : xAxisTitleColor`), symmetric with the writer's `<c:title>` presence gate so a stray pin on a title-less axis silently disappears at emit time rather than being faithfully preserved into a no-op.
- 43 new tests across reader (12), writer (14), clone (17) cover lossless round-trip, theme-color collapse, malformed token drops, leak guards, axis-independence, multi-knob composition, and full `writeXlsx` packaging.

CI is green. Merging.

PR #259, opened 2026-05-06
feat(xlsx): chart title strikethrough flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr strike=\"..\"/></a:pPr><a:r><a:rPr strike=\"..\"/></a:r></a:p></c:rich></c:tx></c:title>` (ST_TextStrikeType on CT_TextCharacterProperties, ECMA-376 Part 1, §21.1.2.3.7) at the read, write, and clone layers so a template's custom chart-title strikethrough flag survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

Modeled as a boolean for symmetry with `titleBold` (#252) / `titleItalic` (#253): `true` emits `strike=\"sngStrike\"` (Excel's UI checkbox — single line); absence and explicit `false` collapse to omitting the attribute (the OOXML default `\"noStrike\"` collapses to absence, mirroring how Excel's reference serialization emits a non-strikethrough title). The non-UI variant `\"dblStrike\"` (double line) is read-only — the writer emits only `\"sngStrike\"` to keep the surfaced shape consistent with what Excel's reference UI authors, and the reader collapses `\"dblStrike\"` to `undefined` so a templated chart that pinned the double-line variant in raw OOXML round-trips lossless rather than silently downgrading on re-emit.

Mirrors `titleFontSize` (#251), `titleBold` (#252), `titleItalic` (#253), and `titleColor` (#254) for the same `<c:title>` body, so all six chart-title typography knobs (size, rotation, bold, italic, color, strike) compose freely on the same title — the writer threads them through a single `buildTitle` call, the reader picks them up off the shared `<a:defRPr>` slot, and the clone resolver follows the standard `undefined` (inherit) / `null` (drop) / `boolean` (replace) override grammar.

Bridges another title-customization gap for the dashboard composition flow tracked in #136 — useful for marking \"before / after\" or \"deprecated\" dashboard tile headers without leaking the styling out of the chart's own typography slot.

Refs #136

## API

\`\`\`ts

// -- Read side --
console.log(source.titleStrike);
// true       <- when the template pinned <a:defRPr strike=\"sngStrike\"/>
// undefined  <- when the attribute is absent / equals \"noStrike\"
// undefined  <- when the attribute equals \"dblStrike\" (non-UI variant)
// undefined  <- when the chart omits <c:title> or the title is a <c:strRef>

// -- Write side --
await writeXlsx({
  sheets: [{
    name: \"Dashboard\",
    rows: [/* ... */],
    charts: [{
      type: \"column\",
      title: \"Q3 Revenue (deprecated)\",
      titleStrike: true,
      series: [/* ... */],
    }],
  }],
});

// -- Clone-through --
  titleStrike: true,    // pin a strikethrough on a non-strikethrough source
  // titleStrike: null  // drops the inherited flag (writer omits the attribute)
  // titleStrike: false // explicit non-strikethrough (functionally identical to null)
});
\`\`\`

## Implementation notes

- **Type model**: `SheetChart.titleStrike?: boolean` (writer side); `Chart.titleStrike?: boolean` (reader side); `CloneChartOptions.titleStrike?: boolean | null` (clone override).
- **Reader** (`parseTitleStrike`): walks `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr strike=\"...\"/>` for the `strike` attribute. Only the UI-default `\"sngStrike\"` surfaces as `true`; `\"noStrike\"`, `\"dblStrike\"`, and unknown / malformed tokens all collapse to `undefined` to keep the round-trip lossless against the writer's `\"sngStrike\"`-only emit grammar. Lookup is scoped to the chart-level title's rich-text body so a stray `<a:defRPr strike=\"..\">` elsewhere on the chart (e.g. on an axis title or a data-labels block) cannot leak in.
- **Writer**: `buildTitle` now resolves through a `strike?: boolean` parameter — `normalizeTitleStrike` accepts only literal `true` / `false`; every other token (typed escape from an untyped caller — `\"sngStrike\"`, `1`, `null`) collapses to `undefined`. When `true`, the writer lands `strike=\"sngStrike\"` on both the default-paragraph `<a:defRPr>` and the literal run's `<a:rPr>` so a re-parse picks the value up off either canonical slot. Absence / `false` omit the attribute entirely (matches Excel's reference serialization for a non-strikethrough title — the OOXML default `\"noStrike\"` collapses to absence).
- **Clone** (`resolveTitleStrike`): follows the standard `boolean | null` override grammar — `undefined` inherits, `null` drops, a literal boolean replaces. Non-boolean overrides collapse the same way as the writer's normalizer, so the cloned `SheetChart` always carries a value the writer will accept. The title-presence scope rule applies via the same gate the writer uses — when the resolved title is dropped (`title: null`, `showTitle: false`, or a sourceless chart with no `title` override), the inherited flag drops silently so the cloned `SheetChart` accurately reflects what the chart will paint.

## Test plan

- [x] `parseChart — title strike` (10 new tests) — surfaces `true` from `\"sngStrike\"`, collapses `\"noStrike\"` / absence / `\"dblStrike\"` / unknown tokens to `undefined`, returns `undefined` when `<c:title>` / `<c:rich>` / `<a:pPr>` is missing, does not leak from a stray axis-title `<a:defRPr strike=\"sngStrike\"/>`, co-surfaces alongside `titleBold` / `titleItalic` / `titleColor` / `titleFontSize` / `titleRotation` / `titleOverlay`.
- [x] `writeChart — title strike` (13 new tests) — omits the `strike` attribute by default, emits `strike=\"sngStrike\"` on `true`, omits on `false` and on non-boolean inputs (`\"true\"` / `1` / `\"sngStrike\"`), ignores when the chart renders no title (`showTitle: false` / no literal title), composes with `titleBold` / `titleItalic` / `titleColor` / `titleFontSize` / `titleRotation` / `titleOverlay`, lands on both `<a:defRPr>` and `<a:rPr>`, threads through every chart family (bar / column / line / pie / scatter / area), parse round-trip, default round-trip collapses to `undefined`, explicit `false` round-trip collapses to `undefined`, full `writeXlsx` package round-trip.
- [x] `cloneChart — title strike` (12 new tests) — inherits the source's value, override replaces (true / false), `null` drops, undefined source + undefined override yields undefined, override pins `true` on a non-strike source, drops on non-boolean override, drops when `title: null` / `showTitle: false`, preserves when an override pins a fresh title on a sourceless chart, composes with the rest of the chart-title knobs, threads through `line -> column` flatten, end-to-end parse -> clone -> write -> reparse round-trip.
- [x] `pnpm test` — all 4683 tests pass (lint + format + typecheck + vitest).

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1463,6 +1463,56 @@ export interface SheetChart {
    */
   titleColor?: string;
   /**
+   * Chart title strikethrough flag. Maps to
+   * `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr strike=".."/></a:pPr>
+   * <a:r><a:rPr strike=".."/></a:r></a:p></c:rich></c:tx></c:title>` —
+   * Excel's "Format Chart Title -> Font -> Strikethrough" toggle. The
+   * OOXML attribute is the `ST_TextStrikeType` enum on
+   * `CT_TextCharacterProperties` (ECMA-376 Part 1, §21.1.2.3.7) with
+   * three values: `"noStrike"` (the OOXML default — no strikethrough),
+   * `"sngStrike"` (single horizontal line, the value Excel's UI
+   * checkbox emits), and `"dblStrike"` (double horizontal line, a
+   * non-UI variant Excel does not surface in its ribbon). The writer
+   * lands the value on both the default-paragraph `<a:defRPr>` and
+   * the literal run's `<a:rPr>` so a re-parse picks the flag up off
+   * either canonical slot — Excel keeps the two attributes in sync.
+   *
+   * Modeled as a boolean for symmetry with {@link titleBold} /
+   * {@link titleItalic}: `true` emits `strike="sngStrike"` (Excel's
+   * UI "Strikethrough" checkbox — single line). Absence and
+   * non-boolean tokens collapse to omitting the attribute (Excel's
+   * reference serialization for a non-strikethrough title — the
+   * application-default `"noStrike"` collapses to absence). Set
+   * `false` explicitly to pin the non-default `strike="noStrike"`
+   * (functionally identical to omission, but useful when overriding a
+   * templated title that had strikethrough pinned upstream).
+   *
+   * Hucre's writer emits only `"sngStrike"` to keep the surfaced shape
+   * consistent with what Excel's reference UI authors. The reader
+   * collapses the non-UI `"dblStrike"` to `undefined` so a templated
+   * chart that pinned the double-line variant in raw OOXML round-trips
+   * to the same `undefined` an unmarked chart parses to (i.e. the
+   * double-line variant silently downgrades to the single-line write
+   * grammar rather than fabricate a value the writer would re-emit
+   * incorrectly).
+   *
+   * Default: omitted — the title renders without strikethrough (no
+   * `strike` attribute on either slot, Excel's reference serialization
+   * for a fresh chart title). Set `true` to render the title with a
+   * single strikethrough line, useful for marking "before / after" or
+   * "deprecated" dashboard tile headers.
+   *
+   * Silently ignored when no title is rendered (`showTitle === false`
+   * or `title` is absent) — there is no `<c:title>` block to host the
+   * flag in either case. Composes independently with {@link titleBold}
+   * / {@link titleItalic} / {@link titleColor} /
+   * {@link titleFontSize} / {@link titleRotation} /
+   * {@link titleOverlay}: all seven fields land on the same
+   * `<c:title>` element so a single configuration call threads
+   * cleanly through every chart-title knob Excel exposes.
+   */
+  titleStrike?: boolean;
+  /**
    * Auto-title-deleted flag. Maps to `<c:chart><c:autoTitleDeleted
    * val=".."/>` — Excel's record of whether the user explicitly deleted
    * the auto-generated title that single-series charts synthesise from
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed end-to-end. The implementation mirrors the prior sibling PRs (#252 titleBold, #253 titleItalic, #254 titleColor) precisely:

- **Reader** (`parseTitleStrike`): correctly scoped to `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr>` and bails on every missing link in the chain. The collapse of `"dblStrike"` to `undefined` is the right call — surfacing it as `true` would silently downgrade to single-line on round-trip since the writer only emits `"sngStrike"`. The axis-title-leak test confirms the scope guard works.
- **Writer** (`buildTitle` + `normalizeTitleStrike` + `resolveTitleStrike`): lands the attribute on both `<a:defRPr>` and `<a:rPr>` for slot-symmetry with how Excel keeps them in sync. Non-boolean inputs collapse via the normalizer; absence and `false` both omit the attribute (matches Excel reference serialization for non-strike titles).
- **Clone** (`resolveTitleStrike`): standard `undefined`-inherit / `null`-drop / `boolean`-replace grammar; gated on resolved title visibility so the cloned `SheetChart` accurately reflects what gets painted.
- **Tests**: 35 new tests across reader (10), writer (13), clone (12). Coverage hits the dblStrike collapse, axis-title-leak guard, every chart family, every co-composition with the other 6 title knobs, full `writeXlsx` package round-trip, and the typed-escape malformed-input branches.

CI passing (build green), mergeable, no concerns. Merging.

PR #260, opened 2026-05-06
feat(xlsx): chart title underline flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr u=".."/></a:pPr><a:r><a:rPr u=".."/></a:r></a:p></c:rich></c:tx></c:title>` (ST_TextUnderlineType on CT_TextCharacterProperties, ECMA-376 Part 1, §21.1.2.3.7) at the read, write, and clone layers so a template's custom chart-title underline flag survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

Modeled as a boolean for symmetry with `titleBold` (#252) / `titleItalic` (#253) / `titleStrike` (#259): `true` emits `u="sng"` (Excel's UI checkbox — single line); absence and explicit `false` collapse to omitting the attribute (the OOXML default `"none"` collapses to absence, mirroring how Excel's reference serialization emits a non-underlined title). The non-UI variant `"dbl"` (double line) and the sixteen exotic `ST_TextUnderlineType` tokens (`"words"`, `"heavy"`, `"dotted"`, `"dottedHeavy"`, `"dash"`, `"dashHeavy"`, `"dashLong"`, `"dashLongHeavy"`, `"dotDash"`, `"dotDashHeavy"`, `"dotDotDash"`, `"dotDotDashHeavy"`, `"wavy"`, `"wavyHeavy"`, `"wavyDbl"`) are read-only — the writer emits only `"sng"` to keep the surfaced shape consistent with what Excel's reference UI authors, and the reader collapses every non-`"sng"` token to `undefined` so a templated chart that pinned a non-single underline in raw OOXML round-trips lossless rather than silently downgrading on re-emit.

Mirrors `titleFontSize` (#251), `titleBold` (#252), `titleItalic` (#253), `titleColor` (#254), and `titleStrike` (#259) for the same `<c:title>` body, so all eight chart-title typography knobs (size, rotation, bold, italic, color, strike, underline) compose freely on the same title — the writer threads them through a single `buildTitle` call, the reader picks them up off the shared `<a:defRPr>` slot, and the clone resolver follows the standard `undefined` (inherit) / `null` (drop) / `boolean` (replace) override grammar.

Bridges another title-customization gap for the dashboard composition flow tracked in #136 — useful for emphasising "key metric" or "section header" dashboard tile titles without leaking the styling out of the chart's own typography slot.

Refs #136

## API

```ts

// -- Read side --
console.log(source.titleUnderline);
// true       <- when the template pinned <a:defRPr u="sng"/>
// undefined  <- when the attribute is absent / equals "none"
// undefined  <- when the attribute equals "dbl" or any of the 16 exotic types
// undefined  <- when the chart omits <c:title> or the title is a <c:strRef>

// -- Write side --
await writeXlsx({
  sheets: [{
    name: "Dashboard",
    rows: [/* ... */],
    charts: [{
      type: "column",
      title: "Key Metric",
      titleUnderline: true,
      series: [/* ... */],
    }],
  }],
});

// -- Clone-through --
  titleUnderline: true,    // pin an underline on a non-underlined source
  // titleUnderline: null  // drops the inherited flag (writer omits the attribute)
  // titleUnderline: false // explicit non-underline (functionally identical to null)
});
```

## Implementation notes

- **Type model**: `SheetChart.titleUnderline?: boolean` (writer side); `Chart.titleUnderline?: boolean` (reader side); `CloneChartOptions.titleUnderline?: boolean | null` (clone override).
- **Reader** (`parseTitleUnderline`): walks `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr u="...">` for the `u` attribute. Only the UI-default `"sng"` surfaces as `true`; `"none"`, `"dbl"`, every exotic token, and unknown / malformed tokens all collapse to `undefined` to keep the round-trip lossless against the writer's `"sng"`-only emit grammar. Lookup is scoped to the chart-level title's rich-text body so a stray `<a:defRPr u="..">` elsewhere on the chart (e.g. on an axis title or a data-labels block) cannot leak in.
- **Writer**: `buildTitle` now resolves through a `underline?: boolean` parameter — `normalizeTitleUnderline` accepts only literal `true` / `false`; every other token (typed escape from an untyped caller — `"sng"`, `1`, `null`) collapses to `undefined`. When `true`, the writer lands `u="sng"` on both the default-paragraph `<a:defRPr>` and the literal run's `<a:rPr>` so a re-parse picks the value up off either canonical slot. Absence / `false` omit the attribute entirely (matches Excel's reference serialization for a non-underlined title — the OOXML default `"none"` collapses to absence).
- **Clone** (`resolveTitleUnderline`): follows the standard `boolean | null` override grammar — `undefined` inherits, `null` drops, a literal boolean replaces. Non-boolean overrides collapse the same way as the writer's normalizer, so the cloned `SheetChart` always carries a value the writer will accept. The title-presence scope rule applies via the same gate the writer uses — when the resolved title is dropped (`title: null`, `showTitle: false`, or a sourceless chart with no `title` override), the inherited flag drops silently so the cloned `SheetChart` accurately reflects what the chart will paint.

## Test plan

- [x] `parseChart — title underline` (10 new tests) — surfaces `true` from `"sng"`, collapses `"none"` / absence / `"dbl"` / every exotic token / unknown tokens to `undefined`, returns `undefined` when `<c:title>` / `<c:rich>` / `<a:pPr>` is missing, does not leak from a stray axis-title `<a:defRPr u="sng"/>`, co-surfaces alongside `titleBold` / `titleItalic` / `titleStrike` / `titleColor` / `titleFontSize` / `titleRotation` / `titleOverlay`.
- [x] `writeChart — title underline` (13 new tests) — omits the `u` attribute by default, emits `u="sng"` on `true`, omits on `false` and on non-boolean inputs (`"true"` / `1` / `"sng"` / `"dbl"`), ignores when the chart renders no title (`showTitle: false` / no literal title), composes with `titleBold` / `titleItalic` / `titleStrike` / `titleColor` / `titleFontSize` / `titleRotation` / `titleOverlay`, lands on both `<a:defRPr>` and `<a:rPr>`, threads through every chart family (bar / column / line / pie / scatter / area), parse round-trip, default round-trip collapses to `undefined`, explicit `false` round-trip collapses to `undefined`, full `writeXlsx` package round-trip.
- [x] `cloneChart — title underline` (13 new tests) — inherits the source's value, override replaces (true / false), `null` drops, undefined source + undefined override yields undefined, override pins `true` on a non-underlined source, drops on non-boolean override, drops when `title: null` / `showTitle: false`, preserves when an override pins a fresh title on a sourceless chart, composes with the rest of the chart-title knobs, threads through `line -> column` flatten, end-to-end parse -> clone -> write -> reparse round-trip.
- [x] `pnpm test` — all 4719 tests pass (lint + format + typecheck + vitest).

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1513,6 +1513,61 @@ export interface SheetChart {
    */
   titleStrike?: boolean;
   /**
+   * Chart title underline flag. Maps to
+   * `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr u=".."/></a:pPr>
+   * <a:r><a:rPr u=".."/></a:r></a:p></c:rich></c:tx></c:title>` —
+   * Excel's "Format Chart Title -> Font -> Underline" picker. The
+   * OOXML attribute is the `ST_TextUnderlineType` enum on
+   * `CT_TextCharacterProperties` (ECMA-376 Part 1, §21.1.2.3.7) with
+   * eighteen values; the two Excel surfaces in the ribbon are
+   * `"sng"` (single-line underline — the default checkbox) and
+   * `"dbl"` (double-line underline). The remaining sixteen tokens
+   * (`"none"`, `"words"`, `"heavy"`, `"dotted"`, `"dottedHeavy"`,
+   * `"dash"`, `"dashHeavy"`, `"dashLong"`, `"dashLongHeavy"`,
+   * `"dotDash"`, `"dotDashHeavy"`, `"dotDotDash"`,
+   * `"dotDotDashHeavy"`, `"wavy"`, `"wavyHeavy"`, `"wavyDbl"`) are
+   * non-UI variants Excel does not surface in its ribbon. The writer
+   * lands the value on both the default-paragraph `<a:defRPr>` and
+   * the literal run's `<a:rPr>` so a re-parse picks the flag up off
+   * either canonical slot — Excel keeps the two attributes in sync.
+   *
+   * Modeled as a boolean for symmetry with {@link titleBold} /
+   * {@link titleItalic} / {@link titleStrike}: `true` emits `u="sng"`
+   * (Excel's UI "Underline" checkbox — single line). Absence and
+   * non-boolean tokens collapse to omitting the attribute (Excel's
+   * reference serialization for a non-underlined title — the
+   * application-default `"none"` collapses to absence). Set `false`
+   * explicitly to pin the non-default omission (functionally
+   * identical to omission, but useful when overriding a templated
+   * title that had underline pinned upstream).
+   *
+   * Hucre's writer emits only `"sng"` to keep the surfaced shape
+   * consistent with what Excel's reference UI authors. The reader
+   * collapses every non-`"sng"` token (the non-UI `"dbl"` variant
+   * and the sixteen exotic types) to `undefined` so a templated
+   * chart that pinned a non-single underline in raw OOXML round-trips
+   * to the same `undefined` an unmarked chart parses to (i.e. the
+   * exotic underline silently downgrades to the single-line write
+   * grammar rather than fabricate a value the writer would re-emit
+   * incorrectly).
+   *
+   * Default: omitted — the title renders without underline (no `u`
+   * attribute on either slot, Excel's reference serialization for a
+   * fresh chart title). Set `true` to render the title with a single
+   * underline, useful for emphasising "key metric" or "section header"
+   * dashboard tile titles.
+   *
+   * Silently ignored when no title is rendered (`showTitle === false`
+   * or `title` is absent) — there is no `<c:title>` block to host the
+   * flag in either case. Composes independently with {@link titleBold}
+   * / {@link titleItalic} / {@link titleStrike} / {@link titleColor} /
+   * {@link titleFontSize} / {@link titleRotation} /
+   * {@link titleOverlay}: all eight fields land on the same
+   * `<c:title>` element so a single configuration call threads
+   * cleanly through every chart-title knob Excel exposes.
+   */
+  titleUnderline?: boolean;
+  /**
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed the diff. Implementation matches the established titleBold (#252) / titleItalic (#253) / titleColor (#254) / titleStrike (#259) pattern cleanly across all three layers:

- **Reader** (parseTitleUnderline): correctly walks the canonical `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr u="..">` chain, scoped to the chart-level title only. Only `"sng"` surfaces as `true`; `"none"`/`"dbl"`/exotics/absence all collapse to `undefined`. Conservative collapse keeps the round-trip lossless against the writer's `"sng"`-only emit.
- **Writer**: `underlineAttr` lands on both `<a:defRPr>` and `<a:rPr>` — matches Excel's reference behavior of keeping the two slots in sync. `normalizeTitleUnderline` properly defends against non-boolean escapes from untyped callers.
- **Clone**: `resolveTitleUnderline` follows the standard `undefined`(inherit)/`null`(drop)/`boolean`(replace) grammar; the title-presence gate sits in the same `if (resolvedTitle)` block as the other title knobs, so suppressed titles correctly drop the override.

Test coverage is thorough: 10 parser tests (sng/none/dbl/16 exotics/strRef/missing-pPr/axis-title leak/co-surface), 13 writer tests (default omit/true emits sng/false omits/non-boolean drop/showTitle gate/family threading/round-trip), and 13 clone tests (inherit/replace/drop/sourceless/title-null/multi-knob compose/parse->clone->write->reparse). CI is green.

LGTM, merging.

PR #261, opened 2026-05-06
feat(xlsx): chart axis title strikethrough flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:catAx>`/`<c:valAx>`/`<c:dateAx>`/`<c:serAx>`'s `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr strike=".."/></a:pPr><a:r><a:rPr strike=".."/></a:r></a:p></c:rich></c:tx></c:title>` (ST_TextStrikeType on CT_TextCharacterProperties, ECMA-376 Part 1, §21.1.2.3.7) at the read, write, and clone layers so a template's custom axis-title strikethrough flag survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

Modeled as a boolean for symmetry with `axisTitleBold` (#256) / `axisTitleItalic` (#257) and the chart-level `titleStrike` (#259): `true` emits `strike="sngStrike"` (Excel's UI checkbox — single line); absence and explicit `false` collapse to omitting the attribute (the OOXML default `"noStrike"` collapses to absence, mirroring how Excel's reference serialization emits a non-strikethrough axis title). The non-UI variant `"dblStrike"` is read-only — the writer emits only `"sngStrike"` to keep the surfaced shape consistent with what Excel's reference UI authors, and the reader collapses `"dblStrike"` to `undefined` so a templated axis title that pinned the double-line variant in raw OOXML round-trips lossless rather than silently downgrading on re-emit.

Mirrors `axisTitleRotation` (#248), `axisTitleFontSize` (#255), `axisTitleBold` (#256), `axisTitleItalic` (#257), and `axisTitleColor` (#258) for the same `<c:title>` body, so all six axis-title typography knobs (rotation, size, bold, italic, color, strike) compose freely on the same axis — the writer threads them through a single `buildAxisTitle` call, the reader picks them up off the shared `<a:defRPr>` slot, and the clone resolver follows the standard `undefined` (inherit) / `null` (drop) / `boolean` (replace) override grammar. The flag sits on every axis flavour (bar / column / line / area / scatter); pie / doughnut have no axes at all so the field is silently dropped, and an axis with no `title` has no `<c:title>` block to host the flag in either case.

Refs #136

## Test plan

- [x] `pnpm test` passes (lint + typecheck + 4758 vitest tests across 89 files)
- [x] `pnpm build` succeeds
- [x] New parser tests cover `sngStrike` / `noStrike` / `dblStrike` / absence / unknown tokens / `<c:strRef>` titles / missing `<a:pPr>` / cross-axis isolation / chart-level title leak guard / tick-label `<c:txPr>` leak guard
- [x] New writer tests cover default omission / `true` emit / explicit `false` / non-boolean escape drops / X+Y axis independence / no-title gating / composition with rotation/size/bold/italic/color / line/area/scatter family threading / preserved font defaults / round-trip through `parseChart`
- [x] New clone tests cover inherit / `null` drop / boolean replace / unset-source override / non-boolean escape drops / drop on `title=null` / fresh-title pin / six-knob composition / cross-axis independence / end-to-end parse -> clone -> write -> reparse round-trip / `writeXlsx` package round-trip

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -2188,6 +2188,63 @@ export interface SheetChart {
        * family that has axes (bar / column / line / area / scatter).
        */
       axisTitleColor?: string;
+      /**
+       * Axis title strikethrough flag. Maps to
+       * `<c:catAx><c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr strike=".."/></a:pPr>
+       * <a:r><a:rPr strike=".."/></a:r></a:p></c:rich></c:tx></c:title></c:catAx>`
+       * (or `<c:valAx>` for scatter / value axes) — Excel's "Format Axis
+       * Title -> Font -> Strikethrough" toggle. The OOXML attribute is
+       * the `ST_TextStrikeType` enum on `CT_TextCharacterProperties`
+       * (ECMA-376 Part 1, §21.1.2.3.7) with three values: `"noStrike"`
+       * (the OOXML default — no strikethrough), `"sngStrike"` (single
+       * horizontal line, the value Excel's UI checkbox emits), and
+       * `"dblStrike"` (double horizontal line, a non-UI variant Excel
+       * does not surface in its ribbon). The writer lands the value on
+       * both the default-paragraph `<a:defRPr>` and the literal run's
+       * `<a:rPr>` so a re-parse picks the flag up off either canonical
+       * slot — Excel keeps the two attributes in sync.
+       *
+       * Modeled as a boolean for symmetry with {@link axisTitleBold} /
+       * {@link axisTitleItalic}: `true` emits `strike="sngStrike"`
+       * (Excel's UI "Strikethrough" checkbox — single line). Absence
+       * and non-boolean tokens collapse to omitting the attribute
+       * (Excel's reference serialization for a non-strikethrough axis
+       * title — the application-default `"noStrike"` collapses to
+       * absence). Set `false` explicitly to pin the non-default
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed the parse / write / clone layers and the test suite. The implementation cleanly mirrors the established axis-title pattern from #248/#255/#256/#257/#258 and the chart-level titleStrike from #259: `true` emits `strike="sngStrike"` on both `<a:defRPr>` and the literal `<a:rPr>`, absence and `false` collapse to omitting the attribute, and the reader collapses `dblStrike` / unknown tokens / non-UI variants to `undefined` to keep round-trips lossless against the writer's single-line emit. The clone resolver uses the standard `undefined` (inherit) / `null` (drop) / `boolean` (replace) grammar and gates the resolved flag on the matching axis title's presence so a stray inherited flag silently disappears when `title` resolves to undefined. Test coverage is thorough — parser tests cover sngStrike / noStrike / dblStrike / absence / unknown tokens / strRef titles / missing pPr / cross-axis isolation / chart-level title leak guard / tick-label txPr leak guard; writer tests cover default omission / true / false / non-boolean escapes / X+Y independence / no-title gating / six-knob composition / line+area+scatter family threading / preserved font defaults / parseChart round-trip; clone tests cover all override modes plus end-to-end parse->clone->write->reparse and writeXlsx package round-trip. CI is green and the branch is mergeable. Merging.

PR #262, opened 2026-05-06
feat(xlsx): chart axis title underline flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:catAx>`/`<c:valAx>`/`<c:dateAx>`/`<c:serAx>`'s `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr u=\"..\"/></a:pPr><a:r><a:rPr u=\"..\"/></a:r></a:p></c:rich></c:tx></c:title>` (ST_TextUnderlineType on CT_TextCharacterProperties, ECMA-376 Part 1, §21.1.2.3.7) at the read, write, and clone layers so a template's custom axis-title underline flag survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

Modeled as a boolean for symmetry with `axisTitleBold` (#256) / `axisTitleItalic` (#257) / `axisTitleStrike` (#261) and the chart-level `titleUnderline` (#260): `true` emits `u=\"sng\"` (Excel's UI checkbox — single line); absence and explicit `false` collapse to omitting the attribute (the OOXML default `\"none\"` collapses to absence, mirroring how Excel's reference serialization emits a non-underlined axis title). The non-UI variant `\"dbl\"` and the sixteen exotic types (`\"words\"`, `\"heavy\"`, `\"dotted\"`, `\"dottedHeavy\"`, `\"dash\"`, `\"dashHeavy\"`, `\"dashLong\"`, `\"dashLongHeavy\"`, `\"dotDash\"`, `\"dotDashHeavy\"`, `\"dotDotDash\"`, `\"dotDotDashHeavy\"`, `\"wavy\"`, `\"wavyHeavy\"`, `\"wavyDbl\"`) are read-only — the writer emits only `\"sng\"` to keep the surfaced shape consistent with what Excel's reference UI authors, and the reader collapses every non-`\"sng\"` token to `undefined` so a templated axis title that pinned a non-single underline in raw OOXML round-trips lossless rather than silently downgrading on re-emit.

Mirrors `axisTitleRotation` (#248), `axisTitleFontSize` (#255), `axisTitleBold` (#256), `axisTitleItalic` (#257), `axisTitleColor` (#258), and `axisTitleStrike` (#261) for the same `<c:title>` body, so all seven axis-title typography knobs (rotation, size, bold, italic, color, strike, underline) compose freely on the same axis — the writer threads them through a single `buildAxisTitle` call, the reader picks them up off the shared `<a:defRPr>` slot, and the clone resolver follows the standard `undefined` (inherit) / `null` (drop) / `boolean` (replace) override grammar. The flag sits on every axis flavour (bar / column / line / area / scatter); pie / doughnut have no axes at all so the field is silently dropped, and an axis with no `title` has no `<c:title>` block to host the flag in either case.

Refs #136

## Test plan

- [x] `pnpm test` passes (lint + typecheck + 4795 vitest tests across 89 files)
- [x] `pnpm build` succeeds
- [x] New parser tests cover `sng` / `none` / `dbl` / absence / unknown tokens / every exotic ST_TextUnderlineType variant / `<c:strRef>` titles / missing `<a:pPr>` / cross-axis isolation / chart-level title leak guard / tick-label `<c:txPr>` leak guard
- [x] New writer tests cover default omission / `true` emit / explicit `false` / non-boolean escape drops / X+Y axis independence / no-title gating / composition with rotation/size/bold/italic/color/strike / line/area/scatter family threading / preserved font defaults / round-trip through `parseChart`
- [x] New clone tests cover inherit / `null` drop / boolean replace / unset-source override / non-boolean escape drops / drop on `title=null` / fresh-title pin / seven-knob composition / cross-axis independence / end-to-end parse -> clone -> write -> reparse round-trip / `writeXlsx` package round-trip

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -2245,6 +2245,69 @@ export interface SheetChart {
        * family that has axes (bar / column / line / area / scatter).
        */
       axisTitleStrike?: boolean;
+      /**
+       * Axis title underline flag. Maps to
+       * `<c:catAx><c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr u=".."/></a:pPr>
+       * <a:r><a:rPr u=".."/></a:r></a:p></c:rich></c:tx></c:title></c:catAx>`
+       * (or `<c:valAx>` for scatter / value axes) — Excel's "Format Axis
+       * Title -> Font -> Underline" picker. The OOXML attribute is the
+       * `ST_TextUnderlineType` enum on `CT_TextCharacterProperties`
+       * (ECMA-376 Part 1, §21.1.2.3.7) with eighteen values; Excel's UI
+       * exposes only `"sng"` (single line — the default underline
+       * checkbox) and `"dbl"` (double line). The remaining sixteen
+       * tokens (`"none"`, `"words"`, `"heavy"`, `"dotted"`,
+       * `"dottedHeavy"`, `"dash"`, `"dashHeavy"`, `"dashLong"`,
+       * `"dashLongHeavy"`, `"dotDash"`, `"dotDashHeavy"`,
+       * `"dotDotDash"`, `"dotDotDashHeavy"`, `"wavy"`, `"wavyHeavy"`,
+       * `"wavyDbl"`) are non-UI variants Excel does not surface in its
+       * ribbon. The writer lands the value on both the
+       * default-paragraph `<a:defRPr>` and the literal run's `<a:rPr>`
+       * so a re-parse picks the flag up off either canonical slot —
+       * Excel keeps the two attributes in sync.
+       *
+       * Modeled as a boolean for symmetry with {@link axisTitleBold} /
+       * {@link axisTitleItalic} / {@link axisTitleStrike}: `true` emits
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed: implementation cleanly mirrors the established axis-title typography pattern from #256 / #257 / #258 / #261 — same canonical-slot pair (`<a:defRPr>` and `<a:rPr>`) with the value kept in sync, same drop-on-default semantics (only `u="sng"` round-trips, every other ST_TextUnderlineType token collapses to `undefined`), same `undefined` (inherit) / `null` (drop) / boolean (replace) clone grammar, same axis-title-presence gate in `resolveAxes`. Composes cleanly with the existing six axis-title knobs through a single `buildAxisTitle` call. Test coverage is thorough — exotic ST_TextUnderlineType variants, scope guards (chart-title leak, tick-label `<c:txPr>` leak), title=null drop, fresh-title pin, full `writeXlsx` package round-trip. CI is green. Looks good — merging.

PR #263, opened 2026-05-06
feat(xlsx): chart axis tick-label font size — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:catAx><c:txPr><a:p><a:pPr><a:defRPr sz=\"N\"/></a:pPr></a:p></c:txPr></c:catAx>` (and the matching `<c:valAx>` slot) at the read, write, and clone layers so a template's tick-label font size pin survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `axes.x.labelFontSize?: number` and `axes.y.labelFontSize?: number` on `SheetChart` (writer side) plus the matching `labelFontSize?: number` on `ChartAxisInfo` (reader side). Callers pin the value in points; the writer converts to 100ths at emit time. Range: `1..400`pt (the OOXML `ST_TextFontSize` band) with 0.5pt half-step granularity matching Excel's UI. Out-of-range, non-finite, and non-numeric inputs collapse to `undefined` so a fresh chart inherits Excel's reference 10pt tick-label size.

Until now hucre's writer omitted `<c:txPr>` whenever the rotation was unset and the reader ignored the `<a:defRPr sz=\"..\"/>` slot entirely, so a templated dashboard chart whose user shrunk dense numeric tick labels (a common pattern for fitting wide currency totals on a Y axis column) silently lost the choice on every parse -> clone -> write loop. Bridges another axis-customization gap for the dashboard composition flow tracked in #136.

Composes cleanly with the existing `labelRotation` field added in #245 — both knobs land on the same `<c:txPr>` body, the writer emits a single block per axis carrying whichever subset is pinned, and the reader picks up each value off its canonical slot. `<a:bodyPr rot=>` absent collapses to the OOXML default 0, `<a:defRPr sz=>` absent collapses to the OOXML default (theme-inherited), so a chart that pins only one of the two knobs round-trips without leaking a fabricated value through the other slot.

The clone layer follows the standard `number | null` override grammar: `undefined` inherits, `null` drops, and a literal number replaces (clamped / rounded the same way as the writer). The catAx / valAx scope rule applies — pie / doughnut clones drop the inherited field silently so the cloned `SheetChart` accurately reflects what the chart will paint.

Refs #136

## API

```ts

// -- Read side --
console.log(source.axes?.x?.labelFontSize);
// 12         ← when the template pinned <a:defRPr sz=\"1200\"/> on the catAx
// undefined  ← when the template emits no size (theme-inherited)

// -- Write side --
await writeXlsx({
  sheets: [{
    name: \"Dashboard\",
    charts: [{
      type: \"column\",
      // Bump category-axis labels up to match the dashboard typography,
      // and shrink the Y-axis numeric labels to fit the tight column.
      axes: { x: { labelFontSize: 12 }, y: { labelFontSize: 9 } },
    }],
  }],
});

// -- Clone-through --
//   → inherits the source's parsed labelFontSize unchanged.
cloneChart(source, {
  axes: { x: { labelFontSize: null } },
});
//   → drops the inherited size (writer falls back to the theme default).
cloneChart(source, {
  axes: { x: { labelFontSize: 14 } },
});
//   → replaces the inherited size with 14pt.
```

## Implementation notes

- **Type model**: `axes.x.labelFontSize?: number` and `axes.y.labelFontSize?: number` on `SheetChart` (writer side) plus the matching `labelFontSize?: number` on `ChartAxisInfo` (reader side). Same shape as `axisTitleFontSize` so a parsed value flows straight from the read side back into the write side without transformation.
- **Reader** (`parseAxisLabelFontSize`): walks `<c:txPr><a:p><a:pPr><a:defRPr sz=\"N\"/></a:pPr></a:p></c:txPr>` for the `sz` attribute. The lookup is scoped to the axis-level `<c:txPr>` so a stray `<a:defRPr sz=\"..\"/>` inside `<c:title><c:tx><c:rich>` (which belongs to `axisTitleFontSize`) cannot leak in. `Number.parseInt` rejects non-numeric tokens; out-of-range values drop rather than clamp (matching the chart-level title-size contract). Surfaces on every axis flavour.
- **Writer**: extends the existing `buildAxisTxPr(rotationDeg, fontSizePt)` helper to optionally include a `<a:defRPr sz=\"N\"/>` inside `<a:pPr>`. The block is emitted whenever either knob is pinned; absent inputs collapse so a chart that pins only one knob keeps the other slot at its OOXML default. Threads through bar / column / line / area via `buildBarAxes` and scatter via `buildScatterAxes`; pie / doughnut have no axes at all and the field is silently dropped on those families.
  - `normalizeAxisLabelFontSize(value)` rounds inputs to the nearest 0.5pt and collapses out-of-range / non-finite / non-numeric inputs to `undefined`.
- **Clone**: `applyLabelFontSizeOverride` follows the standard `number | null` override grammar — `undefined` inherits, `null` drops, a literal number replaces. Out-of-range / non-numeric overrides collapse via `normalizeTitleFontSize` so the cloned `SheetChart` always carries a value the writer will accept. Pie / doughnut clones drop the inherited field silently via the same axis-scope short-circuit the surrounding axis helpers use.
- **Validation**: only literal finite numbers in the `1..400`pt band survive the type guard at every layer; non-numeric inputs collapse to the default.

## Test plan

- [x] `parseChart — axis labelFontSize` (14 new tests) — surfaces the size in points from the `sz` attribute, surfaces the reference 10pt size, rounds half-point inputs, returns undefined when `<c:txPr>` / `<a:defRPr>` / `<a:p>` are absent, drops out-of-range and non-numeric tokens, surfaces on the value axis, surfaces independently on both axes, co-surfaces alongside rotation, isolates from the axis title's `<c:rich>` block, co-surfaces alongside other axis fields.
- [x] `writeChart — axis labelFontSize` (18 new tests) — omits `<c:txPr>` by default, emits the X-axis size, emits the Y-axis size, emits independently on both axes, rounds fractional inputs, drops out-of-range / non-finite / non-numeric inputs, composes with `labelRotation` in a single block, emits `<a:bodyPr/>` (no rot) when only the size is pinned, emits `<a:defRPr/>` (no sz) when only rotation is pinned, threads through bar / column / line / area / scatter, drops on pie / doughnut, places `<c:txPr>` between `<c:tickLblPos>` and `<c:crossAx>` per the OOXML schema, parse round-trip, half-point round-trip, default round-trip collapse, full `writeXlsx` package round-trip.
- [x] `cloneChart — axis labelFontSize` (16 new tests) — inherit / null drop / replace semantics, round fractional overrides, collapse out-of-range / non-finite / non-numeric overrides, add to a source axis without one, drop on pie / doughnut chart-type coercion, carry through bar -> column coercion, compose with rotation, compose with other axis overrides, end-to-end parse -> clone -> write -> reparse, full `writeXlsx` cloned-chart round-trip.
- [x] `pnpm test` — all 4808 tests pass (lint + typecheck + vitest).

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -2301,6 +2301,39 @@ export interface SheetChart {
        */
       labelRotation?: number;
       /**
+       * Tick-label font size in whole or half points. Maps to
+       * `<c:catAx><c:txPr><a:p><a:pPr><a:defRPr sz="N"/></a:pPr></a:p></c:txPr></c:catAx>`
+       * (or `<c:valAx>` for scatter / value axes) — Excel's "Format
+       * Axis -> Font -> Size" knob applied to the tick labels. The
+       * OOXML attribute is in 100ths of a point, so 12pt serializes as
+       * `sz="1200"` and 10pt (Excel's reference default for tick
+       * labels) as `sz="1000"`; the writer performs the conversion at
+       * emit time and lands the value on the default-paragraph
+       * `<a:defRPr>` slot.
+       *
+       * Mirrors {@link SheetChart.axes.x.axisTitleFontSize} for tick
+       * labels — same `1..400`pt band the OOXML `ST_TextFontSize`
+       * schema exposes, same 0.5pt half-step granularity Excel's UI
+       * exposes, same out-of-range / non-finite drop semantics. Useful
+       * for shrinking dense numeric tick labels on a value axis to fit
+       * a tight chart frame, or bumping category-axis labels up to
+       * match a presentation slide's typography.
+       *
+       * Default: omitted — the tick labels render at Excel's reference
+       * size (10pt). Composes independently with
+       * {@link labelRotation}: both fields land on the same
+       * `<c:txPr>` body, so a single configuration call threads cleanly
+       * through both knobs.
+       *
+       * Sits on every axis flavour — `<c:catAx>` / `<c:valAx>` /
+       * `<c:dateAx>` / `<c:serAx>` all carry an optional `<c:txPr>` per
+       * the OOXML schema, so the field round-trips on every chart family
+       * that has axes (bar / column / line / area / scatter). Pie /
+       * doughnut have no axes at all, so the field is silently dropped
+       * on those families.
+       */
+      labelFontSize?: number;
+      /**
        * Reverse the axis plotting order. Maps to
        * `<c:scaling><c:orientation val="maxMin"/></c:scaling>` —
        * Excel's "Categories in reverse order" / "Values in reverse
@@ -2647,6 +2680,22 @@ export interface SheetChart {
        * charts (no axes at all).
        */
       labelRotation?: number;
+      /**
+       * Tick-label font size in points for the value axis. Maps to
+       * `<c:valAx><c:txPr><a:p><a:pPr><a:defRPr sz="N"/></a:pPr></a:p></c:txPr></c:valAx>`.
+       * Mirrors {@link SheetChart.axes.x.labelFontSize} for the value
+       * axis — see that field for the full semantics. The OOXML `sz`
+       * attribute is in 100ths of a point; the writer converts at emit
+       * time so callers pin the value in points. Range: `1..400`pt
+       * (the OOXML `ST_TextFontSize` band), with 0.5pt granularity.
+       *
+       * Useful for shrinking dense numeric tick labels (e.g. wide
+       * currency totals) so they fit a tight Y-axis column, or
+       * bumping the value-axis labels up to match a chart-level
+       * typography pin. Silently dropped on `pie` / `doughnut`
+       * charts (no axes at all).
+       */
+      labelFontSize?: number;
       /**
        * Hide the entire value axis (line, tick marks, tick labels).
        * Maps to `<c:valAx><c:delete val="1"/></c:valAx>`. Default:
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed: cleanly extends `buildAxisTxPr` to share the `<c:txPr>` slot between `labelRotation` (#245) and the new `labelFontSize` — when only one knob is pinned the other slot collapses to its OOXML default (rotation absent → no `rot` on `<a:bodyPr>`, size absent → no `sz` on `<a:defRPr>`) so a chart that pins a single field round-trips without leaking a fabricated value through the other. The reader / writer / clone normalizers all defer to the existing `normalizeTitleFontSize` for the `1..400`pt band and 0.5pt half-step granularity, keeping the title and tick-label sizing rules in lockstep. The lookup is correctly scoped to the axis-level `<c:txPr>` so a stray `<a:defRPr sz=".."/>` inside `<c:title><c:tx><c:rich>` (which feeds `axisTitleFontSize`) cannot leak in. Test coverage is thorough — half-point rounding, out-of-range / non-finite drops, composition with rotation, single-knob isolation, full `writeXlsx` package round-trip, scatter / line / area / bar threading, pie / doughnut drop. CI is green. Looks good — merging.

PR #264, opened 2026-05-06
feat(xlsx): chart axis tick-label bold flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:catAx><c:txPr><a:p><a:pPr><a:defRPr b=\"..\"/></a:pPr></a:p></c:txPr></c:catAx>` (and the matching `<c:valAx>` slot) at the read, write, and clone layers so a template's tick-label bold pin survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `axes.x.labelBold?: boolean` and `axes.y.labelBold?: boolean` on `SheetChart` (writer side) plus the matching `labelBold?: boolean` on `ChartAxisInfo` (reader side). Until now hucre's writer omitted the `b` attribute on `<a:defRPr>` for tick-label `<c:txPr>` and the reader ignored the slot entirely, so a templated dashboard chart whose user bolded the bottom-axis category names (a common pattern when an axis label needs to read as a header in a busy chart frame) silently lost the choice on every parse -> clone -> write loop. Bridges another typography-customization gap for the dashboard composition flow tracked in #136.

Composes cleanly with the existing `labelRotation` (#245) and `labelFontSize` (#263) fields — all three knobs land on the same `<c:txPr>` body, the writer emits a single block per axis carrying whichever subset is pinned, and the reader picks up each value off its canonical slot. `<a:bodyPr rot=>` absent collapses to the OOXML default 0, `<a:defRPr sz=>` absent collapses to the theme-inherited size, and `<a:defRPr b=>` absent collapses to the theme-inherited weight, so a chart that pins only one of the three knobs round-trips without leaking a fabricated value through the other slots.

The clone layer follows the standard `boolean | null` override grammar: `undefined` inherits, `null` drops, and a literal `boolean` replaces. Non-boolean overrides (typed escapes from an untyped caller) collapse to `undefined` so the cloned `SheetChart` always carries a value the writer will accept. Pie / doughnut clones drop the inherited field silently via the same axis-scope short-circuit the surrounding axis helpers use.

Refs #136

## API

\`\`\`ts

// -- Read side --
console.log(source.axes?.x?.labelBold);
// true       <- when the template pinned <a:defRPr b=\"1\"/> on the catAx
// undefined  <- when the template emits no flag (theme-inherited)

// -- Write side --
await writeXlsx({
  sheets: [{
    name: \"Dashboard\",
    charts: [{
      type: \"column\",
      // Bold the category-axis labels so they read as headers,
      // and leave the Y-axis numeric labels at the theme weight.
      axes: { x: { labelBold: true } },
    }],
  }],
});

// -- Clone-through --
//   -> inherits the source's parsed labelBold unchanged.
cloneChart(source, {
  axes: { x: { labelBold: null } },
});
//   -> drops the inherited flag (writer falls back to the theme default).
cloneChart(source, {
  axes: { x: { labelBold: false } },
});
//   -> replaces the inherited flag with the explicit OOXML default b=\"0\".
\`\`\`

## Implementation notes

- **Type model**: `axes.x.labelBold?: boolean` and `axes.y.labelBold?: boolean` on `SheetChart` (writer side) plus the matching `labelBold?: boolean` on `ChartAxisInfo` (reader side). Same shape as `axisTitleBold` so a parsed value flows straight from the read side back into the write side without transformation.
- **Reader** (`parseAxisLabelBold`): walks `<c:txPr><a:p><a:pPr><a:defRPr b=\"..\"/></a:pPr></a:p></c:txPr>` for the `b` attribute and runs it through the existing `parseBoolAttr` (accepts `\"0\"` / `\"1\"` / `\"true\"` / `\"false\"`). The lookup is scoped to the axis-level `<c:txPr>` so a stray `<a:defRPr b=\"..\"/>` inside `<c:title><c:tx><c:rich>` (which belongs to `axisTitleBold`) cannot leak in. The OOXML default `false` collapses to `undefined` so absence and `b=\"0\"` round-trip identically; only an explicit `b=\"1\"` surfaces `true`. Surfaces on every axis flavour.
- **Writer**: extends the existing `buildAxisTxPr(rotationDeg, fontSizePt, bold)` helper to optionally include a `b=\"1\"` / `b=\"0\"` on `<a:defRPr>`. The block is emitted whenever any of the three knobs is pinned; absent inputs collapse so a chart that pins only one knob keeps the other slots at their OOXML defaults. Threads through bar / column / line / area via `buildBarAxes` and scatter via `buildScatterAxes`; pie / doughnut have no axes at all and the field is silently dropped on those families.
  - `normalizeAxisLabelBold(value)` delegates to the chart-level `normalizeTitleBold` — `true` / `false` pass through literally, every other token collapses to `undefined`.
- **Clone**: `applyLabelBoldOverride` follows the standard `boolean | null` override grammar — `undefined` inherits, `null` drops, a literal boolean replaces. Non-boolean overrides collapse to `undefined`. Pie / doughnut clones drop the inherited field silently via the same axis-scope short-circuit the surrounding axis helpers use.

## Test plan

- [x] `parseChart — axis labelBold` (12 new tests) — surfaces the flag from `b=\"1\"` / `b=\"true\"`, drops the OOXML default `b=\"0\"` / `b=\"false\"` to undefined, drops malformed tokens, returns undefined when `<c:txPr>` / `<a:defRPr>` / `<a:p>` are absent, surfaces on the value axis, surfaces independently on both axes, co-surfaces alongside rotation and font size, isolates from the axis title's `<c:rich>` block, co-surfaces alongside other axis fields.
- [x] `writeChart — axis labelBold` (16 new tests) — omits `<c:txPr>` by default, emits `b=\"1\"` / `b=\"0\"`, emits independently on both axes, drops non-boolean inputs, composes with `labelRotation` / `labelFontSize` in a single block, threads through bar / column / line / area / scatter, drops on pie / doughnut, places `<c:txPr>` between `<c:tickLblPos>` and `<c:crossAx>`, parse round-trip, default-collapse round-trip, full `writeXlsx` package round-trip.
- [x] `cloneChart — axis labelBold` (12 new tests) — inherit / null drop / replace semantics, collapse non-boolean overrides, add to a source axis without one, drop on pie / doughnut chart-type coercion, carry through bar -> column coercion, compose with rotation and font size, compose with other axis overrides, end-to-end parse -> clone -> write -> reparse, full `writeXlsx` cloned-chart round-trip.
- [x] `pnpm vitest run --dir=test` — all 4889 tests pass.
- [x] `pnpm typecheck` — clean.

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -2397,6 +2397,43 @@ export interface SheetChart {
        */
       labelFontSize?: number;
       /**
+       * Tick-label bold flag. Maps to
+       * `<c:catAx><c:txPr><a:p><a:pPr><a:defRPr b=".."/></a:pPr></a:p></c:txPr></c:catAx>`
+       * (or `<c:valAx>` for scatter / value axes) — Excel's "Format
+       * Axis -> Font -> Bold" toggle applied to the tick labels. The
+       * OOXML attribute is the `xsd:boolean` `b` on
+       * `CT_TextCharacterProperties` (ECMA-376 Part 1, §21.1.2.3.7);
+       * the writer lands the value on the default-paragraph
+       * `<a:defRPr>` slot inside the same `<c:txPr>` block that
+       * carries {@link labelRotation} / {@link labelFontSize}.
+       *
+       * Mirrors {@link SheetChart.axes.x.axisTitleBold} for tick
+       * labels — same `boolean | null` clone grammar, same silent drop
+       * on `pie` / `doughnut` (no axes at all). Useful for emphasising
+       * dashboard tick labels (e.g. bolding the bottom-axis category
+       * names so they read as headers in a busy chart frame).
+       *
+       * Default: omitted — the tick labels render non-bold (the OOXML
+       * default; the writer elides the `b` attribute when no value is
+       * pinned). Set `true` to emit `b="1"` on the default-paragraph
+       * slot; set `false` explicitly to pin the OOXML default `b="0"`
+       * (functionally identical to omission, but useful when overriding
+       * a templated chart that had bold pinned upstream).
+       *
+       * Composes independently with {@link labelRotation} /
+       * {@link labelFontSize}: all three knobs land on the same
+       * `<c:txPr>` body, so a single configuration call threads cleanly
+       * through every tick-label knob the writer exposes.
+       *
+       * Sits on every axis flavour — `<c:catAx>` / `<c:valAx>` /
+       * `<c:dateAx>` / `<c:serAx>` all carry an optional `<c:txPr>` per
+       * the OOXML schema, so the field round-trips on every chart family
+       * that has axes (bar / column / line / area / scatter). Pie /
+       * doughnut have no axes at all, so the field is silently dropped
+       * on those families.
+       */
+      labelBold?: boolean;
+      /**
        * Reverse the axis plotting order. Maps to
        * `<c:scaling><c:orientation val="maxMin"/></c:scaling>` —
        * Excel's "Categories in reverse order" / "Values in reverse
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed end-to-end: types, reader (`parseAxisLabelBold`), writer (`buildAxisTxPr` extension + `normalizeAxisLabelBold`), and clone (`applyLabelBoldOverride`). The implementation mirrors the established pattern from PR #245 (`labelRotation`) and PR #263 (`labelFontSize`), reuses the same `<c:txPr>` block, and shares scoping logic with `parseAxisLabelFontSize` so the lookup cannot leak in from `<c:title><c:rich>`.

Strengths:
- Consistent `boolean | null` clone grammar with the rest of the axis override surface.
- Default-`false` collapse to `undefined` keeps absence and `b="0"` indistinguishable through round-trip — matches `parseAxisTitleBold`.
- Shared `normalizeTitleBold` reuse for non-boolean coercion.
- 40 new tests across reader / writer / clone covering composition with rotation/size, axis isolation, title-vs-tick-label scoping, and pie/doughnut drop. CI is green.

Minor observations (not blocking):
- The single-line `applyLabelBoldOverride` body could be simplified to `override === undefined ? (typeof source === "boolean" ? source : undefined) : override === null ? undefined : (typeof override === "boolean" ? override : undefined)`, but the current expanded form matches the surrounding helpers' style.
- README/web docs do not currently surface any of `labelRotation`/`labelFontSize`/`labelBold`, consistent with prior practice in this area.

Merging.

PR #265, opened 2026-05-06
feat(xlsx): chart axis tick-label italic flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:catAx><c:txPr><a:p><a:pPr><a:defRPr i=\"..\"/></a:pPr></a:p></c:txPr></c:catAx>` (and the matching `<c:valAx>` slot) at the read, write, and clone layers so a template's tick-label italic pin survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `axes.x.labelItalic?: boolean` and `axes.y.labelItalic?: boolean` on `SheetChart` (writer side) plus the matching `labelItalic?: boolean` on `ChartAxisInfo` (reader side). Until now hucre's writer omitted the `i` attribute on `<a:defRPr>` for tick-label `<c:txPr>` and the reader ignored the slot entirely, so a templated dashboard chart whose user italicised the bottom-axis category names (a common pattern when a typography accent should land on the tick labels without bolding them) silently lost the choice on every parse -> clone -> write loop. Bridges another typography-customization gap for the dashboard composition flow tracked in #136.

Composes cleanly with the existing `labelRotation` (#245), `labelFontSize` (#263), and `labelBold` (#264) fields — all four knobs land on the same `<c:txPr>` body, the writer emits a single block per axis carrying whichever subset is pinned, and the reader picks up each value off its canonical slot. `<a:bodyPr rot=>` absent collapses to the OOXML default 0, `<a:defRPr sz=>` absent collapses to the theme-inherited size, `<a:defRPr b=>` absent collapses to the theme-inherited weight, and `<a:defRPr i=>` absent collapses to the theme-inherited slant, so a chart that pins only one of the four knobs round-trips without leaking a fabricated value through the other slots.

The clone layer follows the standard `boolean | null` override grammar: `undefined` inherits, `null` drops, and a literal `boolean` replaces. Non-boolean overrides (typed escapes from an untyped caller) collapse to `undefined` so the cloned `SheetChart` always carries a value the writer will accept. Pie / doughnut clones drop the inherited field silently via the same axis-scope short-circuit the surrounding axis helpers use.

Refs #136

## API

```ts

// -- Read side --
console.log(source.axes?.x?.labelItalic);
// true       <- when the template pinned <a:defRPr i=\"1\"/> on the catAx
// undefined  <- when the template emits no flag (theme-inherited)

// -- Write side --
await writeXlsx({
  sheets: [{
    name: \"Dashboard\",
    charts: [{
      type: \"column\",
      // Italicise the category-axis labels for a typographic accent
      // without bolding them; leave the Y-axis numeric labels at the
      // theme slant.
      axes: { x: { labelItalic: true } },
    }],
  }],
});

// -- Clone-through --
//   -> inherits the source's parsed labelItalic unchanged.
cloneChart(source, {
  axes: { x: { labelItalic: null } },
});
//   -> drops the inherited flag (writer falls back to the theme default).
cloneChart(source, {
  axes: { x: { labelItalic: false } },
});
//   -> replaces the inherited flag with the explicit OOXML default i=\"0\".
```

## Implementation notes

- **Type model**: `axes.x.labelItalic?: boolean` and `axes.y.labelItalic?: boolean` on `SheetChart` (writer side) plus the matching `labelItalic?: boolean` on `ChartAxisInfo` (reader side). Same shape as `labelBold` so a parsed value flows straight from the read side back into the write side without transformation.
- **Reader** (`parseAxisLabelItalic`): walks `<c:txPr><a:p><a:pPr><a:defRPr i=\"..\"/></a:pPr></a:p></c:txPr>` for the `i` attribute and runs it through the existing `parseBoolAttr` (accepts `\"0\"` / `\"1\"` / `\"true\"` / `\"false\"`). The lookup is scoped to the axis-level `<c:txPr>` so a stray `<a:defRPr i=\"..\"/>` inside `<c:title><c:tx><c:rich>` (which belongs to `axisTitleItalic`) cannot leak in. The OOXML default `false` collapses to `undefined` so absence and `i=\"0\"` round-trip identically; only an explicit `i=\"1\"` surfaces `true`. Surfaces on every axis flavour.
- **Writer**: extends the existing `buildAxisTxPr(rotationDeg, fontSizePt, bold, italic)` helper to optionally include an `i=\"1\"` / `i=\"0\"` on `<a:defRPr>`. The block is emitted whenever any of the four knobs is pinned; absent inputs collapse so a chart that pins only one knob keeps the other slots at their OOXML defaults. Threads through bar / column / line / area via `buildBarAxes` and scatter via `buildScatterAxes`; pie / doughnut have no axes at all and the field is silently dropped on those families.
  - `normalizeAxisLabelItalic(value)` delegates to the chart-level `normalizeTitleItalic` — `true` / `false` pass through literally, every other token collapses to `undefined`.
- **Clone**: `applyLabelItalicOverride` follows the standard `boolean | null` override grammar — `undefined` inherits, `null` drops, a literal boolean replaces. Non-boolean overrides collapse to `undefined`. Pie / doughnut clones drop the inherited field silently via the same axis-scope short-circuit the surrounding axis helpers use.

## Test plan

- [x] `parseChart — axis labelItalic` (12 new tests) — surfaces the flag from `i=\"1\"` / `i=\"true\"`, drops the OOXML default `i=\"0\"` / `i=\"false\"` to undefined, drops malformed tokens, returns undefined when `<c:txPr>` / `<a:defRPr>` / `<a:p>` are absent, surfaces on the value axis, surfaces independently on both axes, co-surfaces alongside rotation / font size / bold, isolates from the axis title's `<c:rich>` block, co-surfaces alongside other axis fields.
- [x] `writeChart — axis labelItalic` (17 new tests) — omits `<c:txPr>` by default, emits `i=\"1\"` / `i=\"0\"`, emits independently on both axes, drops non-boolean inputs, composes with `labelRotation` / `labelFontSize` / `labelBold` in a single block, threads through bar / column / line / area / scatter, drops on pie / doughnut, places `<c:txPr>` between `<c:tickLblPos>` and `<c:crossAx>`, parse round-trip, default-collapse round-trip, full `writeXlsx` package round-trip.
- [x] `cloneChart — axis labelItalic` (12 new tests) — inherit / null drop / replace semantics, collapse non-boolean overrides, add to a source axis without one, drop on pie / doughnut chart-type coercion, carry through bar -> column coercion, compose with rotation / font size / bold, compose with other axis overrides, end-to-end parse -> clone -> write -> reparse, full `writeXlsx` cloned-chart round-trip.
- [x] `pnpm test` — 4934 tests pass, lint clean, typecheck clean.

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -2434,6 +2434,47 @@ export interface SheetChart {
        */
       labelBold?: boolean;
       /**
+       * Tick-label italic flag. Maps to
+       * `<c:catAx><c:txPr><a:p><a:pPr><a:defRPr i=".."/></a:pPr></a:p></c:txPr></c:catAx>`
+       * (or `<c:valAx>` for scatter / value axes) — Excel's "Format
+       * Axis -> Font -> Italic" toggle applied to the tick labels. The
+       * OOXML attribute is the `xsd:boolean` `i` on
+       * `CT_TextCharacterProperties` (ECMA-376 Part 1, §21.1.2.3.7);
+       * the writer lands the value on the default-paragraph
+       * `<a:defRPr>` slot inside the same `<c:txPr>` block that
+       * carries {@link labelRotation} / {@link labelFontSize} /
+       * {@link labelBold}.
+       *
+       * Mirrors {@link SheetChart.axes.x.axisTitleItalic} for tick
+       * labels — same `boolean | null` clone grammar, same silent drop
+       * on `pie` / `doughnut` (no axes at all). Useful for italicising
+       * dashboard tick labels (e.g. emphasising the bottom-axis
+       * category names so they stand out against the chart frame
+       * without bolding them).
+       *
+       * Default: omitted — the tick labels render non-italic (the
+       * OOXML default; the writer elides the `i` attribute when no
+       * value is pinned). Set `true` to emit `i="1"` on the
+       * default-paragraph slot; set `false` explicitly to pin the
+       * OOXML default `i="0"` (functionally identical to omission, but
+       * useful when overriding a templated chart that had italic
+       * pinned upstream).
+       *
+       * Composes independently with {@link labelRotation} /
+       * {@link labelFontSize} / {@link labelBold}: all four knobs land
+       * on the same `<c:txPr>` body, so a single configuration call
+       * threads cleanly through every tick-label knob the writer
+       * exposes.
+       *
+       * Sits on every axis flavour — `<c:catAx>` / `<c:valAx>` /
+       * `<c:dateAx>` / `<c:serAx>` all carry an optional `<c:txPr>` per
+       * the OOXML schema, so the field round-trips on every chart family
+       * that has axes (bar / column / line / area / scatter). Pie /
+       * doughnut have no axes at all, so the field is silently dropped
+       * on those families.
+       */
+      labelItalic?: boolean;
+      /**
        * Reverse the axis plotting order. Maps to
        * `<c:scaling><c:orientation val="maxMin"/></c:scaling>` —
        * Excel's "Categories in reverse order" / "Values in reverse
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed: the implementation cleanly mirrors the existing labelBold / labelFontSize / labelRotation pattern across reader, writer, and clone layers, with the italic flag landing on the same <c:txPr> body (sz / b / i sharing one <a:defRPr> emission). Type model is symmetric on both SheetChart.axes.x|y and ChartAxisInfo. Reader is correctly scoped to the axis-level <c:txPr> so it cannot bleed in from <c:title><c:rich>, OOXML default i="0" collapses to undefined for round-trip equivalence with absence, clone follows the standard inherit / null-drop / replace grammar with non-boolean overrides collapsing safely. Test coverage (41 new tests) is thorough — co-surfacing with the other three knobs, scatter / bar / column / line / area threading, pie / doughnut drop, schema-correct positioning between <c:tickLblPos> and <c:crossAx>, and end-to-end parse → clone → write → reparse round-trips. CI green, no concerns. Merging.

PR #266, opened 2026-05-06
feat(xlsx): chart axis tick-label font color — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:catAx><c:txPr><a:p><a:pPr><a:defRPr><a:solidFill><a:srgbClr val=\"RRGGBB\"/></a:solidFill></a:defRPr></a:pPr></a:p></c:txPr></c:catAx>` (and the matching `<c:valAx>` slot) at the read, write, and clone layers so a template's tick-label color survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `axes.x.labelColor?: string` and `axes.y.labelColor?: string` on `SheetChart` (writer side) plus the matching `labelColor?: string` on `ChartAxisInfo` (reader side). Until now hucre's writer omitted the `<a:solidFill>` block on the tick-label `<a:defRPr>` and the reader ignored the slot entirely, so a templated dashboard chart whose user tinted the tick labels (a common pattern when a stylistic accent should land on the value-axis numeric ticks without affecting the chart frame) silently lost the choice on every parse -> clone -> write loop. Bridges another typography-customization gap for the dashboard composition flow tracked in #136.

Composes cleanly with the existing `labelRotation` (#245), `labelFontSize` (#263), `labelBold` (#264), and `labelItalic` (#265) fields — all five knobs land on the same `<c:txPr>` body, the writer emits a single block per axis carrying whichever subset is pinned, and the reader picks up each value off its canonical slot. `<a:bodyPr rot=>` absent collapses to the OOXML default 0, `<a:defRPr sz=>` absent collapses to the theme-inherited size, `<a:defRPr b=>` / `i=>` absent collapse to the theme-inherited weight / slant, and the absence of `<a:solidFill>` collapses to the theme-inherited text color, so a chart that pins only one of the five knobs round-trips without leaking a fabricated value through the other slots.

Mirrors the axis-title counterpart `axisTitleColor` (#258) — same accept-with-or-without-`#` hex grammar (`\"FF0000\"` / `\"#FF0000\"` / `\"ff0000\"` all collapse to the uppercase canonical form), same `string | null` clone grammar (`undefined` inherits, `null` drops, a hex string replaces). Malformed overrides (typed escapes from an untyped caller, wrong length, non-hex characters, alpha-channel forms) collapse to `undefined` so the cloned `SheetChart` always carries a value the writer will accept. Theme references (`<a:schemeClr>`), `<a:hslClr>`, `<a:sysClr>`, and `<a:prstClr>` all collapse to `undefined` on read since only the literal RGB triple round-trips losslessly through the writer. Pie / doughnut clones short-circuit upstream — neither family has axes — so the field silently drops on those families.

Refs #136

## Test plan

- [x] `pnpm test` passes (4980 tests across 89 files, lint + format + typecheck + vitest)
- [x] `pnpm build` succeeds
- [x] New describe blocks: `parseChart — axis labelColor` (10 cases), `writeChart — axis labelColor` (16 cases), `cloneChart — axis labelColor` (18 cases) covering parse, write, end-to-end `writeXlsx`, malformed input, theme-color references, pie / doughnut short-circuit, chart-type coercion, and composition with the four sibling typography knobs.

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -2475,6 +2475,54 @@ export interface SheetChart {
        */
       labelItalic?: boolean;
       /**
+       * Tick-label font color. Maps to
+       * `<c:catAx><c:txPr><a:p><a:pPr><a:defRPr><a:solidFill>
+       * <a:srgbClr val="RRGGBB"/></a:solidFill></a:defRPr></a:pPr></a:p>
+       * </c:txPr></c:catAx>` (or `<c:valAx>` for scatter / value axes) —
+       * Excel's "Format Axis -> Font -> Font Color" picker applied to
+       * the tick labels. The OOXML `<a:srgbClr val=".."/>` carries the
+       * 6-character uppercase hex sRGB color (`CT_SRgbColor`,
+       * ECMA-376 Part 1, §20.1.2.3.32); the writer lands the fill on
+       * the default-paragraph `<a:defRPr>` slot inside the same
+       * `<c:txPr>` block that carries {@link labelRotation} /
+       * {@link labelFontSize} / {@link labelBold} / {@link labelItalic}.
+       *
+       * Mirrors {@link SheetChart.axes.x.axisTitleColor} for tick
+       * labels — same accept-with-or-without-`#` grammar (`"FF0000"` /
+       * `"#FF0000"` / `"ff0000"` all collapse to the uppercase
+       * canonical form), same `string | null` clone grammar
+       * (`undefined` inherits, `null` drops, a hex string replaces).
+       * Useful for tinting dashboard tick labels (e.g. dimming the
+       * value-axis numeric ticks to a muted grey so the data series
+       * read as the visual focus of a busy chart frame).
+       *
+       * Default: omitted — the tick labels render in Excel's reference
+       * inherited theme color (no `<a:solidFill>` element, the writer
+       * skips the fill block entirely). Pin a hex value to render the
+       * tick labels in that color (e.g. `"1070CA"` for the dashboard
+       * hero blue the issue-#136 example reaches for). The 8-character
+       * `#RRGGBBAA` form is *not* accepted — alpha lives on
+       * `<a:srgbClr><a:alpha val=".."/>` which is a separate runs-level
+       * knob; pinning `labelColor` carries the RGB triple only.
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
LGTM. Reviewed the full diff and traced the implementation against the existing axis tick-label typography pipeline:

- Reader (`parseAxisLabelColor`): walks the canonical `<c:txPr><a:p><a:pPr><a:defRPr><a:solidFill><a:srgbClr>` chain, scoped strictly to the axis-level `<c:txPr>` so axis-title `<c:rich>` fills cannot leak in. Theme-color forms (`<a:schemeClr>`, `<a:hslClr>`, `<a:sysClr>`, `<a:prstClr>`) and malformed `val` tokens correctly drop to `undefined`.
- Writer (`buildAxisTxPr`): all four call sites (`buildBarAxes` x/y, `buildScatterAxes` x/y) pass the new `xLabelColor`/`yLabelColor` arg, covering bar/column/line/area/scatter (pie/doughnut have no axes per the existing routing). The `<a:defRPr>` slot expands from self-closing to wrapping `<a:solidFill>` only when a color is pinned; absence keeps the existing self-closing form so byte-for-byte output matches Excel reference for the no-color path.
- Normalizer: shares `normalizeTitleColor` with the chart-title color, so accept-with-or-without-`#`, lowercase coercion, and alpha-channel rejection (`FF0000FF` -> `undefined` via the `length !== 6` guard) are all consistent with the sibling field.
- Clone (`applyLabelColorOverride`): standard `undefined` (inherit) / `null` (drop) / value (replace) grammar matching the four sibling tick-label knobs; pie/doughnut short-circuit upstream as expected.
- Tests: parse/write/clone covers happy path, lowercase normalization, `#` stripping, malformed/non-string inputs, theme-color references, all chart families including scatter (both axes are valAx), pie/doughnut drop, end-to-end `parseChart -> cloneChart -> writeChart` and `writeXlsx` packaging, axis-title color isolation, and composition with `labelRotation`/`labelFontSize`/`labelBold`/`labelItalic` in a single `<c:txPr>` block.

CI is green. Merging.

PR #267, opened 2026-05-06
feat(xlsx): chart axis tick-label underline flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:catAx><c:txPr><a:p><a:pPr><a:defRPr u=\".."/></a:pPr></a:p></c:txPr></c:catAx>` (and the matching `<c:valAx>` slot) at the read, write, and clone layers so a template's tick-label underline pin survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `axes.x.labelUnderline?: boolean` and `axes.y.labelUnderline?: boolean` on `SheetChart` (writer side) plus the matching `labelUnderline?: boolean` on `ChartAxisInfo` (reader side). Sits on the same `<c:txPr>` body as `labelRotation` (#245), `labelFontSize` (#263), `labelBold` (#264), `labelItalic` (#265), and `labelColor` (#266); the writer emits a single block per axis carrying whichever subset is pinned. Bridges another typography-customization gap for the dashboard composition flow tracked in #136 — a templated dashboard chart whose user underlined the tick labels (e.g. an inline-link styling for the bottom-axis category names) now round-trips that flag through the parse / clone / write loop.

Modeled as a boolean for symmetry with the axis-title counterpart `axisTitleUnderline` (#262) and the chart-level `titleUnderline` (#260): `true` emits `u=\"sng\"` (Excel's UI checkbox — single line); absence collapses to omitting the attribute (the OOXML default `\"none\"` collapses to absence, mirroring how Excel's reference serialization emits a non-underlined tick label). The non-UI variant `\"dbl\"` and the sixteen exotic types (`\"words\"`, `\"heavy\"`, `\"dotted\"`, `\"dottedHeavy\"`, `\"dash\"`, `\"dashHeavy\"`, `\"dashLong\"`, `\"dashLongHeavy\"`, `\"dotDash\"`, `\"dotDashHeavy\"`, `\"dotDotDash\"`, `\"dotDotDashHeavy\"`, `\"wavy\"`, `\"wavyHeavy\"`, `\"wavyDbl\"`) are read-only — the writer emits only `\"sng\"` to keep the surfaced shape consistent with what Excel's reference UI authors, and the reader collapses every non-`\"sng\"` token to `undefined` so a templated tick label that pinned a non-single underline in raw OOXML round-trips lossless rather than silently downgrading on re-emit.

The clone layer follows the standard `boolean | null` override grammar: `undefined` inherits, `null` drops, and a literal `boolean` replaces. Non-boolean overrides (typed escapes from an untyped caller) collapse to `undefined` so the cloned `SheetChart` always carries a value the writer will accept. Pie / doughnut charts (no axes at all) silently drop the field. The flag sits on every axis flavour — `<c:catAx>` / `<c:valAx>` / `<c:dateAx>` / `<c:serAx>` all carry an optional `<c:txPr>` per the OOXML schema, so the field round-trips on every chart family that has axes (bar / column / line / area / scatter).

Refs #136

## Test plan

- [x] `pnpm test` passes (5022 tests across 89 files, lint + format + typecheck + vitest)
- [x] `pnpm build` succeeds
- [x] New describe blocks: `parseChart — axis labelUnderline` (12 cases), `writeChart — axis labelUnderline` (15 cases), `cloneChart — axis labelUnderline` (15 cases) covering parse, write, end-to-end `writeXlsx`, malformed input, the OOXML default / `\"dbl\"` / sixteen exotic underline tokens, axis-title scope isolation, pie / doughnut short-circuit, chart-type coercion, and composition with the five sibling typography knobs

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -2523,6 +2523,66 @@ export interface SheetChart {
        */
       labelColor?: string;
       /**
+       * Tick-label underline flag. Maps to
+       * `<c:catAx><c:txPr><a:p><a:pPr><a:defRPr u=".."/></a:pPr></a:p></c:txPr></c:catAx>`
+       * (or `<c:valAx>` for scatter / value axes) — Excel's "Format
+       * Axis -> Font -> Underline" toggle applied to the tick labels.
+       * The OOXML attribute is the `ST_TextUnderlineType` enum on
+       * `CT_TextCharacterProperties` (ECMA-376 Part 1, §21.1.2.3.7);
+       * the writer lands the value on the default-paragraph
+       * `<a:defRPr>` slot inside the same `<c:txPr>` block that
+       * carries {@link labelRotation} / {@link labelFontSize} /
+       * {@link labelBold} / {@link labelItalic} / {@link labelColor}.
+       *
+       * Modeled as a boolean for symmetry with {@link labelBold} /
+       * {@link labelItalic} and the axis-title counterpart
+       * {@link SheetChart.axes.x.axisTitleUnderline}: `true` emits
+       * `u="sng"` (Excel's UI checkbox — single line); absence and
+       * explicit `false` collapse to omitting the attribute (the
+       * OOXML default `"none"` collapses to absence, mirroring how
+       * Excel's reference serialization emits a non-underlined tick
+       * label). The non-UI variant `"dbl"` and the sixteen exotic
+       * types (`"words"`, `"heavy"`, `"dotted"`, `"dottedHeavy"`,
+       * `"dash"`, `"dashHeavy"`, `"dashLong"`, `"dashLongHeavy"`,
+       * `"dotDash"`, `"dotDashHeavy"`, `"dotDotDash"`,
+       * `"dotDotDashHeavy"`, `"wavy"`, `"wavyHeavy"`, `"wavyDbl"`)
+       * are read-only — the writer emits only `"sng"` to keep the
+       * surfaced shape consistent with what Excel's reference UI
+       * authors, and the reader collapses every non-`"sng"` token to
+       * `undefined` so a templated tick label that pinned a non-single
+       * underline in raw OOXML round-trips lossless rather than
+       * silently downgrading on re-emit.
+       *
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
LGTM. The implementation mirrors the established pattern for the sibling tick-label typography knobs (`labelRotation` #245, `labelFontSize` #263, `labelBold` #264, `labelItalic` #265, `labelColor` #266) and the chart/axis-title underline counterparts (`titleUnderline` #260, `axisTitleUnderline` #262):

- Reader: `parseAxisLabelUnderline` correctly walks `c:txPr -> a:p -> a:pPr -> a:defRPr`, surfaces only `"sng"` as `true`, and collapses `"none"`, `"dbl"`, the sixteen exotic `ST_TextUnderlineType` tokens, malformed values, and absence to `undefined` — matches the lossless round-trip rule (writer emits only `"sng"`, so reporting any non-single underline as `true` would silently downgrade on re-emit).
- Writer: `buildAxisTxPr` threads the new `u` attribute alongside `sz` / `b` / `i` and the optional `solidFill` child without disturbing the existing self-close vs. open-tag branching, and `normalizeAxisLabelUnderline` delegates to the chart-level `normalizeTitleUnderline` for consistent boolean coercion. `false` correctly collapses to omitting the `u` attribute (since the OOXML default `"none"` itself collapses to absence) while still emitting the `<c:txPr>` block.
- Clone: `applyLabelUnderlineOverride` follows the standard `undefined` (inherit) / `null` (drop) / boolean (replace) grammar with non-boolean fallbacks, and pie/doughnut short-circuit upstream.
- Test coverage is thorough: 42 new cases across reader, writer, clone, the OOXML default + `dbl` + sixteen exotic tokens, axis-title scope isolation, pie/doughnut, chart-type coercion, end-to-end `parseChart -> cloneChart -> writeChart -> writeXlsx`, and composition with the five sibling typography knobs.

CI is green (5022 tests, lint+format+typecheck+vitest+build). Merging.

PR #268, opened 2026-05-06
feat(xlsx): chart axis tick-label strikethrough flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:catAx><c:txPr><a:p><a:pPr><a:defRPr strike=".."/></a:pPr></a:p></c:txPr></c:catAx>` (and the matching `<c:valAx>` slot) at the read, write, and clone layers so a template's tick-label strikethrough pin survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `axes.x.labelStrike?: boolean` and `axes.y.labelStrike?: boolean` on `SheetChart` (writer side) plus the matching `labelStrike?: boolean` on `ChartAxisInfo` (reader side). Sits on the same `<c:txPr>` body as `labelRotation` (#245), `labelFontSize` (#263), `labelBold` (#264), `labelItalic` (#265), and `labelColor` (#266); the writer emits a single block per axis carrying whichever subset is pinned. Bridges another typography-customization gap for the dashboard composition flow tracked in #136 — a templated dashboard chart whose user struck through the tick labels (e.g. crossing out an obsolete bucket on a snapshot report) now round-trips that flag through the parse / clone / write loop.

Modeled as a boolean for symmetry with the axis-title counterpart `axisTitleStrike` (#261) and the chart-level `titleStrike` (#259): `true` emits `strike="sngStrike"` (Excel's UI checkbox — single line); absence and explicit `false` collapse to omitting the attribute (the OOXML default `"noStrike"` collapses to absence, mirroring how Excel's reference serialization emits a non-strikethrough tick label). The non-UI variant `"dblStrike"` is read-only — the writer emits only `"sngStrike"` to keep the surfaced shape consistent with what Excel's reference UI authors, and the reader collapses `"dblStrike"` to `undefined` so a templated tick label that pinned the double-line variant in raw OOXML round-trips lossless rather than silently downgrading on re-emit.

The clone layer follows the standard `boolean | null` override grammar: `undefined` inherits, `null` drops, and a literal `boolean` replaces. Non-boolean overrides (typed escapes from an untyped caller) collapse to `undefined` so the cloned `SheetChart` always carries a value the writer will accept. Pie / doughnut charts (no axes at all) silently drop the field. The flag sits on every axis flavour — `<c:catAx>` / `<c:valAx>` / `<c:dateAx>` / `<c:serAx>` all carry an optional `<c:txPr>` per the OOXML schema, so the field round-trips on every chart family that has axes (bar / column / line / area / scatter).

Refs #136

## Test plan

- [x] `pnpm test` passes (5021 tests across 89 files, lint + format + typecheck + vitest)
- [x] `pnpm build` succeeds
- [x] New describe blocks: `parseChart — axis labelStrike` (11 cases), `writeChart — axis labelStrike` (14 cases), `cloneChart — axis labelStrike` (15 cases) covering parse, write, end-to-end `writeXlsx`, malformed input, the OOXML default / `"dblStrike"` variant, axis-title scope isolation, pie / doughnut short-circuit, chart-type coercion, and composition with the five sibling typography knobs

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -2523,6 +2523,61 @@ export interface SheetChart {
        */
       labelColor?: string;
       /**
+       * Tick-label strikethrough flag. Maps to
+       * `<c:catAx><c:txPr><a:p><a:pPr><a:defRPr strike=".."/></a:pPr></a:p></c:txPr></c:catAx>`
+       * (or `<c:valAx>` for scatter / value axes) — Excel's "Format
+       * Axis -> Font -> Strikethrough" toggle applied to the tick
+       * labels. The OOXML attribute is the `ST_TextStrikeType` enum on
+       * `CT_TextCharacterProperties` (ECMA-376 Part 1, §21.1.2.3.7);
+       * the writer lands the value on the default-paragraph
+       * `<a:defRPr>` slot inside the same `<c:txPr>` block that
+       * carries {@link labelRotation} / {@link labelFontSize} /
+       * {@link labelBold} / {@link labelItalic} / {@link labelColor}.
+       *
+       * Modeled as a boolean for symmetry with {@link labelBold} /
+       * {@link labelItalic} and the axis-title counterpart
+       * {@link SheetChart.axes.x.axisTitleStrike}: `true` emits
+       * `strike="sngStrike"` (Excel's UI checkbox — single line);
+       * absence and explicit `false` collapse to omitting the
+       * attribute (the OOXML default `"noStrike"` collapses to absence,
+       * mirroring how Excel's reference serialization emits a non-
+       * strikethrough tick label). The non-UI variant `"dblStrike"` is
+       * read-only — the writer emits only `"sngStrike"` to keep the
+       * surfaced shape consistent with what Excel's reference UI
+       * authors, and the reader collapses `"dblStrike"` to `undefined`
+       * so a templated tick label that pinned the double-line variant
+       * in raw OOXML round-trips lossless rather than silently
+       * downgrading on re-emit.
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed the diff — the implementation mirrors the labelUnderline / labelBold / labelItalic siblings precisely:

- Reader (parseAxisLabelStrike): scopes to axis-level <c:txPr> only (verified via the dedicated test that ensures a <c:title>/<c:rich> defRPr strike doesn't leak into labelStrike), surfaces only sngStrike as true, collapses dblStrike / noStrike / malformed tokens to undefined.
- Writer (buildAxisTxPr / normalizeAxisLabelStrike): single-block emission with sz / b / i / u / strike on the same defRPr, attribute order preserved, txPr elision when no knob is pinned.
- Clone (applyLabelStrikeOverride): standard undefined / null / boolean grammar, non-boolean escapes collapse to undefined.
- Routing: line / area route through buildBarAxes('column', opts), scatter through buildScatterAxes — both wired with xLabelStrike / yLabelStrike. Pie / doughnut have no axis builder at all so labelStrike is silently dropped, consistent with the doc and exercised by the dedicated tests.
- Tests: 11 + 14 + 15 cases cover the absence/default round-trip, the dblStrike lossless drop, the title-vs-tick scope isolation, multi-axis composition with the five sibling typography knobs, chart-family threading, the OOXML schema-order placement (between tickLblPos and crossAx), and the end-to-end writeXlsx package flow.

CI green. LGTM.

PR #250, opened 2026-05-06
feat(ods): numFmt data styles for number formatting

productdevbook:
## Summary

The ODS writer ignored `CellStyle.numFmt` — every numeric value
serialised as a raw float with `office:value-type="float"`, so format
codes like `[HH]:MM`, `0.00%`, or currency patterns silently dropped.
This wires `numFmt` through both the writer and the reader so cells
carry their format through ODS roundtrips.

The writer translates Excel-style format codes into the ODS
`urn:oasis:names:tc:opendocument:xmlns:datastyle:1.0` namespace and
emits one of the five spec'd data-style flavours inside
`<office:automatic-styles>`:

- `<number:number-style>` for plain numeric formats (decimals,
  thousands grouping)
- `<number:percentage-style>` for `0%` / `0.00%`
- `<number:currency-style>` for `$#,##0.00`, `[$$-409]#,##0.00`, etc.
- `<number:date-style>` for `yyyy-mm-dd`, `dd/mm/yyyy hh:mm`, etc.
- `<number:time-style>` for clock and `[HH]:MM` elapsed-time formats —
  the elapsed flavour gets `number:truncate-on-overflow="false"` per
  ODS-1.2

Each cell's `<style:style>` element references the data style through
`style:data-style-name`, and the writer emits all data styles before
the `<style:style>` elements that reference them so parsers see the
names in document order. Identical `numFmt` codes deduplicate to a
single data-style entry.

The reader's `readStyles` path runs an extra first pass over
`<office:automatic-styles>` to collect data-style definitions and
reverse-translates them into Excel-compatible format codes, so a
`numFmt` set on a cell roundtrips intact.

Closes #249

## API

```ts
import { writeOds, readOds } from "hucre/ods";

const data = await writeOds({
  sheets: [{
    name: "Sheet1",
    rows: [["Duration", 0.020833]],
    cells: new Map([
      ["1,1", { style: { numFmt: "[HH]:MM" } }],
    ]),
  }],
});

// content.xml now contains:
// <number:time-style style:name="N100" number:truncate-on-overflow="false">
//   <number:hours number:style="long"/>
//   <number:text>:</number:text>
//   <number:minutes number:style="long"/>
// </number:time-style>
// <style:style style:name="ce1" style:family="table-cell"
//              style:data-style-name="N100"/>

const wb = await readOds(data, { readStyles: true });
wb.sheets[0].cells!.get("1,1")?.style?.numFmt;
// → "[hh]:mm"
```

## Test plan

[...]

PR #269, opened 2026-05-06
feat(xlsx): chart legend font size — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:legend><c:txPr><a:p><a:pPr><a:defRPr sz="N"/></a:pPr></a:p></c:txPr></c:legend>` at the read, write, and clone layers so a template's legend font-size pin survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `legendFontSize?: number` on `SheetChart` (writer side) and the matching `legendFontSize?: number` on `Chart` (reader side). Sits on the same `<c:legend>` element as `legend` / `legendOverlay` / `legendEntries`; the writer emits a single `<c:txPr>` block on the legend whenever the field is set. Bridges another typography-customization gap for the dashboard composition flow tracked in #136 — a templated dashboard chart whose user pinned a custom legend size (e.g. shrinking the swatch labels to fit a tight sidebar slot) now round-trips that pin through the parse / clone / write loop.

Modeled in points (whole / half) for symmetry with the chart-title `titleFontSize` (#250-ish), the axis-title `axisTitleFontSize` (#255), and the axis tick-label `labelFontSize` (#263): same `1..400`pt band, same 0.5pt half-step rounding (Excel's UI granularity), same OOXML 100ths-of-a-point conversion factor. Out-of-range / non-finite / non-numeric inputs collapse to `undefined` so the writer drops the `<c:txPr>` block entirely and the rendered legend falls back to Excel's theme-default 9pt.

The clone layer follows the standard `number | null` override grammar: `undefined` inherits, `null` drops, and a literal number replaces. Non-numeric overrides (typed escapes from an untyped caller) collapse to `undefined` so the cloned `SheetChart` always carries a value the writer will accept. A hidden legend (`legend === false`) silently drops the inherited / overridden value — there is no `<c:txPr>` slot to host the size on a hidden legend, so leaking the value into the output would carry a pin Excel never reads.

The `<c:txPr>` block lands after `<c:overlay>` and before the optional `<c:extLst>`, matching the CT_Legend schema sequence (ECMA-376 Part 1, §21.2.2.114). The `<a:bodyPr>` is emitted without a `rot` attribute — the legend is not rotatable in Excel's UI, so the writer keeps the body-properties element minimal and only carries the `<a:defRPr sz="N"/>` slot inside the paragraph stub. The flag rides on every chart family that emits a legend (bar / column / line / area / scatter / pie / doughnut).

Refs #136

## Test plan

- [x] `pnpm test` passes (5118 tests across 89 files, lint + format + typecheck + vitest)
- [x] `pnpm build` succeeds
- [x] New describe blocks: `parseChart — legendFontSize` (15 cases), `writeChart — legendFontSize` (17 cases), `cloneChart — legendFontSize` (20 cases) covering parse, write, end-to-end `writeXlsx`, half-step rounding, out-of-range / non-numeric / non-finite drops, hidden-legend scoping, OOXML element ordering, multi-family coverage (bar / column / line / pie / doughnut / area / scatter), and composition with the sibling legend knobs (`legendOverlay` / `legendEntries`)

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1289,6 +1289,43 @@ export interface SheetChart {
    * emits in that order so a re-parse sees the canonical sequence.
    */
   legendEntries?: ChartLegendEntry[];
+  /**
+   * Legend font size in whole or half points. Maps to
+   * `<c:legend><c:txPr><a:p><a:pPr><a:defRPr sz="N"/></a:pPr></a:p>
+   * </c:txPr></c:legend>` — Excel's "Format Legend -> Font -> Size"
+   * knob. The OOXML attribute is in 100ths of a point, so 12pt
+   * serializes as `sz="1200"` and 9pt (Excel's reference default for
+   * the legend) as `sz="900"`; the writer performs the conversion at
+   * emit time and lands the value on the default-paragraph
+   * `<a:defRPr>` slot inside the legend's `<c:txPr>` block so a
+   * re-parse picks the size up off the canonical slot the OOXML schema
+   * exposes.
+   *
+   * Accepted range: `1..400`pt (the band the OOXML `ST_TextFontSize`
+   * schema exposes — `100..400000` in 100ths of a point). Fractional
+   * inputs round to the nearest 0.5pt (Excel's UI granularity); inputs
+   * outside the band, `NaN`, `Infinity`, and non-numeric inputs all
+   * collapse to `undefined` so the writer drops the `<c:txPr>` block
+   * rather than emit a token Excel would reject — the rendered legend
+   * falls back to the theme-default 9pt.
+   *
+   * Default: omitted — the legend renders at Excel's reference 9pt
+   * (the OOXML schema's application-default for chart legend text). Set
+   * an explicit value to scale the legend up for a hero dashboard tile
+   * (e.g. `14`) or down to fit a tight sidebar slot (e.g. `7`).
+   * Silently ignored when `legend === false` (no `<c:legend>` element
+   * is emitted) — there is no slot to host the size in that case.
+   *
+   * Mirrors {@link titleFontSize} / {@link axes.x.axisTitleFontSize} /
+   * {@link axes.x.labelFontSize} — same range, same normalization,
+   * same OOXML conversion factor — so a caller can thread a single
+   * size value through every typography-pinning slot without bookkeeping
+   * the units. Composes independently with {@link legend} /
+   * {@link legendOverlay} / {@link legendEntries}: all four fields land
+   * on the same `<c:legend>` element so a single configuration call
+   * threads cleanly through every legend knob Excel exposes.
+   */
+  legendFontSize?: number;
   /** Show the chart-level title element. Default: `true` when `title` is set. */
   showTitle?: boolean;
   /**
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed the diff end-to-end.

Reader (`parseChart`): `parseLegendFontSize` walks `legend → txPr → p → pPr → defRPr` with proper nil-bailouts at every link, parses the `sz` attribute as a base-10 int, finite-checks, and converts via the shared `TITLE_FONT_SZ_PER_POINT` half-step normalizer that mirrors the chart-title / axis-title / tick-label parsers — the conversion factor and the `Math.round(parsed/100*2)/2` pattern are byte-identical, so the typography readers compose without unit-bookkeeping. Range gating uses the same `TITLE_FONT_SIZE_MIN_PT` / `TITLE_FONT_SIZE_MAX_PT` band. The hidden-legend scoping (`legend === false` => no surface) lives in the call-site guard inside `parseChart` (the `if (legendPos)` block, mirroring `legendOverlay` / `legendEntries`), which is correct and exercised by the `<c:delete val="1"/>` test.

Writer (`writeChart`): `buildLegend` signature gains `fontSizePt: number | undefined` as a fourth positional arg. `<c:txPr>` lands after `<c:overlay>` per CT_Legend (`legendPos? legendEntry* layout? overlay? spPr? txPr? extLst?`), with an empty `<a:bodyPr/>` (no `rot` — legend is not rotatable in Excel's UI), `<a:lstStyle/>`, and `<a:p><a:pPr><a:defRPr sz=N/></a:pPr><a:endParaRPr lang="en-US"/></a:p>` paragraph stub. `buildLegendTxPr` returns `undefined` when no knob is pinned so a fresh legend matches Excel's reference serialization. `resolveLegendFontSize` runs the input through the shared `normalizeTitleFontSize` so out-of-range / non-finite / non-numeric / fractional escapes collapse cleanly.

Clone: `resolveLegendFontSize` follows the canonical `undefined → inherit (after normalize) / null → drop / number → replace (after normalize)` grammar. The hidden-legend scoping is handled by the surrounding `if (legendPos)` block (same scoping that drops `legendOverlay` / `legendEntries`), which is the correct guard.

Tests: parse 15 cases (sz=900 default, half-step rounding cells at 924 / 930 / 950 / 980, missing chains at every link, `<c:delete val="1"/>` hidden-legend drop, range edges at sz=49 / 100 / 40000 / 400001, non-numeric / empty tokens, multi-family coverage); write 17 cases (sz=1200 / 1400 / 1050 half-step, OOXML ordering vs `<c:overlay>` / `<c:legendPos>`, single-emission, hidden-legend silent drop, every chart family including scatter, range / non-finite / non-numeric drops, the `<a:bodyPr/>` no-rot guard, parse-back round-trip, composition with `legendOverlay` / `legendEntries`, end-to-end `writeXlsx`); clone 20 cases (inherit / override / null / absent, half-step rounding on both override and inherit paths, range / non-finite / non-numeric drops, family flatten through column and doughnut, hidden-legend drop on both inherit and override paths, override re-enabling a hidden source legend, composition, three end-to-end `writeXlsx` paths). Coverage is thorough.

CI is green (build success). The signature change to `buildLegend` is consistent with the other typography knobs landing on the same element. LGTM.

PR #271, opened 2026-05-06
feat(xlsx): chart legend italic flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:legend><c:txPr><a:p><a:pPr><a:defRPr i=".."/></a:pPr></a:p></c:txPr></c:legend>` at the read, write, and clone layers so a template's legend italic pin survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `legendItalic?: boolean` on `SheetChart` (writer side) and the matching `legendItalic?: boolean` on `Chart` (reader side). Sits on the same `<c:legend>` element as `legend` / `legendOverlay` / `legendEntries`; the writer emits a single `<c:txPr>` block on the legend whenever the field is set. Bridges another typography-customization gap for the dashboard composition flow tracked in #136 — a templated dashboard chart whose user pinned a custom legend slant (e.g. italicizing the legend swatch labels for a quote-style dashboard tile) now round-trips that pin through the parse / clone / write loop.

Modeled as a boolean for symmetry with the chart-title `titleItalic` (#253), the axis-title `axisTitleItalic` (#257), and the axis tick-label `labelItalic` (#265): `true` emits `i="1"`; `false` emits the OOXML default `i="0"` explicitly so a clone target can override an upstream `i="1"` from a templated chart; absence collapses to omitting the entire `<c:txPr>` block. The reader collapses the OOXML default `false` to `undefined` so absence and `i="0"` round-trip identically — only an explicit `i="1"` surfaces `true`.

The clone layer follows the standard `boolean | null` override grammar: `undefined` inherits, `null` drops, and a literal `boolean` replaces. Non-boolean overrides (typed escapes from an untyped caller) collapse to `undefined` so the cloned `SheetChart` always carries a value the writer will accept. A hidden legend (`legend === false`) silently drops the inherited / overridden value — there is no `<c:txPr>` slot to host the flag on a hidden legend, so leaking the value into the output would carry a pin Excel never reads.

The `<c:txPr>` block lands after `<c:overlay>` and before the optional `<c:extLst>`, matching the CT_Legend schema sequence (ECMA-376 Part 1, §21.2.2.114). The `<a:bodyPr>` is emitted without a `rot` attribute — the legend is not rotatable in Excel's UI, so the writer keeps the body-properties element minimal and only carries the `<a:defRPr i=".."/>` slot inside the paragraph stub. The flag rides on every chart family that emits a legend (bar / column / line / area / scatter / pie / doughnut).

Refs #136

## Test plan

- [x] \`pnpm test\` passes (5107 tests across 89 files, lint + format + typecheck + vitest)
- [x] \`pnpm build\` succeeds
- [x] New describe blocks: \`parseChart — legendItalic\` (12 cases), \`writeChart — legendItalic\` (16 cases), \`cloneChart — legendItalic\` (17 cases) covering parse, write, end-to-end \`writeXlsx\`, the OOXML default \`false\` collapse, malformed input, hidden-legend scoping, OOXML element ordering, multi-family coverage (bar / column / line / pie / doughnut / area / scatter), and composition with the sibling legend knobs (\`legendOverlay\` / \`legendEntries\`)

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1356,6 +1356,38 @@ export interface SheetChart {
    * cleanly through every legend knob Excel exposes.
    */
   legendBold?: boolean;
+  /**
+   * Legend italic flag. Maps to `<c:legend><c:txPr><a:p><a:pPr>
+   * <a:defRPr i=".."/></a:pPr></a:p></c:txPr></c:legend>` — Excel's
+   * "Format Legend -> Font -> Italic" toggle. The OOXML attribute is
+   * the `xsd:boolean` `i` on `CT_TextCharacterProperties` (ECMA-376
+   * Part 1, §21.1.2.3.7); the writer lands the value on the
+   * default-paragraph `<a:defRPr>` slot inside the legend's
+   * `<c:txPr>` block so a re-parse picks the flag up off the canonical
+   * slot the OOXML schema exposes.
+   *
+   * Default: omitted — the legend renders non-italic (no `i`
+   * attribute, matching Excel's reference serialization for a fresh
+   * chart legend whose typography has not been customized; the OOXML
+   * default `false` collapses to absence). Set `true` to emit `i="1"`
+   * so the legend renders italic; set `false` explicitly to pin the
+   * non-default `i="0"` (functionally identical to omission, but
+   * useful when overriding a templated legend that had italic pinned
+   * upstream).
+   *
+   * Silently ignored when `legend === false` (no `<c:legend>` element
+   * is emitted) — there is no `<c:txPr>` slot to host the flag in
+   * that case. Mirrors {@link titleItalic} /
+   * {@link axes.x.axisTitleItalic} / {@link axes.x.labelItalic} —
+   * same boolean-with-explicit-default shape, same OOXML
+   * `<a:defRPr i=".."/>` mapping — so a caller can thread a single
+   * italic value through every typography-pinning slot. Composes
+   * independently with {@link legend} / {@link legendOverlay} /
+   * {@link legendEntries}: all four fields land on the same
+   * `<c:legend>` element so a single configuration call threads
+   * cleanly through every legend knob Excel exposes.
+   */
+  legendItalic?: boolean;
   /** Show the chart-level title element. Default: `true` when `title` is set. */
   showTitle?: boolean;
   /**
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed the diff end-to-end after the PR #269 / #270 rebase.

Reader (`parseChart`): `parseLegendItalic` mirrors `parseLegendBold` exactly — same `legend → txPr → p → pPr → defRPr` walk with nil-bailouts at every link. The OOXML truthy/falsy collapse is the canonical pattern: `i="1"` / `i="true"` surface `true`; `i="0"` / `i="false"` / unknown / missing all collapse to `undefined` so the OOXML default and an explicit `i="0"` round-trip identically through `cloneChart`. The hidden-legend scoping (`legend === false` => no surface) lives in the call-site guard inside `parseChart` (the `if (legendPos)` block, mirroring the rest of the legend-typography knobs), exercised by the `<c:delete val="1"/>` test.

Writer (`writeChart`): the `buildLegend` signature now accumulates three typography knobs (`fontSizePt`, `bold`, `italic`) — the natural extension of the pattern landed during PR #270's rebase. `buildLegendTxPr` returns `undefined` only when all three inputs are unset; when any knob is pinned the block emits with the appropriate attrs (`sz` and/or `b` and/or `i`). `<c:txPr>` lands after `<c:overlay>` per CT_Legend, with the standard `<a:bodyPr/>` (no `rot`) / `<a:lstStyle/>` / `<a:p><a:pPr><a:defRPr.../>` / `<a:endParaRPr lang="en-US"/>` shape. `resolveLegendItalic` is the canonical pass-through (only literal `true` / `false` survive; non-boolean tokens drop).

Clone: `resolveLegendItalic` follows the canonical `undefined → inherit (after normalize) / null → drop / boolean → replace (after normalize)` grammar. The `normalizeLegendItalic` helper passes through literal booleans and drops everything else (including non-boolean source values via typed escapes). Hidden-legend scoping handled by the surrounding `if (legendPos)` block (same scoping that drops `legendOverlay` / `legendEntries` / `legendFontSize` / `legendBold`).

Tests: parse 12 cases (truthy / falsy spellings, missing chains at every link, hidden-legend drop, multi-family coverage, malformed token drops); write 16 cases (i=1 / i=0 emit, OOXML ordering, single-emission, hidden-legend silent drop, every chart family, non-boolean / null drops, the `<a:bodyPr/>` no-rot guard, parse-back round-trip including the `i="0"` collapse, composition with `legendOverlay` / `legendEntries`, end-to-end `writeXlsx`); clone 17 cases (inherit / override / null / absent, non-boolean drops on both override and source paths, family flatten through column and doughnut, hidden-legend drop on both inherit and override paths, override re-enabling a hidden source legend, composition, three end-to-end `writeXlsx` paths).

I rebased the branch onto main after #269 and #270 merged, resolved the expected conflicts in the shared `buildLegend` / `<c:txPr>` code paths by extending the signature to accumulate all three knobs (`fontSizePt`, `bold`, `italic`), and verified locally with `pnpm vitest run --dir=test` (5206 passed) and `pnpm build` (success). LGTM.

PR #272, opened 2026-05-06
feat(xlsx): chart legend underline flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:legend><c:txPr><a:p><a:pPr><a:defRPr u=\"..\"/></a:pPr></a:p></c:txPr></c:legend>` at the read, write, and clone layers so a template's legend underline pin survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `legendUnderline?: boolean` on `SheetChart` (writer side) and the matching `legendUnderline?: boolean` on `Chart` (reader side). Sits on the same `<c:legend>` element as `legend` / `legendOverlay` / `legendEntries` / `legendFontSize`; the writer emits a single `<c:txPr>` block on the legend whenever any typography knob is pinned, and the underline value lands on the same `<a:defRPr>` slot as the size. Bridges another typography-customization gap for the dashboard composition flow tracked in #136 — a templated dashboard chart whose user pinned a custom legend underline (e.g. underlining the legend swatch labels for a hyperlinked dashboard tile) now round-trips that pin through the parse / clone / write loop.

Modeled as a boolean for symmetry with the chart-title `titleUnderline` (#260), the axis-title `axisTitleUnderline` (#262), and the axis tick-label `labelUnderline` (#267): `true` emits `u=\"sng\"` (Excel's UI variant — single underline); `false` emits the OOXML default `u=\"none\"` explicitly so a clone target can override an upstream `u=\"sng\"` from a templated chart; absence collapses to omitting the `u` attribute entirely. The reader collapses the OOXML default `\"none\"` (and every non-`\"sng\"` variant — `\"dbl\"`, `\"heavy\"`, `\"dotted\"`, `\"dotDash\"`, `\"wavy\"`, etc.) to `undefined` so absence and `u=\"none\"` round-trip identically — only an explicit `u=\"sng\"` surfaces `true`. Reporting any non-single underline as `true` would silently downgrade the choice to a single line on round-trip; the writer emits only `\"sng\"` / `\"none\"`.

The clone layer follows the standard `boolean | null` override grammar: `undefined` inherits, `null` drops, and a literal `boolean` replaces. Non-boolean overrides (typed escapes from an untyped caller) collapse to `undefined` so the cloned `SheetChart` always carries a value the writer will accept. A hidden legend (`legend === false`) silently drops the inherited / overridden value — there is no `<c:txPr>` slot to host the flag on a hidden legend, so leaking the value into the output would carry a pin Excel never reads.

The `<c:txPr>` block lands after `<c:overlay>` and before the optional `<c:extLst>`, matching the CT_Legend schema sequence (ECMA-376 Part 1, §21.2.2.114). The `<a:bodyPr>` is emitted without a `rot` attribute — the legend is not rotatable in Excel's UI, so the writer keeps the body-properties element minimal. The flag rides on every chart family that emits a legend (bar / column / line / area / scatter / pie / doughnut). Composes with the sibling `legendFontSize` (#269) — both attributes land on the same `<a:defRPr>` slot when both are pinned.

Refs #136

## Test plan

- [x] `pnpm test` passes (5164 tests across 89 files, lint + format + typecheck + vitest)
- [x] `pnpm build` succeeds
- [x] New describe blocks: `parseChart — legendUnderline` (12 cases), `writeChart — legendUnderline` (17 cases), `cloneChart — legendUnderline` (17 cases) covering parse, write, end-to-end `writeXlsx`, the OOXML default `\"none\"` collapse, the non-`\"sng\"` variant collapse, malformed input, hidden-legend scoping, OOXML element ordering, multi-family coverage (bar / column / line / pie / doughnut / area / scatter), and composition with the sibling legend knobs (`legendOverlay` / `legendEntries` / `legendFontSize`)

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1388,6 +1388,39 @@ export interface SheetChart {
    * cleanly through every legend knob Excel exposes.
    */
   legendItalic?: boolean;
+  /**
+   * Legend underline flag. Maps to `<c:legend><c:txPr><a:p><a:pPr>
+   * <a:defRPr u=".."/></a:pPr></a:p></c:txPr></c:legend>` — Excel's
+   * "Format Legend -> Font -> Underline" toggle. The OOXML attribute is
+   * the `ST_TextUnderlineType` enumeration on
+   * `CT_TextCharacterProperties` (ECMA-376 Part 1, §21.1.2.3.7); the
+   * writer lands `u="sng"` (Excel's UI variant — single underline) or
+   * `u="none"` on the default-paragraph `<a:defRPr>` slot inside the
+   * legend's `<c:txPr>` block so a re-parse picks the flag up off the
+   * canonical slot the OOXML schema exposes.
+   *
+   * Default: omitted — the legend renders non-underlined (no `u`
+   * attribute, matching Excel's reference serialization for a fresh
+   * chart legend whose typography has not been customized; the OOXML
+   * default `"none"` collapses to absence). Set `true` to emit
+   * `u="sng"` so the legend renders single-underlined; set `false`
+   * explicitly to pin the OOXML default `u="none"` (functionally
+   * identical to omission, but useful when overriding a templated
+   * legend that had underline pinned upstream).
+   *
+   * Silently ignored when `legend === false` (no `<c:legend>` element
+   * is emitted) — there is no `<c:txPr>` slot to host the flag in that
+   * case. Mirrors {@link titleUnderline} /
+   * {@link axes.x.axisTitleUnderline} / {@link axes.x.labelUnderline}
+   * — same boolean-with-explicit-default shape, same OOXML
+   * `<a:defRPr u=".."/>` mapping — so a caller can thread a single
+   * underline value through every typography-pinning slot. Composes
+   * independently with {@link legend} / {@link legendOverlay} /
+   * {@link legendEntries} / {@link legendFontSize}: all five fields
+   * land on the same `<c:legend>` element so a single configuration
+   * call threads cleanly through every legend knob Excel exposes.
+   */
+  legendUnderline?: boolean;
   /** Show the chart-level title element. Default: `true` when `title` is set. */
   showTitle?: boolean;
   /**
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed the diff end-to-end after the PR #271 rebase.

Reader (`parseChart`): `parseLegendUnderline` walks the canonical `legend → txPr → p → pPr → defRPr` chain with nil-bailouts, then surfaces `true` only when `u="sng"` (Excel's UI variant — single underline). The OOXML default `"none"` and every non-`"sng"` variant the schema allows (`"dbl"`, `"heavy"`, `"dotted"`, `"dotDash"`, `"wavy"`, etc.) collapse to `undefined` — the right call, since the writer only emits `"sng"` / `"none"` and reporting any non-single underline as `true` would silently downgrade the choice to a single line on round-trip. Hidden-legend scoping uses the standard `if (legendPos)` call-site guard, exercised by the `<c:delete val="1"/>` test.

Writer (`writeChart`): `buildLegend` now accumulates four typography knobs (`fontSizePt`, `bold`, `italic`, `underline`) — the natural extension of the pattern. `buildLegendTxPr` returns `undefined` only when all four are unset; when any knob is pinned the block emits with the appropriate attrs accumulated on a single `<a:defRPr>` slot. `true` writes `u="sng"`, `false` writes `u="none"` — `false` pins the OOXML default explicitly (functionally identical to absence but lets a clone target override an upstream `u="sng"`). `<c:txPr>` lands after `<c:overlay>` per CT_Legend, with the standard `<a:bodyPr/>` (no `rot`) / `<a:lstStyle/>` / paragraph-stub shape. `resolveLegendUnderline` is the canonical pass-through.

Clone: `resolveLegendUnderline` follows the canonical `undefined → inherit (after normalize) / null → drop / boolean → replace (after normalize)` grammar. `normalizeLegendUnderline` passes through literal booleans and drops everything else. Hidden-legend scoping handled by the surrounding `if (legendPos)` block.

Tests: parse 12 cases (`u="sng"` surface, OOXML default `"none"` collapse, missing chains at every link, hidden-legend drop, multi-family coverage, the full non-`"sng"` variant matrix collapse including `"dbl"` / `"heavy"` / all the dotted / dashed / wavy variants, garbage / unknown tokens, composition with `legendOverlay` / `legendEntries` / `legendFontSize`); write 17 cases (sng / none emit, OOXML ordering, single-emission, hidden-legend silent drop, every chart family, non-boolean / null / number drops, the `<a:bodyPr/>` no-rot guard, parse-back round-trip including the `u="none"` collapse, composition with sibling knobs, the co-emission test that verifies `sz` and `u` land on a single `<a:defRPr>` slot when both are pinned, end-to-end `writeXlsx`); clone 17 cases (inherit / override / null / absent, explicit `false` pins through, non-boolean / numeric drops on both override and source paths, family flatten through column and doughnut, hidden-legend drop on both inherit and override paths, override re-enabling a hidden source legend, composition, three end-to-end `writeXlsx` paths).

I rebased the branch onto main after #269 / #270 / #271 merged, resolved the expected conflicts in the shared `buildLegend` / `<c:txPr>` code paths by extending the signature to accumulate all four typography knobs (`fontSizePt`, `bold`, `italic`, `underline`), and verified locally with `pnpm vitest run --dir=test` (5252 passed) and `pnpm build` (success). LGTM.

PR #276, opened 2026-05-06
feat(xlsx): chart data labels font size — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:dLbls><c:txPr><a:p><a:pPr><a:defRPr sz=\"N\"/></a:pPr></a:p></c:txPr></c:dLbls>` at the read, write, and clone layers so a template's data-labels font size survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `fontSize?: number` on `ChartDataLabels` (writer side) and the matching `fontSize?: number` on `ChartDataLabelsInfo` (reader side). Sits on the same `<c:dLbls>` block as `position` / `numberFormat` / `separator` / `showLeaderLines` / the `show*` toggles; the writer threads the value through the shared `buildDataLabelsBody` helper so chart-level and per-series data labels both pick up the typography pin. Bridges another typography-customization gap for the dashboard composition flow tracked in #136 — a templated dashboard chart whose user pinned a custom data-labels font size (e.g. larger labels for emphasis on the \"primary\" KPI tile) now round-trips that value through the parse / clone / write loop.

Mirrors the chart-title `titleFontSize` (#251), the axis-title `axisTitleFontSize` (#255), the axis tick-label `labelFontSize` (#263), and the legend `legendFontSize` (#269) exactly — same range (`1..400` points), same fractional rounding (Excel's UI step is 0.5pt — `12.3` → `12.5`, `12.24` → `12`), same OOXML conversion factor (×100 at emit time, ÷100 at parse time). Out-of-range values and non-numeric tokens (typed escapes from an untyped caller — strings, `null`, `NaN`, `Infinity`) collapse to `undefined` so the writer never emits a malformed `sz` attribute and absence + malformed source value round-trip identically through `cloneChart`.

The `<c:txPr>` block lands between `<c:numFmt>` and `<c:dLblPos>` inside `<c:dLbls>`, matching the CT_DLbls schema sequence (ECMA-376 Part 1, §21.2.2.50). The block is skipped entirely when no font knob is pinned so a fresh chart matches Excel's reference shape byte-for-byte (Excel itself omits `<c:txPr>` from `<c:dLbls>` when the labels render at the theme-default style). The clone path threads the value through the existing `resolveChartDataLabels` resolver (which spreads the read-side `ChartDataLabelsInfo` into the write-side `ChartDataLabels` shape — they share field semantics). Future typography pins (color, bold, italic, underline, strikethrough) will land on the same `<a:defRPr>` slot.

Refs #136

## Test plan

- [x] `pnpm test` passes (5336 tests across 89 files, lint + format + typecheck + vitest)
- [x] `pnpm build` succeeds
- [x] New describe blocks: `parseChart — data labels fontSize` (10 cases), `writeChart — dataLabels.fontSize` (17 cases), `cloneChart — dataLabels.fontSize` (12 cases) covering parse, write, end-to-end `writeXlsx`, fractional rounding, out-of-range / NaN / Infinity drops, malformed input drops, OOXML element ordering (`<c:numFmt>` -> `<c:txPr>` -> `<c:dLblPos>`), multi-family coverage (bar / column / line / pie / doughnut / area), composition with the sibling dLbls knobs (`position` / `numberFormat` / `separator` / `show*` toggles), and the standard `null`-drop / explicit-replace clone grammar

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -654,6 +654,39 @@ export interface ChartDataLabels {
    * out and render labels without the connecting lines.
    */
   showLeaderLines?: boolean;
+  /**
+   * Data-label font size in points (range `1..400`), pinned via
+   * `<c:dLbls><c:txPr><a:p><a:pPr><a:defRPr sz="N"/></a:pPr></a:p>
+   * </c:txPr></c:dLbls>`. Mirrors Excel's "Format Data Labels -> Font
+   * -> Size" knob — the OOXML attribute is in 100ths of a point on
+   * `CT_TextCharacterProperties`' `sz` slot (ECMA-376 Part 1,
+   * §21.1.2.3.7); the writer holds the value in points and converts
+   * at emit time so a caller can pass the value directly without
+   * doing the multiplication.
+   *
+   * Default: omitted — the data labels render at the theme-default
+   * size (Excel's reference behavior for fresh data labels whose
+   * typography has not been customized; the writer skips the entire
+   * `<c:txPr>` block when no font knob is pinned).
+   *
+   * Out-of-range values (`< 1` or `> 400`) and non-numeric tokens
+   * (typed escapes from an untyped caller — strings, `null`, `NaN`,
+   * `Infinity`) collapse to `undefined` so the writer never emits a
+   * malformed `sz` attribute. Fractional points round to the nearest
+   * half-point (Excel's UI step) — `12.3` → `12.5`, `12.24` → `12`.
+   *
+   * The `<c:txPr>` block lands between `<c:spPr>` and `<c:dLblPos>`,
+   * matching the CT_DLbls schema sequence (ECMA-376 Part 1,
+   * §21.2.2.50). Composes independently with the other dLbls knobs —
+   * {@link position} / {@link separator} / {@link numberFormat} /
+   * {@link showLeaderLines} / the `show*` toggles — so a caller can
+   * pin the font size without touching the rest of the configuration.
+   * Mirrors {@link SheetChart.titleFontSize} /
+   * {@link SheetChart.legendFontSize} — same range, same conversion
+   * factor — so a caller can thread a single point value through every
+   * typography-pinning slot.
+   */
+  fontSize?: number;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


productdevbook (commented):
Reviewed the diff. The implementation faithfully follows the established typography-pin pattern (chart title #251, axis title #255, tick label #263, legend #269):

- Reader `parseDataLabelsFontSize` walks `<c:dLbls><c:txPr><a:p><a:pPr><a:defRPr sz="N"/>` and surfaces the size in points after dividing by 100.
- Writer threads `resolveDataLabelsFontSize` (delegates to `normalizeTitleFontSize` so the range / fractional rounding / OOXML conversion all match the chart-title / axis-title / tick-label / legend resolvers).
- New `buildDataLabelsTxPr` helper emits the canonical `<c:txPr>` stub (`<a:bodyPr/>` / `<a:lstStyle/>` / `<a:p><a:pPr><a:defRPr sz="N"/></a:pPr><a:endParaRPr/></a:p>`) and is skipped entirely when no font knob is pinned so a fresh chart matches Excel's reference shape.
- The `<c:txPr>` block lands between `<c:numFmt>` and `<c:dLblPos>` — the comment says "between `<c:spPr>` and `<c:dLblPos>`" but the writer currently skips `<c:spPr>`, so in practice it sits after `<c:numFmt>` and the test assertion (`numFmt < txPr < dLblPos`) confirms the shipped order is correct per CT_DLbls (ECMA-376 Part 1, §21.2.2.50).

One consistency fix I pushed in a follow-up commit: `parseDataLabelsFontSize` was using hardcoded `100` / `1` / `400` literals, missing the empty-string / whitespace trim guard, and not rounding parsed values to the 0.5pt UI step that every sibling reader applies (`parseTitleFontSize` / `parseAxisTitleFontSize` / `parseLegendFontSize` all use `TITLE_FONT_SZ_PER_POINT` / `TITLE_FONT_SIZE_MIN_PT` / `TITLE_FONT_SIZE_MAX_PT` and the `Math.round((parsed / 100) * 2) / 2` half-step quantization). Pushed the alignment so a hand-edited template with `sz="1234"` surfaces 12.5 here just as it would on the chart-title slot.

Verification on the rebased branch:
- `pnpm lint` clean
- `pnpm typecheck` clean
- `pnpm vitest run --dir=test` green — 5336 tests across 89 files (39 cases on the dataLabels.fontSize surfaces)

LGTM after CI confirms the follow-up commit.

PR #277, opened 2026-05-06
feat(xlsx): chart data labels font color — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:dLbls><c:txPr><a:p><a:pPr><a:defRPr><a:solidFill><a:srgbClr val=\"RRGGBB\"/></a:solidFill></a:defRPr></a:pPr></a:p></c:txPr></c:dLbls>` at the read, write, and clone layers so a template's data-labels font color survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `fontColor?: string` on `ChartDataLabels` (writer side) and the matching `fontColor?: string` on `ChartDataLabelsInfo` (reader side). Sits on the same `<c:dLbls>` block as `position` / `numberFormat` / `separator` / `showLeaderLines` / the `show*` toggles; the writer threads the value through the shared `buildDataLabelsBody` helper so chart-level and per-series data labels both pick up the typography pin. Bridges another typography-customization gap for the dashboard composition flow tracked in #136 — a templated dashboard chart whose user tinted the data labels (e.g. a high-contrast hex on a dark-background dashboard tile) now round-trips that color through the parse / clone / write loop.

Mirrors the chart-title `titleColor` (#254), the axis-title `axisTitleColor` (#258), the axis tick-label `labelColor` (#266), and the legend `legendFontColor` exactly — same accept-with-or-without-`#` hex grammar (`\"FF0000\"` / `\"#FF0000\"` / `\"ff0000\"` all collapse to the OOXML uppercase canonical form `\"FF0000\"`). Malformed inputs (wrong length, non-hex characters, alpha-channel forms like `\"#FF0000FF\"`, non-string escapes from an untyped caller) collapse to `undefined` so the writer skips the entire `<a:solidFill>` block and the labels inherit the theme text color (Excel's reference behavior for fresh data labels that have not had a custom color picked). Theme references (`<a:schemeClr>`) and preset color tokens drop on the reader side too — only the literal RGB triple round-trips losslessly through the writer.

[...]

PR #278, opened 2026-05-06
feat(xlsx): chart data labels bold flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:dLbls><c:txPr><a:p><a:pPr><a:defRPr b=\"..\"/></a:pPr></a:p></c:txPr></c:dLbls>` at the read, write, and clone layers so a template's data-labels bold flag survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `bold?: boolean` on `ChartDataLabels` (writer side) and the matching `bold?: boolean` on `ChartDataLabelsInfo` (reader side). Sits on the same `<c:dLbls>` block as `position` / `numberFormat` / `separator` / `showLeaderLines` / the `show*` toggles; the writer threads the value through the shared `buildDataLabelsBody` helper so chart-level and per-series data labels both pick up the typography pin. Bridges another typography-customization gap for the dashboard composition flow tracked in #136 — a templated dashboard chart whose user pinned bold data labels (e.g. for emphasis on a \"primary\" KPI tile) now round-trips that flag through the parse / clone / write loop.

Mirrors the chart-title `titleBold` (#252), the axis-title `axisTitleBold` (#256), the axis tick-label `labelBold` (#264), and the legend `legendBold` (#270) exactly: `true` emits `b=\"1\"`; `false` emits the OOXML default `b=\"0\"` explicitly so a clone target can override an upstream `b=\"1\"` from a templated chart; absence collapses to omitting the attribute entirely. The reader collapses the OOXML default `0` (and absence) to `undefined` so absence and `b=\"0\"` round-trip identically — only an explicit `b=\"1\"` surfaces `true`. Non-boolean inputs (typed escapes from an untyped caller — strings, `null`, numbers) collapse to `undefined` so the writer never emits a malformed `b` attribute.

The `<c:txPr>` block lands between `<c:numFmt>` and `<c:dLblPos>` inside `<c:dLbls>`, matching the CT_DLbls schema sequence (ECMA-376 Part 1, §21.2.2.50). The block is skipped entirely when no font knob is pinned so a fresh chart matches Excel's reference shape byte-for-byte (Excel itself omits `<c:txPr>` from `<c:dLbls>` when the labels render at the theme-default style). The clone path threads the value through the existing `resolveChartDataLabels` resolver (which spreads the read-side `ChartDataLabelsInfo` into the write-side `ChartDataLabels` shape — they share field semantics). Future typography pins (size, color, italic, underline, strikethrough) will land on the same `<a:defRPr>` slot.

Refs #136

## Test plan

- [x] `pnpm test` passes (5385 tests across 89 files, lint + format + typecheck + vitest)
- [x] `pnpm build` succeeds
- [x] New describe blocks: `parseChart — data labels bold` (9 cases), `writeChart — dataLabels.bold` (15 cases), `cloneChart — dataLabels.bold` (11 cases) covering parse, write, end-to-end `writeXlsx`, the OOXML default `0` collapse, malformed input drops, OOXML element ordering (`<c:numFmt>` -> `<c:txPr>` -> `<c:dLblPos>`), multi-family coverage (bar / column / line / pie / doughnut / area), composition with the sibling dLbls knobs (`position` / `numberFormat` / `separator` / `show*` toggles), and the standard `null`-drop / explicit-replace clone grammar

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -723,6 +723,36 @@ export interface ChartDataLabels {
    * thread a single hex string through every typography-pinning slot.
    */
   fontColor?: string;
+  /**
+   * Data-label bold flag. Maps to `<c:dLbls><c:txPr><a:p><a:pPr>
+   * <a:defRPr b=".."/></a:pPr></a:p></c:txPr></c:dLbls>` — Excel's
+   * "Format Data Labels -> Font -> Bold" toggle. The OOXML attribute
+   * is the `xsd:boolean` bold flag on `CT_TextCharacterProperties`
+   * (ECMA-376 Part 1, §21.1.2.3.7); the writer lands `b="1"` (bold)
+   * or `b="0"` (the OOXML default — non-bold) on the
+   * default-paragraph `<a:defRPr>` slot inside the data-label's
+   * `<c:txPr>` block so a re-parse picks the flag up off the
+   * canonical slot the OOXML schema exposes.
+   *
+   * Default: omitted — the data labels render non-bold (no `b`
+   * attribute, matching Excel's reference serialization for fresh
+   * data labels whose typography has not been customized; the OOXML
+   * default `0` collapses to absence). Set `true` to emit `b="1"` so
+   * the labels render bold; set `false` explicitly to pin `b="0"`
+   * (functionally identical to omission, but useful when overriding
+   * a templated chart that had bold pinned upstream).
+   *
+   * The `<c:txPr>` block lands between `<c:numFmt>` and `<c:dLblPos>`
+   * (CT_DLbls schema, ECMA-376 Part 1, §21.2.2.50). Composes
+   * independently with the other dLbls knobs — {@link position} /
+   * {@link separator} / {@link numberFormat} / {@link showLeaderLines}
+   * / the `show*` toggles. Mirrors the chart-title `titleBold` /
+   * axis-title `axisTitleBold` / axis tick-label `labelBold` / legend
+   * `legendBold` knobs — same boolean shape, same OOXML
+   * `<a:defRPr b=".."/>` mapping — so a caller can thread a single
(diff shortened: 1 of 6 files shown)


productdevbook (commented):
Reviewed the diff. Mirrors the chart-title `titleBold` (#252), axis-title `axisTitleBold` (#256), tick-label `labelBold` (#264), and legend `legendBold` (#270) patterns:

- Reader `parseDataLabelsBold` walks `<c:dLbls><c:txPr><a:p><a:pPr><a:defRPr b=...>` and surfaces `true` only on `b="1"` / `b="true"`. The OOXML default `b="0"` collapses to `undefined`.
- Writer `resolveDataLabelsBold` passes literal booleans through; non-boolean tokens collapse to `undefined`. The flag emits `b="1"` / `b="0"` whenever set so a clone target can override an upstream pin from a templated chart.
- `buildDataLabelsTxPr` was widened to accept the `bold` parameter alongside `fontSize` and `rgbHex` (the existing parameters from #276 / #277). All three pin into the same `<a:defRPr>` slot and the block is skipped entirely when none are set.

Rebased onto current main to resolve the conflicts with #276 (data labels font size) and #277 (data labels font color) — the empty-record short-circuit in `parseDataLabels` was widened to include all three new fields, the `buildDataLabelsTxPr` signature was widened from `(fontSize, rgbHex)` to `(fontSize, rgbHex, bold)`, and the resolver / parser pairs stack cleanly.

Verification on the rebased branch:
- `pnpm lint` clean (45 pre-existing warnings, 0 errors)
- `pnpm typecheck` clean
- `pnpm vitest run --dir=test` green — 5460 tests across 89 files
- `pnpm build` succeeds

LGTM after CI confirms.

PR #279, opened 2026-05-06
feat(xlsx): chart data labels italic flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:dLbls><c:txPr><a:p><a:pPr><a:defRPr i=".."/></a:pPr></a:p></c:txPr></c:dLbls>` at the read, write, and clone layers so a template's data-labels italic flag survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `italic?: boolean` on `ChartDataLabels` (writer side) and the matching `italic?: boolean` on `ChartDataLabelsInfo` (reader side). Sits on the same `<c:dLbls>` block as `position` / `numberFormat` / `separator` / `showLeaderLines` / the `show*` toggles / `bold`; the writer threads the value through the shared `buildDataLabelsBody` helper so chart-level and per-series data labels both pick up the typography pin. Bridges another typography-customization gap for the dashboard composition flow tracked in #136 — a templated dashboard chart whose user pinned italic data labels (e.g. for emphasis on a quote-style KPI tile) now round-trips that flag through the parse / clone / write loop.

Mirrors the chart-title `titleItalic` (#253), the axis-title `axisTitleItalic` (#257), the axis tick-label `labelItalic` (#265), the legend `legendItalic` (#271), and the data-labels `bold` (#278) exactly: `true` emits `i="1"`; `false` emits the OOXML default `i="0"` explicitly so a clone target can override an upstream `i="1"` from a templated chart; absence collapses to omitting the attribute entirely. The reader collapses the OOXML default `0` (and absence) to `undefined` so absence and `i="0"` round-trip identically — only an explicit `i="1"` surfaces `true`. Non-boolean inputs (typed escapes from an untyped caller — strings, `null`, numbers) collapse to `undefined` so the writer never emits a malformed `i` attribute.

Refs #136

## Test plan

- [x] `pnpm vitest run --dir=test` — 5495 passing across 89 files
- [x] `pnpm build` — clean
- [x] New parser tests cover: `i="1"` / `i="true"` surface `true`; `i="0"` / `i="false"` / absence collapse to `undefined`; missing `<c:txPr>` returns `undefined`; garbage tokens collapse; co-surfaces with `bold` / `position` / `numberFormat`
- [x] New writer tests cover: unset omits `<c:txPr>`; `true` emits `i="1"`; `false` emits `i="0"`; CT_DLbls element ordering after `<c:numFmt>` and before `<c:dLblPos>`; only one `<c:txPr>`; threads through bar/column/line/pie/doughnut/area; composes with `bold` on the same `<a:defRPr>`; non-boolean inputs (`null` / string / number) collapse to omission; emits the standard txPr stub shape; round-trips `true` and collapses `false`/unset back to `undefined`; `writeXlsx` package round-trip
- [x] New clone tests cover: inherits source value; explicit override replaces; `null` drops the inherited block; absence on both sides collapses to `undefined`; composes with other dLbls fields; `writeXlsx` round-trips through `parseChart`; flatten preserves italic through `bar -> line` and `bar -> doughnut`

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -753,6 +753,36 @@ export interface ChartDataLabels {
    * bold value through every typography-pinning slot.
    */
   bold?: boolean;
+  /**
+   * Data-label italic flag. Maps to `<c:dLbls><c:txPr><a:p><a:pPr>
+   * <a:defRPr i=".."/></a:pPr></a:p></c:txPr></c:dLbls>` — Excel's
+   * "Format Data Labels -> Font -> Italic" toggle. The OOXML attribute
+   * is the `xsd:boolean` italic flag on `CT_TextCharacterProperties`
+   * (ECMA-376 Part 1, §21.1.2.3.7); the writer lands `i="1"` (italic)
+   * or `i="0"` (the OOXML default — upright) on the default-paragraph
+   * `<a:defRPr>` slot inside the data-label's `<c:txPr>` block so a
+   * re-parse picks the flag up off the canonical slot the OOXML
+   * schema exposes.
+   *
+   * Default: omitted — the data labels render upright (no `i`
+   * attribute, matching Excel's reference serialization for fresh
+   * data labels whose typography has not been customized; the OOXML
+   * default `0` collapses to absence). Set `true` to emit `i="1"` so
+   * the labels render italic; set `false` explicitly to pin `i="0"`
+   * (functionally identical to omission, but useful when overriding
+   * a templated chart that had italic pinned upstream).
+   *
+   * The `<c:txPr>` block lands between `<c:numFmt>` and `<c:dLblPos>`
+   * (CT_DLbls schema, ECMA-376 Part 1, §21.2.2.50). Composes
+   * independently with the other dLbls knobs — {@link position} /
+   * {@link separator} / {@link numberFormat} / {@link showLeaderLines}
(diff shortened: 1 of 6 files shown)


productdevbook (commented):
Reviewed end-to-end — the implementation mirrors the established `<a:defRPr i=".."/>` pattern that landed for chart-title (#253), axis-title (#257), axis tick-label (#265), legend (#271), and the data-labels `bold` (#278) knobs. The reader (`parseDataLabelsItalic`) correctly collapses `i="0"` / `"false"` / absence to `undefined` and only surfaces `true` for `i="1"` / `"true"`, so absence and the OOXML default round-trip identically. The writer threads the resolved boolean into the shared `buildDataLabelsTxPr` helper alongside `bold`; both attributes land on the same `<a:defRPr>` element, and the `<c:txPr>` block is gated on at least one font knob being pinned so a fresh chart still matches Excel's reference shape byte-for-byte. CT_DLbls element ordering (numFmt -> txPr -> dLblPos) is preserved and explicitly tested. Test coverage is thorough across parser (`i="1"`/`"true"`/`"0"`/`"false"`/garbage), writer (true/false/non-boolean escapes/composition with bold/standard txPr stub shape), clone (inherit/override/null/flatten through bar->line/bar->doughnut), and `writeXlsx` end-to-end round-trips. CI green, no concerns — merging.

PR #280, opened 2026-05-06
feat(xlsx): chart data labels underline flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:dLbls><c:txPr><a:p><a:pPr><a:defRPr u=".."/></a:pPr></a:p></c:txPr></c:dLbls>` at the read, write, and clone layers so a template's data-labels underline flag survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `underline?: boolean` on `ChartDataLabels` (writer side) and the matching `underline?: boolean` on `ChartDataLabelsInfo` (reader side). Sits on the same `<c:dLbls>` block as `position` / `numberFormat` / `separator` / `showLeaderLines` / the `show*` toggles / `bold`; the writer threads the value through the shared `buildDataLabelsBody` helper so chart-level and per-series data labels both pick up the typography pin. Bridges another typography-customization gap for the dashboard composition flow tracked in #136 — useful for emphasising key metric data labels on dashboard tiles where the underline visually separates the value from the surrounding context.

Modeled as a boolean for symmetry with the data-labels `bold` (#278) and the chart-title `titleUnderline` (#260), axis-title `axisTitleUnderline` (#262), axis tick-label `labelUnderline` (#267), and legend `legendUnderline` (#272) knobs: `true` emits `u="sng"` (Excel's UI variant — single line), `false` emits `u="none"` explicitly so a clone target can override an upstream `u="sng"` from a templated chart, absence collapses to omitting the attribute entirely. The non-UI variant `"dbl"` (double line) and the sixteen exotic `ST_TextUnderlineType` tokens (`"words"`, `"heavy"`, `"dotted"`, `"dotDash"`, `"wavy"`, etc.) are read-only — the writer emits only `"sng"` / `"none"` to keep the surfaced shape consistent with what Excel's reference UI authors, and the reader collapses every non-`"sng"` token to `undefined` so absence and `u="none"` round-trip identically. Non-boolean inputs (typed escapes from an untyped caller — strings, `null`, numbers) collapse to `undefined` so the writer never emits a malformed `u` attribute.

Refs #136

## Test plan

- [x] `pnpm vitest run --dir=test` — full suite green (89 files)
- [x] `pnpm build` — clean
- [x] New parser tests (9): `u="sng"` surfaces `true`; `u="none"` / `u="dbl"` / every non-sng `ST_TextUnderlineType` variant collapse to `undefined`; missing `<c:txPr>` returns `undefined`; garbage tokens collapse; co-surfaces with `bold` / `position` / `numberFormat`
- [x] New writer tests (15): unset omits `<c:txPr>`; `true` emits `u="sng"`; `false` emits `u="none"`; CT_DLbls element ordering after `<c:numFmt>` and before `<c:dLblPos>`; only one `<c:txPr>`; threads through bar/column/line/pie/doughnut/area; composes with `bold` on the same `<a:defRPr>`; non-boolean inputs (`null` / string / number) collapse to omission; emits the standard txPr stub shape; round-trips `true` and collapses `false`/unset back to `undefined`; `writeXlsx` package round-trip
- [x] New clone tests (11): inherits source value; explicit override replaces; `null` drops the inherited block; absence on both sides collapses to `undefined`; composes with other dLbls fields; `writeXlsx` round-trips through `parseChart`; flatten preserves underline through `bar -> line` and `bar -> doughnut`

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -783,6 +783,41 @@ export interface ChartDataLabels {
    * a single italic value through every typography-pinning slot.
    */
   italic?: boolean;
+  /**
+   * Data-label underline flag. Maps to `<c:dLbls><c:txPr><a:p><a:pPr>
+   * <a:defRPr u=".."/></a:pPr></a:p></c:txPr></c:dLbls>` — Excel's
+   * "Format Data Labels -> Font -> Underline" toggle. The OOXML
+   * attribute is the `ST_TextUnderlineType` enumeration on
+   * `CT_TextCharacterProperties` (ECMA-376 Part 1, §21.1.2.3.7); the
+   * writer lands `u="sng"` (single underline — Excel's UI checkbox)
+   * or `u="none"` (the OOXML default — no underline) on the
+   * default-paragraph `<a:defRPr>` slot inside the data-label's
+   * `<c:txPr>` block so a re-parse picks the flag up off the
+   * canonical slot the OOXML schema exposes.
+   *
+   * Modeled as a boolean for symmetry with {@link bold} / {@link italic}
+   * / the other `*Underline` knobs across the chart-title, axis-title,
+   * axis tick-label, and legend slots: `true` emits `u="sng"`; `false`
+   * emits `u="none"` explicitly so a clone target can override an
+   * upstream `u="sng"` from a templated chart; absence collapses to
+   * omitting the attribute entirely. The non-UI variant `"dbl"`
+   * (double line) and the sixteen exotic `ST_TextUnderlineType`
+   * tokens (`"words"`, `"heavy"`, `"dotted"`, etc.) are read-only —
+   * the writer emits only `"sng"` / `"none"` to keep the surfaced
+   * shape consistent with what Excel's reference UI authors.
+   *
+   * The `<c:txPr>` block lands between `<c:numFmt>` and `<c:dLblPos>`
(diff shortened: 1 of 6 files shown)


productdevbook (commented):
Reviewed end-to-end — implementation mirrors the established `<a:defRPr u=".."/>` pattern that landed for chart-title (#260), axis-title (#262), axis tick-label (#267), legend (#272), and composes cleanly on top of the data-labels `bold` (#278) and italic (#279) knobs. The reader correctly accepts only `u="sng"` as `true` and collapses every other `ST_TextUnderlineType` variant (`none`/`dbl`/`heavy`/`dotted`/`dotDash`/`wavy`/etc.) to `undefined` — the rationale (don't silently downgrade an exotic upstream variant to single-line on round-trip) is right and matches the existing readers. The writer emits `u="sng"` for `true` and `u="none"` for `false` so a clone target can override an upstream `u="sng"`, and gates the entire `<c:txPr>` block on at least one font knob being pinned. CT_DLbls ordering preserved. Test coverage hits the parser variant matrix, the writer (true/false/non-boolean inputs/standard txPr stub shape/composition with bold), clone inherit/override/null/flatten paths, and a packaged `writeXlsx` round-trip. CI green — merging.

PR #273, opened 2026-05-06
feat(xlsx): chart legend strikethrough flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:legend><c:txPr><a:p><a:pPr><a:defRPr strike=\"..\"/></a:pPr></a:p></c:txPr></c:legend>` at the read, write, and clone layers so a template's legend strikethrough pin survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `legendStrikethrough?: boolean` on `SheetChart` (writer side) and the matching `legendStrikethrough?: boolean` on `Chart` (reader side). Sits on the same `<c:legend>` element as `legend` / `legendOverlay` / `legendEntries` / `legendFontSize` / `legendBold` / `legendItalic` / `legendUnderline`; the writer emits a single `<c:txPr>` block on the legend whenever any typography knob is pinned, and the strikethrough value lands on the same `<a:defRPr>` slot as the size / bold / italic / underline. Bridges another typography-customization gap for the dashboard composition flow tracked in #136 — a templated dashboard chart whose user pinned a custom legend strikethrough (e.g. striking through legend swatch labels for "deprecated" series in a sustainability before/after dashboard) now round-trips that pin through the parse / clone / write loop.

Modeled as a boolean for symmetry with the chart-title `titleStrikethrough` (#259), the axis-title `axisTitleStrike` (#261), and the axis tick-label `labelStrikethrough` (#268): `true` emits `strike=\"sngStrike\"` (Excel's UI checkbox — single line); absence and explicit `false` both collapse to omitting the attribute (the OOXML default `\"noStrike\"` collapses to absence — the writer keeps the surfaced shape consistent with what Excel's UI authors). The non-UI variant `\"dblStrike\"` (double line) is read-only — the writer emits only `\"sngStrike\"`, and the reader collapses `\"dblStrike\"` to `undefined` so a templated legend that pinned the double-line variant in raw OOXML round-trips lossless rather than silently downgrading on re-emit.

[...]

PR #281, opened 2026-05-06
feat(xlsx): chart data labels strikethrough flag — read, write, clone-through

productdevbook:
## Summary

Surfaces `<c:dLbls><c:txPr><a:p><a:pPr><a:defRPr strike=".."/></a:pPr></a:p></c:txPr></c:dLbls>` at the read, write, and clone layers so a template's data-labels strikethrough flag survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `strikethrough?: boolean` on `ChartDataLabels` (writer side) and the matching `strikethrough?: boolean` on `ChartDataLabelsInfo` (reader side). Sits on the same `<c:dLbls>` block as `position` / `numberFormat` / `separator` / `showLeaderLines` / the `show*` toggles / `bold`; the writer threads the value through the shared `buildDataLabelsBody` helper so chart-level and per-series data labels both pick up the typography pin. Bridges another typography-customization gap for the dashboard composition flow tracked in #136.

Modeled as a boolean for symmetry with the data-labels `bold` (#278) and the chart-title `titleStrikethrough` (#259), axis-title `axisTitleStrike` (#261), axis tick-label `labelStrikethrough` (#268), and legend `legendStrikethrough` (#273) knobs: `true` emits `strike="sngStrike"` (Excel's UI variant — single line); absence and explicit `false` both collapse to omitting the attribute (the OOXML default `"noStrike"` is functionally identical to absence — the writer never emits `"noStrike"` or `"dblStrike"` to keep the surfaced shape consistent with what Excel's UI authors). The non-UI variant `"dblStrike"` (double line) is read-only — the reader collapses every non-`"sngStrike"` token to `undefined` so a templated chart that pinned the double-line variant in raw OOXML round-trips lossless rather than silently downgrading on re-emit. Non-boolean inputs (typed escapes from an untyped caller — strings, `null`, numbers) collapse to `undefined` so the writer never emits a malformed `strike` attribute.

Refs #136

## Test plan

- [x] `pnpm vitest run --dir=test` — full suite green (89 files)
- [x] `pnpm build` — clean
- [x] New parser tests (8): `strike="sngStrike"` surfaces `true`; `strike="noStrike"` / `strike="dblStrike"` / absence collapse to `undefined`; missing `<c:txPr>` returns `undefined`; garbage tokens collapse; co-surfaces with `bold` / `position` / `numberFormat`
- [x] New writer tests (15): unset omits `<c:txPr>`; `true` emits `strike="sngStrike"`; `false` collapses to absence (no `<c:txPr>` emitted); CT_DLbls element ordering after `<c:numFmt>` and before `<c:dLblPos>`; only one `<c:txPr>`; threads through bar/column/line/pie/doughnut/area; composes with `bold` on the same `<a:defRPr>`; non-boolean inputs (`null` / string / number) collapse to omission; emits the standard txPr stub shape; round-trips `true` and collapses `false`/unset back to `undefined`; `writeXlsx` package round-trip
- [x] New clone tests (11): inherits source value; explicit override replaces; `null` drops the inherited block; absence on both sides collapses to `undefined`; composes with other dLbls fields; `writeXlsx` round-trips through `parseChart`; flatten preserves strikethrough through `bar -> line` and `bar -> doughnut`

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -818,6 +818,39 @@ export interface ChartDataLabels {
    * typography-pinning slot.
    */
   underline?: boolean;
+  /**
+   * Data-label strikethrough flag. Maps to `<c:dLbls><c:txPr><a:p>
+   * <a:pPr><a:defRPr strike=".."/></a:pPr></a:p></c:txPr></c:dLbls>`
+   * — Excel's "Format Data Labels -> Font -> Strikethrough" toggle.
+   * The OOXML attribute is the `ST_TextStrikeType` enumeration on
+   * `CT_TextCharacterProperties` (ECMA-376 Part 1, §21.1.2.3.7); the
+   * writer lands `strike="sngStrike"` (single line — Excel's UI
+   * checkbox) on the default-paragraph `<a:defRPr>` slot inside the
+   * data-label's `<c:txPr>` block so a re-parse picks the flag up
+   * off the canonical slot the OOXML schema exposes.
+   *
+   * Modeled as a boolean for symmetry with the data-labels {@link bold}
+   * / {@link italic} / {@link underline} and the chart-title
+   * `titleStrikethrough` / axis-title `axisTitleStrike` / axis
+   * tick-label `labelStrikethrough` / legend `legendStrikethrough`
+   * knobs: `true` emits `strike="sngStrike"`; absence and explicit
+   * `false` both collapse to omitting the attribute (the OOXML default
+   * `"noStrike"` is functionally identical to absence — the writer
+   * never emits `"noStrike"` or `"dblStrike"` to keep the surfaced
+   * shape consistent with what Excel's UI authors). The non-UI variant
+   * `"dblStrike"` (double line) is read-only — the reader collapses
+   * every non-`"sngStrike"` token to `undefined` so a templated chart
+   * that pinned the double-line variant in raw OOXML round-trips
+   * lossless rather than silently downgrading on re-emit.
+   *
+   * The `<c:txPr>` block lands between `<c:numFmt>` and `<c:dLblPos>`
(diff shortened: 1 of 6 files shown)


productdevbook (commented):
Reviewed end-to-end — implementation mirrors the established `<a:defRPr strike=".."/>` pattern that landed for chart-title (#259), axis-title (#261), axis tick-label (#268), legend (#273), and composes cleanly on top of the data-labels `bold` (#278), italic (#279), and underline (#280) knobs. The reader correctly accepts only `strike="sngStrike"` as `true` and collapses every other `ST_TextStrikeType` token (`noStrike`/`dblStrike`/garbage) to `undefined` so absence and the OOXML default round-trip identically. The writer's resolver collapses both explicit `false` and absence to skipping the attribute (only `"sngStrike"` ever lands) — this matches the existing `resolveTitleStrike` / `resolveLegendStrikethrough` precedent (the project's strikethrough writers never emit `"noStrike"` or `"dblStrike"`, keeping the surfaced shape consistent with Excel's UI). The shared `buildDataLabelsTxPr` signature now accumulates all four typography flags (bold/italic/underline/strikethrough) on the same `<a:defRPr>` element, and the entire `<c:txPr>` block is gated on at least one font knob being pinned. CT_DLbls element ordering preserved. Test coverage hits the parser variant matrix (sngStrike/noStrike/dblStrike/garbage), writer (true/false/non-boolean inputs/standard txPr stub shape/composition with bold), clone inherit/override/null/flatten paths, and `writeXlsx` package round-trips. Resolved the rebase conflict against #279/#280 by extending the shared `buildDataLabelsTxPr` signature to thread italic, underline, and strikethrough alongside bold; full local test suite (5564 tests) green. CI re-running on the rebased commit — merging once green.

PR #283, opened 2026-05-06
feat(xlsx): chart axis title font family — read, write, clone-through

productdevbook:
Surfaces `<c:catAx>` / `<c:valAx>` / `<c:dateAx>` / `<c:serAx>` -> `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr><a:latin typeface=".."/></a:defRPr></a:pPr><a:r><a:rPr><a:latin typeface=".."/></a:rPr></a:r></a:p></c:rich></c:tx></c:title>` (CT_TextFont on CT_TextCharacterProperties' latin slot — ECMA-376 Part 1, §21.1.2.3.7) at the read, write, and clone layers so a template's axis-title typeface survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `axisTitleFontFamily?: string` on `SheetChart.axes.x|y` (writer side) and on `ChartAxisInfo` (reader side). Sits on the same `<c:title>` body as `axisTitleColor` / `axisTitleBold` / `axisTitleItalic` / `axisTitleStrike` / `axisTitleUnderline` / `axisTitleFontSize` / `axisTitleRotation`; the writer threads the value through the existing `buildAxisTitle` helper so the run-level `<a:rPr>` and the default-paragraph `<a:defRPr>` both pick up the typeface.

Mirrors the chart-level `titleFontFamily` field (#282) — same canonical-slot pair, same accept-and-trim grammar, same `string | null` clone grammar so a single configuration call threads cleanly through both the chart title and either axis title without bookkeeping the canonical OOXML slots.

The `<a:latin>` element follows `<a:solidFill>` per the CT_TextCharacterProperties child sequence so an axis title with both `axisTitleColor` and `axisTitleFontFamily` set lands the children in canonical schema order — a fresh chart matches Excel's reference serialization byte-for-byte. Sits on every axis flavour (`<c:catAx>` / `<c:valAx>` / `<c:dateAx>` / `<c:serAx>`) so the field round-trips on every chart family that has axes (bar / column / line / area / scatter); silently dropped on pie / doughnut.

Refs #136

## Test plan

- [x] `pnpm vitest run --dir=test` — 5602 tests pass
- [x] `pnpm build` — succeeds
- [x] reader: surfaces typeface on catAx and valAx; trims whitespace; collapses empty / whitespace-only / absent / non-string to undefined; scoped so tick-label `<c:txPr><a:latin>` cannot leak in
[...]

PR #282, opened 2026-05-06
feat(xlsx): chart title font family — read, write, clone-through

productdevbook:
Surfaces `<c:title><c:tx><c:rich><a:p><a:pPr><a:defRPr><a:latin typeface=".."/></a:defRPr></a:pPr><a:r><a:rPr><a:latin typeface=".."/></a:rPr></a:r></a:p></c:rich></c:tx></c:title>` (CT_TextFont on CT_TextCharacterProperties' latin slot — ECMA-376 Part 1, §21.1.2.3.7) at the read, write, and clone layers so a template's chart-title typeface survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `titleFontFamily?: string` on `SheetChart` (writer side) and `titleFontFamily?: string` on `Chart` (reader side). Sits on the same `<c:title>` block as `titleColor` / `titleBold` / `titleItalic` / `titleStrike` / `titleUnderline` / `titleFontSize` / `titleRotation` / `titleOverlay`; the writer threads the value through the existing `buildTitle` helper so the run-level `<a:rPr>` and the default-paragraph `<a:defRPr>` both pick up the typeface.

Until now hucre's writer omitted `<a:latin>` entirely (the title inherited the theme typeface) and the reader ignored the element, so a templated dashboard chart whose user picked a custom title font silently lost the choice on every parse -> clone -> write loop. Bridges another title-customization gap for the dashboard composition flow tracked in #136.

Mirrors the chart-level `titleColor` field added in #254 — same canonical-slot pair (`<a:defRPr>` + `<a:rPr>`), same string-typed clone grammar (`undefined` inherits, `null` drops, a non-empty string replaces, empty / whitespace-only strings collapse to a drop). The `<a:latin>` element follows `<a:solidFill>` per the CT_TextCharacterProperties child sequence so a chart with both `titleColor` and `titleFontFamily` set lands the children in canonical schema order — a fresh chart matches Excel's reference serialization byte-for-byte. The writer trims surrounding whitespace and emits the trimmed value verbatim (XML-escaped) so attribute-injection escapes never produce a malformed `typeface` attribute.

Refs #136

## Test plan

- [x] `pnpm vitest run --dir=test` — 5601 tests pass
- [x] `pnpm build` — succeeds
[...]

PR #284, opened 2026-05-06
feat(xlsx): chart axis tick-label font family — read, write, clone-through

productdevbook:
Surfaces `<c:catAx>` / `<c:valAx>` / `<c:dateAx>` / `<c:serAx>` -> `<c:txPr><a:p><a:pPr><a:defRPr><a:latin typeface=".."/></a:defRPr></a:pPr></a:p></c:txPr>` (CT_TextFont on CT_TextCharacterProperties' latin slot — ECMA-376 Part 1, §21.1.2.3.7) at the read, write, and clone layers so a template's axis tick-label typeface survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `labelFontFamily?: string` on `SheetChart.axes.x|y` (writer side) and on `ChartAxisInfo` (reader side). Sits on the same `<c:txPr>` body as `labelColor` / `labelBold` / `labelItalic` / `labelStrike` / `labelUnderline` / `labelFontSize` / `labelRotation`; the writer threads the value through the existing `buildAxisTxPr` helper.

Mirrors the chart-title `titleFontFamily` (#282) and axis-title `axisTitleFontFamily` (#283) — same canonical-slot grammar, same accept-and-trim shape, same `string | null` clone grammar so a single configuration call threads cleanly through every typeface knob the writer exposes.

The `<a:latin>` element follows `<a:solidFill>` per the CT_TextCharacterProperties child sequence so an axis with both `labelColor` and `labelFontFamily` set lands the children in canonical schema order — a fresh chart matches Excel's reference serialization byte-for-byte. Sits on every axis flavour (`<c:catAx>` / `<c:valAx>` / `<c:dateAx>` / `<c:serAx>`) so the field round-trips on every chart family that has axes (bar / column / line / area / scatter); silently dropped on pie / doughnut.

Refs #136

## Test plan

- [x] `pnpm vitest run --dir=test` — 5599 tests pass
- [x] `pnpm build` — succeeds
- [x] reader: surfaces typeface on catAx and valAx; trims whitespace; collapses empty / whitespace-only / absent / non-string to undefined; scoped so axis-title `<c:title><a:latin>` cannot leak in
- [x] writer: omits `<c:txPr>` by default; emits `<a:latin>` inside the `<c:txPr><a:defRPr>`; XML-escapes; composes with color in canonical schema order; composes with bold / italic / strike / underline / size; threads through every chart family
- [x] clone: inherits, overrides, drops on `null`, drops on pie/doughnut; trims whitespace; drops malformed overrides; preserves Y-axis when only X is overridden
- [x] end-to-end: `writeXlsx` -> `extractXml` -> `parseChart` retains the typeface on both axes

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -3043,6 +3043,48 @@ export interface SheetChart {
        */
       labelStrike?: boolean;
       /**
+       * Axis tick-label font family / typeface. Maps to
+       * `<c:catAx>` / `<c:valAx>` / `<c:dateAx>` / `<c:serAx>` ->
+       * `<c:txPr><a:p><a:pPr><a:defRPr><a:latin typeface=".."/>
+       * </a:defRPr></a:pPr></a:p></c:txPr>` — Excel's "Format Axis ->
+       * Number / Font -> Font" picker scoped to the axis's tick
+       * labels. The OOXML `<a:latin typeface=".."/>` element carries
+       * the typeface name (`CT_TextFont`, ECMA-376 Part 1,
+       * §21.1.2.3.7); the writer lands the element on the default-
+       * paragraph `<a:defRPr>` of the axis-level `<c:txPr>` body.
+       *
+       * Accepts any non-empty string typeface name (e.g. `"Calibri"`,
+       * `"Arial"`, `"Times New Roman"`); the writer trims surrounding
+       * whitespace and emits the trimmed value verbatim (XML-escaped)
+       * so Excel can resolve the named font from the workbook's font
+       * scheme or the host system's installed fonts. Empty /
+       * whitespace-only strings and non-string tokens collapse to
+       * `undefined` so the writer skips the entire `<a:latin>`
+       * element and the tick labels inherit Excel's reference theme
+       * typeface.
+       *
+       * Default: omitted — the tick labels render in Excel's
+       * reference theme typeface (no `<a:latin>` element, the writer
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed end-to-end — looks ready to land.

- Reader (`parseAxisLabelFontFamily`): walks the canonical `<c:txPr> -> <a:p> -> <a:pPr> -> <a:defRPr> -> <a:latin>` chain, bails on missing link, trims, drops empty/whitespace/non-string. Scoped to axis-level `<c:txPr>` so the title's `<a:latin>` cannot leak in (this is the symmetric guard to the title-side scope check in #283).
- Writer (`buildAxisTxPr`): extended signature to accept `fontFamily`, the `undefined`-guard now includes it so a fresh axis with no typography knobs still collapses to `undefined` (no `<c:txPr>` emitted). Children built in canonical schema order — solidFill before latin. `normalizeAxisLabelFontFamily` mirrors the title normalize helper.
- Clone: `applyLabelFontFamilyOverride` follows the same `undefined`/`null`/value grammar as siblings; pie/doughnut short-circuit drops the field upstream (verified by the pie test).
- Tests: full parse/write/clone/round-trip coverage; the pie/doughnut drop test, Y-axis preservation, and compose-with-all-knobs test are all here.

CI green (build SUCCESS). Merging.

PR #287, opened 2026-05-06
feat(xlsx): chart axis title overlay flag — read, write, clone-through

productdevbook:
Surfaces `<c:catAx><c:title><c:overlay val=\"..\"/></c:title></c:catAx>` (or `<c:valAx>` / `<c:dateAx>` / `<c:serAx>`) — the OOXML `<c:overlay>` child of `CT_Title` — at the read, write, and clone layers so a template's axis-title overlay flag survives `parseChart` -> `cloneChart` -> `writeXlsx` and can be authored from scratch on a fresh chart.

The model is `axisTitleOverlay?: boolean` on `SheetChart.axes.x|y` (writer side) and on `ChartAxisInfo` (reader side). Sits as a direct child of `<c:title>` per the CT_Title schema (the same OOXML body shape the chart-level title carries). The writer always emits `<c:overlay>` inside every visible axis title (matching Excel's reference serialization); only the `val` attribute flips when the caller pins `axisTitleOverlay: true`.

Mirrors the chart-level `titleOverlay` (#224) — same `boolean | null` clone grammar, same canonical-slot mapping — so a caller can thread a single overlay toggle through both the chart and axis titles. Excel's UI does not surface this toggle for axis titles directly, but the OOXML schema (CT_Title is shared with the chart-level title) carries the element on every emitted axis-title block, so a hand-edited template that pinned `val=\"1\"` round-trips cleanly.

Surfaced on every chart family that has axes (bar / column / line / area / scatter); silently dropped on pie / doughnut. Silently dropped when the matching axis renders no title (no `<c:title>` block to host the overlay flag).

Refs #136

## Test plan

- [x] `pnpm exec vitest run` — all tests pass (35 new)
- [x] `pnpm test` — lint + typecheck + vitest all pass
- [x] `pnpm build` — succeeds
- [x] reader: surfaces `axisTitleOverlay` from `<c:title><c:overlay val=\"1\"/>`; collapses OOXML default `val=\"0\"` to undefined; scoped to `<c:title>` (chart-level overlay does not leak in); returns undefined when `<c:title>` is absent
- [x] writer: emits `<c:overlay val=\"0\"/>` by default; emits `<c:overlay val=\"1\"/>` on `axisTitleOverlay: true`; non-boolean inputs collapse to default; threads independently across X / Y axes; per CT_Title schema order (after `<c:tx>`)
- [x] cloneChart: inherits via `applyAxisTitleOverlayOverride`; `null` drops; `boolean` replaces; survives writeXlsx round-trip; dropped when the resolved axis has no title; dropped on pie / doughnut

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -2797,6 +2797,45 @@ export interface SheetChart {
        * area / scatter).
        */
       axisTitleFontFamily?: string;
+      /**
+       * Axis-title overlay flag. Maps to
+       * `<c:catAx><c:title><c:overlay val=".."/></c:title></c:catAx>`
+       * (or `<c:valAx>` / `<c:dateAx>` / `<c:serAx>`) — the OOXML
+       * `<c:overlay>` child of `CT_Title`. The element controls
+       * whether the axis title is drawn on top of (and may overlap)
+       * the plot area, mirroring how the chart-level
+       * {@link SheetChart.titleOverlay} controls the chart title's
+       * overlap.
+       *
+       * Default: omitted — the writer emits the OOXML default
+       * `val="0"` (the axis title reserves its own slot adjacent to
+       * the axis and the plot area shrinks to make room), matching
+       * Excel's reference serialization for a fresh axis title. Pin
+       * `axisTitleOverlay: true` to draw the axis title on top of the
+       * plot area so the chart series get the full frame
+       * (`val="1"`). Excel's UI does not surface this toggle for axis
+       * titles directly, but the OOXML schema (CT_Title is shared
+       * with the chart-level title) carries the element on every
+       * emitted axis-title block — the writer always emits
+       * `<c:overlay>` so a hand-edited template that pinned `val="1"`
+       * round-trips cleanly.
+       *
+       * Mirrors {@link SheetChart.titleOverlay} — same `boolean`
+       * shape, same OOXML `<c:overlay val=".."/>` mapping — so a
+       * caller can thread a single overlay toggle through both the
+       * chart and axis titles.
+       *
+       * Silently dropped when the axis renders no title (the
+       * `<c:title>` element is absent in either case) and on `pie` /
+       * `doughnut` charts (no axes at all).
+       *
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Review — feat(xlsx): chart axis title overlay flag

Solid implementation. Mirrors the chart-level `titleOverlay` (PR #224) faithfully, extended into the per-axis-title slot.

Reader (`parseAxisTitleOverlay`):
- Scoped to direct `<c:title>` children — verified by the dedicated test that pins both chart-level and axis-level overlays in the same fixture, confirming no cross-leak.
- Accepts the OOXML truthy / falsy spellings (`"1"` / `"true"` / `"0"` / `"false"`); unknown / malformed / missing values drop to `undefined`.
- Default `false` collapses to `undefined` so absence and `<c:overlay val="0"/>` round-trip identically.
- Properly added to the empty-axis-record collapse guard.

Writer (`buildAxisTitle` extension):
- Always emits `<c:overlay>` inside every visible axis title (matches Excel reference). Only `val` flips.
- `chart.axes?.x?.axisTitleOverlay === true` strict-coercion correctly defends against non-boolean type-guard escapes.
- `<c:overlay>` placement after `<c:tx>` per CT_Title schema — confirmed by ordering test.
- Threaded through every chart family that has axes (bar / column / line / area / scatter), short-circuited on pie / doughnut.

Clone (`applyAxisTitleOverlayOverride`):
- Standard `undefined`/`null`/`boolean` grammar.
- Title-presence gate: `xTitle === undefined ? undefined : xAxisTitleOverlay` ensures the cloned shape never carries a flag the writer would silently elide.
- Both X and Y axis output blocks have `xAxisTitleOverlayResolved` / `yAxisTitleOverlayResolved` added to the `!== undefined` emission guard.
- Non-boolean overrides collapse via the resolver.

Tests cover: "1" / "true" / "0" / "false" / unknown / missing forms, chart-level scoping isolation, X / Y independence, composition with axisTitleRotation / Bold / Color / FontSize / FontFamily, every chart family with axes, pie / doughnut short-circuit, no-title gate, full `writeXlsx` round-trip.

CI green. LGTM.

PR #298, opened 2026-05-07
feat(xlsx): chart legend manual layout — read, write, clone-through

productdevbook:
## Summary

- Adds `legendLayout?: ChartManualLayout` to `SheetChart` / `Chart`, mapped to `<c:legend><c:layout><c:manualLayout>` per `CT_Legend` / `CT_ManualLayout` (ECMA-376 Part 1, §21.2.2.114 / §21.2.2.115) — Excel's "Format Legend -> Position -> Custom" placement.
- New generic `ChartManualLayout` shape exposes optional `x` / `y` / `w` / `h` coordinates as fractions of the chart frame in the `0..1` band; each axis is independently optional so callers can pin only the position, only the size, or any combination.
- Reader / writer / clone-through all support the new knob with the established `legendOverlay` / `legendEntries` / `legendFontSize` grammar (`undefined` inherits, `null` drops, a value replaces); hidden-legend scoping drops the field silently.
- Writer emits `<c:layout>` between `<c:legendEntry>` and `<c:overlay>` per `CT_Legend`; mode children (`<c:xMode>` / `<c:yMode>` / `<c:wMode>` / `<c:hMode>` with `val="edge"`) precede value children per the schema sequence. Out-of-range / non-finite / non-numeric coordinates collapse axis-by-axis; an empty layout collapses to no `<c:layout>` block so a fresh chart matches Excel's reference serialization byte-for-byte.
- Reader admits both `xMode="edge"` (absolute) and `xMode="factor"` (delta from auto-layout) and surfaces the same shape; the writer always normalizes to `"edge"` on emit since that is what Excel itself emits when the user drags the legend to a custom position.
- `ChartManualLayout` is exported from the package root so future PRs that add manual placement on the plot area / title / data table can reuse the same generic type.

## Test plan

- [x] `pnpm typecheck` — passes.
- [x] `pnpm lint` — no new errors (existing warnings unchanged).
- [x] `pnpm vitest run --dir=test` — 6169 tests pass (54 new tests across writer / reader / clone-through, including the full `writeXlsx` round-trip).

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -1616,6 +1616,70 @@ export interface ChartLegendEntry {
 }
 
 /**
+ * Manual placement of a chart sub-element (legend / plot area / title /
+ * data table) inside the chart frame. Maps to OOXML's `<c:manualLayout>`
+ * (`CT_ManualLayout`, ECMA-376 Part 1, §21.2.2.115) — the element block
+ * that backs Excel's "Format <element> -> Position -> Custom" knob and
+ * pins where the element draws inside the chart's drawing area.
+ *
+ * All four coordinates are fractions of the chart frame in the range
+ * `0..1` — `(0, 0)` is the upper-left of the chart frame, `(1, 1)` is
+ * the lower-right; widths / heights are sized as fractions of the chart
+ * frame width / height. Each axis is independently optional so a caller
+ * can pin only the position ({@link x} / {@link y}) and let the element
+ * keep its automatic size, only the size ({@link w} / {@link h}) and
+ * let the element keep its automatic anchor, or any combination.
+ *
+ * The writer always emits the matching `<c:xMode>` / `<c:yMode>` /
+ * `<c:wMode>` / `<c:hMode>` children with `val="edge"` (Excel's "Format
+ * Legend -> Position" reference shape — the coordinates are absolute
+ * fractions of the chart frame, not deltas from the auto-layout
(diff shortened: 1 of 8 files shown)


productdevbook (commented):
LGTM. The `legendLayout` knob extends `SheetChart` / `Chart` cleanly with a generic `ChartManualLayout` shape that future PRs can reuse for plot-area / title / data-table placement.

Highlights:
- Reader / writer / clone-through grammar mirrors the existing `legendOverlay` / `legendEntries` / `legendFontSize` triad — `undefined` inherits, `null` drops, a value replaces.
- Writer emits `<c:layout>` between `<c:legendEntry>` and `<c:overlay>` per CT_Legend (ECMA-376 §21.2.2.114); mode children precede value children inside `<c:manualLayout>` per §21.2.2.115.
- Reader admits both `xMode="edge"` and `xMode="factor"`, surfacing the same shape; writer normalizes to `"edge"` on emit (Excel's reference shape for user-dragged custom positions).
- Out-of-range / non-finite / non-numeric coordinates collapse axis-by-axis — an empty layout collapses to no `<c:layout>` block so a fresh chart matches Excel byte-for-byte.
- Hidden-legend scoping drops the field silently (no slot to host the layout when `<c:legend>` itself is absent).
- 54 new tests across writer / reader / clone-through, including a full `writeXlsx` round-trip.

PR #299, opened 2026-05-07
feat(xlsx): chart plot-area manual layout — read, write, clone-through

productdevbook:
## Summary

- Adds `plotAreaLayout?: ChartManualLayout` to `SheetChart` / `Chart`, mapped to `<c:plotArea><c:layout><c:manualLayout>` per `CT_PlotArea` / `CT_ManualLayout` (ECMA-376 Part 1, §21.2.2.145 / §21.2.2.115) — Excel's "Format Plot Area -> Position -> Custom" placement.
- Reuses the existing generic `ChartManualLayout` shape introduced for `legendLayout` in #298, so callers thread the same `(x, y, w, h)` 0..1 fractions through both manual-layout slots without bookkeeping a second type.
- Writer keeps Excel's bare `<c:layout/>` placeholder when the field is unset (or every coordinate dropped on normalization) so a fresh chart matches the reference serialization byte-for-byte; pinning at least one coordinate upgrades the placeholder to `<c:layout><c:manualLayout>...</c:manualLayout></c:layout>` with `<c:xMode>` / `<c:yMode>` / `<c:wMode>` / `<c:hMode>` set to `"edge"` (Excel's reference shape on a user drag).
- Reader admits both `xMode="edge"` and `xMode="factor"` and surfaces the same shape; the bare placeholder collapses cleanly to `undefined` on parse (no `<c:manualLayout>` child).
- Clone-through follows the established `legendLayout` grammar (`undefined` inherits, `null` drops, value replaces). Unlike the legend variant, this layout is never gated on a visibility flag — every chart owns a `<c:plotArea>`, so the field survives through a `legend: false` clone unchanged.

Refs #298

## Test plan

- [x] `pnpm typecheck` — passes.
- [x] `pnpm lint` — no new errors (existing 45 warnings unchanged, formatting clean).
- [x] `pnpm vitest run --dir=test` — 6219 + 60 new tests pass across writer / reader / clone-through, including `writeXlsx` round-trip and `xMode="factor"` admission.
- [x] `pnpm build` — succeeds.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #300, opened 2026-05-07
feat(xlsx): chart title manual layout — read, write, clone-through

productdevbook:
## Summary

- Adds `titleLayout?: ChartManualLayout` to `SheetChart` / `Chart`, mapped to `<c:title><c:layout><c:manualLayout>` per `CT_Title` / `CT_ManualLayout` (ECMA-376 Part 1, §21.2.2.210 / §21.2.2.115) — Excel's "Format Chart Title -> Title Options -> Position -> Custom" placement (the same drag-handle a user sees when grabbing the title block in Excel's chart editor).
- Reuses the existing generic `ChartManualLayout` shape introduced for `legendLayout` in #298, so callers thread the same `(x, y, w, h)` 0..1 fractions through both manual-layout slots without bookkeeping a second type.
- Writer emits `<c:layout>` between `<c:tx>` and `<c:overlay>` per `CT_Title`; mode children (`<c:xMode>` / `<c:yMode>` / `<c:wMode>` / `<c:hMode>` with `val="edge"`) precede value children per the schema sequence. Out-of-range / non-finite / non-numeric coordinates collapse axis-by-axis; an empty layout collapses to no `<c:layout>` block so a fresh chart matches Excel's reference serialization byte-for-byte.
- Reader admits both `xMode="edge"` and `xMode="factor"` and surfaces the same shape; the writer always normalizes to `"edge"` on emit since that is what Excel itself emits when the user drags the title to a custom position.
- Clone-through follows the established title-knob grammar (`undefined` inherits, `null` drops, value replaces) and is gated on the resolved title visibility — `title` resolved to `undefined` or `showTitle === false` drops the field silently, since there is no `<c:title>` block to host the layout in either case.

Refs #298

## Test plan

- [x] `pnpm typecheck` — passes.
- [x] `pnpm lint` — no new errors (existing 45 warnings unchanged, formatting clean).
- [x] `pnpm vitest run --dir=test` — 6226 tests pass (57 new tests across writer / reader / clone-through, including `writeXlsx` round-trip and `xMode="factor"` admission).
- [x] `pnpm build` — succeeds.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #301, opened 2026-05-07
feat(xlsx): chart axis-title manual layout — read, write, clone-through

productdevbook:
## Summary

- Adds `axisTitleLayout?: ChartManualLayout` to `SheetChart.axes.x` / `axes.y` and the parsed `Chart.axes.x` / `axes.y` / `ChartAxisInfo`, mapped to `<c:catAx><c:title><c:layout><c:manualLayout>...</c:layout></c:title></c:catAx>` (or `<c:valAx>` / `<c:dateAx>` / `<c:serAx>`) per `CT_Title` / `CT_ManualLayout` (ECMA-376 Part 1, §21.2.2.210 / §21.2.2.115) — Excel's "Format Axis Title -> Title Options -> Position -> Custom" placement, the same drag-handle a user sees when grabbing the axis-title block in Excel's chart editor.
- Reuses the existing generic `ChartManualLayout` shape introduced for `legendLayout` in #298 and `plotAreaLayout` in #299, so callers thread the same `(x, y, w, h)` 0..1 fractions through axis titles too without bookkeeping a new type. Both X and Y axis titles share the field so a single configuration call composes across category and value axes.
- Writer emits `<c:layout>` between `<c:tx>` and `<c:overlay>` per `CT_Title`; mode children (`<c:xMode>` / `<c:yMode>` / `<c:wMode>` / `<c:hMode>` with `val="edge"`) precede value children per the schema sequence. Out-of-range / non-finite / non-numeric coordinates collapse axis-by-axis through the shared `normalizeManualLayout` helper; an empty layout collapses to no `<c:layout>` block so a fresh chart matches Excel's reference serialization byte-for-byte. The writer gates emission on the matching axis title's presence — a stray pin on an axis without a title silently drops at emit time.
- Reader admits both `xMode="edge"` and `xMode="factor"` and surfaces the same shape; the writer always normalizes to `"edge"` on emit since that is what Excel itself emits when the user drags the title to a custom position. The lookup is scoped to `<c:title>` children so a stray `<c:layout>` elsewhere on the axis (or on `<c:plotArea>` / `<c:legend>`) cannot leak in.
- Clone-through follows the established axis-title-knob grammar (`undefined` inherits, `null` drops, value replaces) and is gated on the resolved axis title's presence — the cloned `SheetChart` drops the field silently when the resolved `title` is `undefined`, symmetric with the writer's title-presence gate. Threads through every chart family that has axes (bar / column / line / area / scatter); pie / doughnut are short-circuited upstream so the field has no slot there.

Refs #299

## Test plan

- [x] `pnpm typecheck` — passes.
- [x] `pnpm lint` — no new errors (existing 45 warnings unchanged, formatting clean).
- [x] `pnpm vitest run --dir=test` — 6276 tests pass (57 new tests across writer / reader / clone-through, including `writeXlsx` round-trip and `xMode="factor"` admission).

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -3486,6 +3486,58 @@ export interface SheetChart {
        * area / scatter).
        */
       axisTitleOverlay?: boolean;
+      /**
+       * Axis-title manual placement. Maps to
+       * `<c:catAx><c:title><c:layout><c:manualLayout>...</c:manualLayout>
+       * </c:layout></c:title></c:catAx>` (or `<c:valAx>` / `<c:dateAx>` /
+       * `<c:serAx>`) — Excel's "Format Axis Title -> Title Options ->
+       * Position -> Custom" placement, the same drag-handle a user sees
+       * when grabbing the axis-title block in Excel's chart editor.
+       *
+       * The OOXML `CT_ManualLayout` block (ECMA-376 Part 1, §21.2.2.115)
+       * sits inside `CT_Title` between `<c:tx>` and `<c:overlay>` and
+       * carries the title's `(x, y)` anchor and `(w, h)` size as
+       * fractions of the chart frame in the `0..1` band. Each of
+       * {@link ChartManualLayout.x} / {@link ChartManualLayout.y} /
+       * {@link ChartManualLayout.w} / {@link ChartManualLayout.h} is
+       * independently optional — pin only the position
+       * ({@link ChartManualLayout.x} / {@link ChartManualLayout.y}) and
+       * let the title keep its automatic size, only the size
+       * ({@link ChartManualLayout.w} / {@link ChartManualLayout.h}) and
+       * let it keep its automatic anchor, or any combination.
+       *
+       * Mirrors {@link SheetChart.legendLayout} /
+       * {@link SheetChart.plotAreaLayout} for axis titles — same
+       * {@link ChartManualLayout} shape, same accept-or-drop grammar
+       * (out-of-range / non-finite / non-numeric coordinates collapse
+       * to `undefined` axis-by-axis), same canonical `xMode="edge"`
+       * normalization on emit. The writer always emits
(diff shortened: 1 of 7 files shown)


productdevbook:
## Review

Verified locally after rebase onto `main` (PR was conflicting with #300 which added the chart-level `titleLayout` describe block at end of `test/charts.test.ts`).

**Conflict resolution**: `test/charts.test.ts` — both branches appended new `describe` blocks at end-of-file. Resolved by keeping the merged-`main` `parseChart — titleLayout` block (from #300) and appending the PR's `parseChart — axisTitleLayout` block after it. Both blocks are independent test suites; no semantic interaction.

**Implementation review**:
- Reader (`parseAxisTitleLayout`): scoped to `<c:title> -> <c:layout> -> <c:manualLayout>` chain; admits both `xMode="edge"` and `xMode="factor"`; collapses to `undefined` when every coordinate drops on the `0..1` filter — symmetric with `parseLegendLayout` / `parsePlotAreaLayout`.
- Writer: `<c:layout>` emitted between `<c:tx>` and `<c:overlay>` per `CT_Title` schema sequence; threaded through `buildBarAxes` (covers bar/column/line/area) and `buildScatterAxes`. Pie/doughnut have no axes so the field has no slot — confirmed in `buildPlotArea` switch.
- Clone-through: gated on resolved title presence (`xTitle === undefined` collapses `xAxisTitleLayoutResolved` to `undefined`), matching the title-presence gate on the other axis-title knobs (overlay/strike/underline/etc.).
- Reuses `ChartManualLayout` type and `normalizeManualLayout` helper from #298/#299/#300 — no new shape introduced.

**Verification**:
- `pnpm typecheck` — passes
- `pnpm lint` — 0 errors, 45 pre-existing warnings (unchanged)
- `pnpm vitest run --dir=test` — 6333 / 6333 pass (89 files)
- CI build — green

LGTM.

PR #309, opened 2026-05-07
feat(xlsx): chart plot-area border color — read, write, clone-through

productdevbook:
## Summary
- Adds `plotAreaBorderColor` knob mapping to `<c:plotArea><c:spPr><a:ln><a:solidFill><a:srgbClr val=\"..\"/></a:solidFill></a:ln></c:spPr></c:plotArea>` — Excel's "Format Plot Area → Border → Solid line → Color" picker.
- Composes independently with `plotAreaFillColor`: the two share the same `<c:spPr>` host on `<c:plotArea>` but land on different children (`<a:solidFill>` for fill, `<a:ln>` for stroke). The writer emits a single `<c:spPr>` whenever either knob is set, in `CT_ShapeProperties` schema order.
- Reader, writer, and clone-through plumbing follows the established hex-color grammar — accepts `#` / case, drops malformed / non-solid / `schemeClr` / alpha-channel forms to `undefined`.

Pivot from the `<a:solidFill>` family (#302–#308) to the `<a:ln>` stroke family on the same `<c:plotArea>` host. First step of stroke knobs (`<c:spPr><a:ln>`) — border width and the legend / title / chartSpace / axisTitle / dataTable / dataLabels border-color counterparts can follow in separate PRs.

## Test plan
- [x] `pnpm vitest run --dir=test` — 89 files / 6671 tests pass (including 47 new tests across `parseChart`, `writeChart`, and `cloneChart`).
- [x] `pnpm build` — succeeds.
- [x] Round-trips covered: writer → reader, writer → `writeXlsx` → packaged `chart1.xml` → reader, plus clone-through with override / null / inheritance / malformed-token / cross-family flatten.
- [x] Composition with `plotAreaFillColor` and `plotAreaLayout` — single `<c:spPr>`, correct child order, layout precedes spPr per `CT_PlotArea`.

Refs #302

PR #310, opened 2026-05-07
feat(xlsx): chart legend border color — read, write, clone-through

productdevbook:
## Summary
- Adds `legendBorderColor` knob mapping to `<c:legend><c:spPr><a:ln><a:solidFill><a:srgbClr val=\"..\"/></a:solidFill></a:ln></c:spPr></c:legend>` — Excel's "Format Legend → Border → Solid line → Color" picker.
- Composes independently with `legendFillColor`: the two share the same `<c:spPr>` host on `<c:legend>` but land on different children (`<a:solidFill>` for fill, `<a:ln>` for stroke). The writer emits a single `<c:spPr>` whenever either knob is set, in `CT_ShapeProperties` schema order.
- Reader, writer, and clone-through plumbing follows the established hex-color grammar — accepts `#` / case, drops malformed / non-solid / `schemeClr` / alpha-channel forms to `undefined`. Silently dropped when `legend === false` (no `<c:legend>` element to host the slot).

Sibling of the in-flight `plotAreaBorderColor` knob — same `<a:ln>` pattern, distinct host element (`<c:legend>` vs `<c:plotArea>`). Continues the stroke (`<c:spPr><a:ln>`) family started by #309.

## Test plan
- [x] `pnpm vitest run --dir=test` — 89 files / 6716 tests pass (45 new tests across `parseChart`, `writeChart`, and `cloneChart`).
- [x] Round-trips covered: writer → reader, writer → `writeXlsx` → packaged `chart1.xml` → reader, plus clone-through with override / null / inheritance / malformed-token / cross-family flatten.
- [x] Composition with `legendFillColor` and `legendLayout` — single `<c:spPr>`, correct child order (`<a:solidFill>` precedes `<a:ln>`), layout precedes spPr per `CT_Legend`.
- [x] Hidden-legend scoping verified — `legend: false` produces no `<c:legend>` element even with the border pinned; legend `<c:delete val=\"1\"/>` drops the parsed value.

Refs #309

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -2297,6 +2297,55 @@ export interface SheetChart {
    */
   legendFillColor?: string;
   /**
+   * Legend border (stroke) solid color as a 6-digit RGB hex string
+   * (e.g. `"1F77B4"`). Maps to `<c:legend><c:spPr><a:ln><a:solidFill>
+   * <a:srgbClr val="RRGGBB"/></a:solidFill></a:ln></c:spPr></c:legend>`
+   * (CT_Legend, ECMA-376 Part 1, §21.2.2.114) — Excel's "Format Legend
+   * -> Border -> Solid line -> Color" picker. The OOXML `<a:srgbClr
+   * val=".."/>` carries the 6-character uppercase hex sRGB color
+   * (`CT_SRgbColor` inside the line's solid fill choice — ECMA-376
+   * Part 1, §20.1.2.3.32 / §20.1.2.3.24). The `<c:spPr>` slot sits
+   * between `<c:overlay>` and `<c:txPr>` per the CT_Legend schema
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Reviewed: legend border color knob is well-implemented across reader, writer, and clone-through paths. The `<a:ln>` slot composes cleanly with the existing `<a:solidFill>` fill in the shared `<c:spPr>` host on `<c:legend>`, with correct CT_ShapeProperties child ordering. Hidden-legend scoping (suppress when `legend === false`) is consistent with the rest of the legend family. Hex normalization, malformed-token drops, and override semantics (undefined / null / string) match the established pattern. Pushed an oxfmt fix for the test file to clear the lint check; CI is now green. LGTM.

PR #311, opened 2026-05-07
feat(xlsx): chart title border color — read, write, clone-through

productdevbook:
## Summary
- Adds `titleBorderColor` knob mapping to `<c:title><c:spPr><a:ln><a:solidFill><a:srgbClr val=\"..\"/></a:solidFill></a:ln></c:spPr></c:title>` — Excel's "Format Chart Title → Border → Solid line → Color" picker.
- Composes independently with `titleFillColor`: the two share the same `<c:spPr>` host on `<c:title>` but land on different children (`<a:solidFill>` for fill, `<a:ln>` for stroke). The writer emits a single `<c:spPr>` whenever either knob is set, in `CT_ShapeProperties` schema order (fill before stroke).
- Reader, writer, and clone-through plumbing follows the established hex-color grammar — accepts `#` / case, drops malformed / non-solid / `schemeClr` / alpha-channel forms to `undefined`. Silently dropped when no title is rendered (`showTitle === false` or `title` is absent — there is no `<c:title>` element to host the slot).

Sibling of the in-flight `legendBorderColor` knob (#310) — same `<a:ln>` pattern, distinct host element (`<c:title>` vs `<c:legend>`). Continues the stroke (`<c:spPr><a:ln>`) family started by #309.

## Test plan
- [x] `pnpm vitest run --dir=test` — 89 files / 6773 tests pass (58 new tests across `parseChart`, `writeChart`, and `cloneChart`).
- [x] Round-trips covered: writer → reader, writer → `writeXlsx` → packaged `chart1.xml` → reader, plus clone-through with override / null / inheritance / malformed-token / cross-family flatten.
- [x] Composition with `titleFillColor`, `titleColor`, and `titleLayout` — single `<c:spPr>`, correct child order (`<a:solidFill>` precedes `<a:ln>`), layout precedes spPr per `CT_Title`.
- [x] Hidden-title scoping verified — `showTitle: false` produces no `<c:title>` element even with the border pinned; `title: null` clone drops the border.
- [x] strRef-bodied title (formula reference, no `<c:rich>` body) still surfaces the border on parse.

Refs #309

### DIFF
--- a/src/_types.ts
+++ b/src/_types.ts
@@ -2880,6 +2880,55 @@ export interface SheetChart {
    */
   titleFillColor?: string;
   /**
+   * Chart title border (stroke) solid color as a 6-digit RGB hex string
+   * (e.g. `"1F77B4"`). Maps to `<c:title><c:spPr><a:ln><a:solidFill>
+   * <a:srgbClr val="RRGGBB"/></a:solidFill></a:ln></c:spPr></c:title>`
+   * — Excel's "Format Chart Title -> Border -> Solid line -> Color"
+   * picker. The OOXML `<a:srgbClr val=".."/>` carries the 6-character
+   * uppercase hex sRGB color (`CT_SRgbColor` inside the line's solid
+   * fill choice — ECMA-376 Part 1, §20.1.2.3.32 / §20.1.2.3.24). The
+   * `<c:spPr>` slot lives between `<c:overlay>` and `<c:txPr>` /
+   * `<c:extLst>` per CT_Title (ECMA-376 Part 1, §21.2.2.210); `<a:ln>`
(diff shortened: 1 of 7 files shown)


productdevbook (commented):
Self-review: titleBorderColor knob on the chart-level <c:title><c:spPr><a:ln><a:solidFill><a:srgbClr> slot — read, write, clone-through. Implementation mirrors titleFillColor / plotAreaBorderColor patterns; CT_ShapeProperties schema order (fill before stroke) is preserved when both knobs are set; reader is scoped to direct children of <c:title> to avoid cross-contamination from plot/series/legend <c:spPr>; theme-color references and non-solid line fills correctly collapse to undefined; clone resolver handles undefined/null/string overrides; comprehensive tests cover writer/parser/clone paths plus end-to-end round-trips. CI green.

PR #332, opened 2026-06-09
fix(xlsx): emit externalReferences before definedNames in workbook.xml

lrecknagel:
## Problem

`writeWorkbookXml` emits the `<definedNames>` block **before** `<externalReferences>` in `xl/workbook.xml`. The OOXML **CT_Workbook** sequence (ECMA-376 §18.2.27) is fixed:

```
sheets → functionGroups → externalReferences → definedNames → calcPr
```

`externalReferences` must come **before** `definedNames`.

Lenient readers (Apple Numbers, macOS QuickLook, Slack preview, openpyxl) ignore element order and open such files fine — but **Excel validates the sequence strictly and reports the workbook as corrupt** ("We found a problem with some content… do you want us to recover"). This is especially fatal when a defined name dereferences the external book, e.g.

```xml
<definedName name="Lst_FaktorenHeizung">[1]Daten!$C$183:$C$192</definedName>
```

since the `[1]` reference resolves against an `<externalReference>` that has not been declared yet at that point in the document.

This was found in the wild: a hucre-generated workbook (`<Application>hucre</Application>`) combining named ranges + an external link opened everywhere except Excel. Reordering these two elements (only) makes Excel open it without the repair dialog.

## Fix

- Move the `externalReferences` emission above the `definedNames` block in `writeWorkbookXml`.
- Correct the inline comment, which previously claimed (backwards) that the spec places `externalReferences` *after* `definedNames`.

The resulting order is now:

```
workbookPr → bookViews → workbookProtection → sheets → externalReferences → definedNames → calcPr → pivotCaches → extLst
```

## Tests

Adds two regression tests in `test/ooxml-spec-compliance.test.ts`, mirroring the existing `styles.xml` ordering test:

- asserts `sheets < externalReferences < definedNames < calcPr` when both blocks are present (fails on the old ordering)
- asserts `definedNames < calcPr` and no `externalReferences` when there is no external link

Full suite green: lint + typecheck + **7559 tests pass**. Verified the new ordering test fails on the pre-fix source and passes after.

### DIFF
--- a/src/xlsx/workbook-writer.ts
+++ b/src/xlsx/workbook-writer.ts
@@ -92,6 +92,20 @@ export function writeWorkbookXml(
 
   parts.push(xmlElement("sheets", undefined, sheetElements))
 
+  // ── externalReferences — must precede definedNames. The CT_Workbook
+  // sequence in ECMA-376 §18.2.27 is: sheets → functionGroups →
+  // externalReferences → definedNames → calcPr. Defined names that
+  // dereference an external book (e.g. "[1]Sheet!$A$1") need that book's
+  // <externalReference> declared first, or Excel rejects the workbook as
+  // corrupt — even though lenient readers (Numbers, QuickLook, openpyxl)
+  // ignore element order and open it fine.
+  if (externalLinkRels && externalLinkRels.length > 0) {
+    const refChildren = externalLinkRels.map((r) =>
+      xmlSelfClose("externalReference", { "r:id": r.rId }),
+    )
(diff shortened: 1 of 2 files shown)


productdevbook:
Thank you for the fix and the thorough write-up! Verified the CT_Workbook ordering against the spec and ran the full suite locally.the regression tests are a great addition.

lrecknagel:
Thanks for the fast review!
Awesome project - thanks so much for your effort!

mrxhaferzaysuz-coder (approved):

PR #335, opened 2026-06-10
fix: reader parity, date/format correctness, CSV edge cases, CLI packaging

productdevbook:
CONSISTENCY/CORRECTNESS wave. Each fix verified in code, with a focused regression test. `pnpm typecheck`, `pnpm lint`, and `vitest run` all pass (7597 tests, 21 new).

## Fixes

1. **CLI packaging (critical)** — `src/cli.ts` imports `citty` + `consola`, but they were only in `devDependencies` and `obuild` transform mode does not bundle, so installed CLI consumers hit `Cannot find module 'citty'`. Moved both to `dependencies`. CLI `meta.version` now read from `package.json` (was hardcoded `"0.0.1"`, package is `0.6.0`). Fixed stale `defter` references in CLI header comments.
2. **`m`/`mm` minutes** — `src/_date.ts formatDate` now treats `m`/`mm` as minutes when adjacent to a *following* `s`/`ss` (not only a preceding `h`/`hh`), so the built-in `mm:ss` renders minutes. Test: `formatValue(0.5, "mm:ss") === "00:00"`.
3. **1904 date system** — added `is1904` to `FormatOptions` and threaded it through `formatValue` → `dateToSerial`/`serialToDate`. Test: `formatValue(0, "yyyy-mm-dd", { is1904: true }) === "1904-01-01"`.
4. **Elapsed time `[h]`/`[m]`/`[s]`** — `tokenize` now emits the bracketed elapsed tokens and `formatDate` renders total hours/minutes/seconds (can exceed 24/60). Test: `formatValue(1.5, "[h]:mm") === "36:00"`.
5. **Implicit column (batch reader)** — `src/xlsx/worksheet.ts` now handles `<c>` elements without an `r` attribute via implicit per-row column position, matching the streaming reader. Regression test builds a worksheet whose later cells omit `r` and asserts both readers return `[1,2,3]`.
6. **Lenient OOXML rel matching** — exported `matchesRelType` from `reader.ts`; `stream-reader.ts` and `xlsb/reader.ts` now reuse it instead of exact Transitional-namespace compares, so Strict-namespace workbooks resolve. Test reads a Strict-namespace workbook through both paths.
7. **Out-of-range SST index** — batch reader (`worksheet.ts`) now returns `null` (was the raw index string); both readers consistent. Test asserts `null` from both.
8. **CSV trailing quoted-empty row** — `parseCsv` and `streamCsvRows` preserve a final row that is a single quoted-empty field (`a,b\n""`). Test included.
9. **CSV quoted comment** — comment filtering only applies to physically-unquoted leading `#`; a quoted `"#not a comment"` field is preserved (both readers). Test included.
10. **`streamCsvRows` option parity** — now honors `maxRows`, `skipLines`, and `preserveLeadingZeros` (aligning its `inferType` default with `parseCsv`). `transformValue`/`transformHeader`/`onRow`/`fastMode` remain unwired with a clear TODO comment. Tests for `maxRows`, `skipLines`, leading zeros.
11. **CLI `validate --sheet` NaN guard** — `Number("abc")` is now caught with the same `Number.isNaN` guard the `inspect` command uses, instead of throwing a raw `TypeError`.

## Skipped
None of the 11 items were skipped. No dedicated CLI integration test was added for items 1/11 (no existing CLI test harness; the commands call `process.exit`), but both code fixes are straightforward and mirror existing patterns.

## Notes
- `pnpm-lock.yaml` updated as a result of moving `citty`/`consola` to `dependencies`.
- A pre-existing `dist/limits.mjs` ERR_MODULE_NOT_FOUND from `obuild` transform mode reproduces on clean `main` and is out of scope for this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,12 +75,14 @@
     "release": "pnpm test && pnpm build && bumpp --commit --tag --push --all",
     "prepack": "pnpm build"
   },
+  "dependencies": {
+    "citty": "^0.2.1",
+    "consola": "^3.4.2"
+  },
   "devDependencies": {
     "@typescript/native-preview": "7.0.0-dev.20260316.1",
     "@vitest/coverage-v8": "^4.1.1",
     "bumpp": "^11.0.1",
-    "citty": "^0.2.1",
-    "consola": "^3.4.2",
     "obuild": "^0.4.32",
     "oxfmt": "^0.42.0",
     "oxlint": "^1.57.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -7,22 +7,23 @@ settings:
 importers:
 
   .:
-    devDependencies:
-      '@typescript/native-preview':
-        specifier: 7.0.0-dev.20260316.1
-        version: 7.0.0-dev.20260316.1
-      '@vitest/coverage-v8':
-        specifier: ^4.1.1
-        version: 4.1.1(vitest@4.1.1(vite@8.0.2(jiti@2.6.1)(yaml@2.8.3)))
-      bumpp:
-        specifier: ^11.0.1
-        version: 11.0.1
+    dependencies:
       citty:
         specifier: ^0.2.1
         version: 0.2.1
       consola:
         specifier: ^3.4.2
         version: 3.4.2
+    devDependencies:
+      '@typescript/native-preview':
+        specifier: 7.0.0-dev.20260316.1
+        version: 7.0.0-dev.20260316.1
+      '@vitest/coverage-v8':
+        specifier: ^4.1.1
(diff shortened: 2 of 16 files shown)


productdevbook:
Also folded in a build-regression hotfix: the #333 security wave added two exports without explicit type annotations (`MAX_DECOMPRESSED_BYTES` and `inflate`'s `maxBytes` default param), which caused obuild's `--isolatedDeclarations` .d.ts generation to silently drop emit for `dist/limits.mjs` and `dist/zip/deflate.mjs` — breaking the published package at runtime. Fixed with explicit `: number` annotations; `pnpm build` now emits all files.

PR #375, opened 2026-08-03
fix: saveXlsx loses sheet background images, and the obvious fix emits a corr...

### LINKED ISSUE #367: saveXlsx loses sheet background images, and the obvious fix emits a corrupt file
Split out of #359, where I tried to fix it and found that the obvious fix makes things worse.

## What happens today

`saveXlsx` does not forward `Sheet.backgroundImage` into the `WriteSheet` it rebuilds, so a workbook with a sheet background loses it on open → save. Straightforward data loss.

## Why it is not a one-line map entry

Adding `backgroundImage: sheet.backgroundImage` to the map at `src/xlsx/roundtrip.ts:181` makes `writeWorksheetXml` emit a `<picture r:id="rIdN"/>` element. But `roundtrip.ts` has **no** `pictureRId` handling anywhere — a repo-wide grep for `pictureRId` in that file returns nothing.

Measured on the output with the field forwarded:

```
saved entries:                    xl/worksheets/sheet1.xml
saved sheet has picture element:  true
```

No `xl/media/image1.png`, no `xl/worksheets/_rels/sheet1.xml.rels`. The sheet references a relationship that was never written. Excel reports that as unreadable content — strictly worse than losing the image.

So the field stays out of the map, with a test pinning that, until the plumbing below exists.

## What the fix needs

`src/xlsx/writer.ts` already does all of it, and can be followed:

1. **Media index allocation** — `writer.ts:186-199` builds `backgroundImagePaths[]`, allocating `xl/media/image{globalImageIndex}.png` per sheet and registering the `png` extension in `imageExtensions`. In `roundtrip.ts` this has to share `globalImageIndex` with the drawing loop so background images and drawing images do not collide.
2. **Per-sheet relationship** — `writer.ts:449-452` emits the picture `Relationship` with `Target: "../media/imageN.png"`. `roundtrip.ts` has an equivalent per-sheet rels block (around lines 846-1000) that needs the same branch, including adding `hasPicture` to the condition that decides whether a rels part is emitted at all.
3. **The media entry itself** — the bytes have to be added to the ZIP.
4. **Content types** — `imageExtensions` already feeds `ContentTypesOptions`, so registering `png` in step 1 covers the `Default` extension entry.

The subtlety is step 1: `roundtrip.ts` currently increments `globalImageIndex` only inside the drawing loop, so background images have to be woven into the same counter rather than numbered independently.

## Test to add

The round-trip test that currently pins the drop should flip to asserting the image survives. The content-types integrity test already added in #366 — walking every `PartName` and asserting the entry exists — would have caught the dangling-reference version on its own, and should stay.

yhay81:
Fixes #367.

## Summary

saveXlsx loses sheet background images, and the obvious fix emits a corrupt file

## Validation

- Mechanical gate: +37/-13, tests passed
- Adversarial review: approved

---
🤖 AI-authored PR, operated by @yhay81.

### DIFF
--- a/src/xlsx/roundtrip.ts
+++ b/src/xlsx/roundtrip.ts
@@ -71,6 +71,7 @@ const NS_RELATIONSHIPS = "http://schemas.openxmlformats.org/package/2006/relatio
 const REL_HYPERLINK =
   "http://schemas.openxmlformats.org/officeDocument/2006/relationships/hyperlink"
 const REL_DRAWING = "http://schemas.openxmlformats.org/officeDocument/2006/relationships/drawing"
+const REL_IMAGE = "http://schemas.openxmlformats.org/officeDocument/2006/relationships/image"
 const REL_COMMENTS = "http://schemas.openxmlformats.org/officeDocument/2006/relationships/comments"
 const REL_VML_DRAWING =
   "http://schemas.openxmlformats.org/officeDocument/2006/relationships/vmlDrawing"
@@ -206,11 +207,7 @@ export async function saveXlsx(
     outlineProperties: sheet.outlineProperties,
     sparklines: sheet.sparklines,
     textBoxes: sheet.textBoxes,
-    // `backgroundImage` is deliberately NOT forwarded. It needs a media
-    // part and a `<picture>` relationship, and this path emits neither —
-    // forwarding it produces a sheet referencing an rId that was never
-    // written, which Excel reports as corrupt. Losing the image is worse
-    // than it should be, but better than that. Tracked separately.
+    backgroundImage: sheet.backgroundImage,
   }))
 
   // Create shared collectors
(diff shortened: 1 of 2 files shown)


productdevbook:
Thanks for this — closing as superseded, not as wrong.

#367 was fixed in #380, which landed while this was open. I compared the two implementations line by line and they reach the same design independently: forward `backgroundImage` into the `writeSheets` map, allocate the media path off the **shared** `globalImageIndex` (so it cannot collide with drawing images), add it to `regeneratedPaths`, emit the `REL_IMAGE` relationship keyed on `result.pictureRId`, and gate the whole thing on `hasPicture`. Even the `Target: '../' + path.slice(3)` trick to strip the `xl/` prefix is the same. The branch is now `CONFLICTING` against those same two files.

Two things worth passing back, since the analysis was sound:

**The test change here is the right instinct.** The `"drops the background image rather than emitting a dangling reference"` test was pinning a deliberate gap, not a defect — it existed so the gap stayed a decision rather than drift. Replacing it once the plumbing lands is exactly correct, and #380 did the same. What #380 added beyond it is a `"does not collide a background image with drawing images"` case (they share the `xl/media` namespace) and a generic `"ships every part its content types declare"` integrity test, which is the one that would have caught the original corrupt-file failure without anyone thinking to look for it.

**Both implementations shared a real bug, and yours helped find it.** Reviewing this PR against `main` is what surfaced #427: the media path hard-codes `.png` for every background image, so a JPEG was stored at `xl/media/image1.png` and declared `image/png`. Verified:

```
bytes at image1.png : magic = 0xff 0xd8 0xff   (JPEG)
```

Excel sniffs image bytes and renders it anyway, which is why it survived review twice. Fixed in #428 — the format is now sniffed from the bytes, and `writeXlsx` and `saveXlsx` share one helper instead of a copy each, since two copies is how they came to agree on one bug in the first place.

So this PR ends up having earned its keep, just not by merging. Sorry for the wasted round trip.

Issue #494, opened 2026-08-13
readXls and readXlsb return ragged rows where readXlsx pads to the sheet width


erannave:
Found while building the Excel-written corpus for #464.

`test/fixtures/excel-basic.{xlsx,xls,xlsb}` are **one authored sheet saved three ways** by the same Excel. Most rows reach column E; the last row has values only in A–C. The three readers disagree about its shape:

| reader | last row |
| --- | --- |
| `readXlsx` | `["Broken", "#DIV/0!", "xy", null, null]` |
| `readXls` | `["Broken", "#DIV/0!", "xy"]` |
| `readXlsb` | `["Broken", "#DIV/0!", "xy"]` |

`readXlsx` normalizes every row to the sheet's column count — the explicit "Ensure all rows have consistent length" pass at `src/xlsx/worksheet.ts:1148`, which computes `colCount = maxCol + 1` and fills every row up to `maxRow`. `readXls` (`src/xls/reader.ts`) and `readXlsb` (`src/xlsx/xlsb/reader.ts`) have no equivalent: their `setCell` pads a row only up to *that row's own* last written column, and nothing pads afterwards.

**The strongest evidence is in the XLS reader itself.** `src/xls/reader.ts:190-191` says:

> BIFF row/col are u16, so each is bounded at 65,535 on its own — but their product is not, and `rows` is a dense rectangle.

and it sizes its cell-count guard on `widestCol` accordingly — but never enforces the density on output. The invariant is assumed by the guard and not implemented by the writer of the rows.

### It also leaves `undefined` holes

Same root cause, and worse than a shape difference: in both readers `rows[row]` is only ever created by `setCell`, so a row with no cell records at all is never allocated. A sheet with cells on rows 0 and 2 and nothing on row 1 comes back with `rows[1] === undefined`:

```
len=3  row1=UNDEFINED-HOLE
```

`CellValue` (`src/_types.ts:3`) is `string | number | boolean | Date | null` — it has no `undefined` member, so a hole is a runtime type violation, not merely an inconsistency. `readXlsx`'s final pass fixes this incidentally by filling any row index with no data.

### On the documentation

[...]

Issue #493, opened 2026-08-13
readXlsx does not apply XML line-ending normalization, so an Excel-written newline comes back as CRLF


erannave:
Found while building the Excel-written corpus for #464.

`test/fixtures/excel-strings.xlsx` has a cell authored in Excel as `"line one" & Chr(10) & "line two"`. Excel wrote that shared string with a **literal CRLF** inside the `<t>` element — the bytes are `6c 69 6e 65 20 6f 6e 65 0d 0a 6c 69 6e 65 20 74 77 6f`. `readXlsx` returns `"line one\r\nline two"`.

XML 1.0 §2.11 ("End-of-Line Handling") requires an XML processor to normalize a literal CRLF, and a lone CR, to a single LF **before** passing content to the application. So the correct value is `"line one\nline two"`. `src/xml/parser.ts` does no such normalization — the only place hucre touches CR at all is `src/xml/writer.ts:103`, which escapes it as `&#13;` on output. The writer is spec-aware; the parser is not.

The asymmetry is visible in the corpus: `test/fixtures/excel-strings.xlsb` is the *same authored workbook* saved by the same Excel as XLSB, which stores a bare LF, and `readXlsb` returns `"line one\nline two"`. One authored cell, two containers, two different strings out.

This is exactly the class of bug #464 predicts: every existing multi-line-string test builds its own XML with `\n`, so the reader has never been shown the CRLF that Excel actually emits.

Marked in `test/real-files.test.ts` as a known defect rather than fixed here — the corpus PR does not change reader behaviour.

erannave:
Two additions after a closer read of the parser, for whoever picks this up:

- **The streaming path needs more care than the buffered one.** `parseSax` (`src/xml/parser.ts:147-151`) strips the BOM and hands raw text runs to `decodeEntities` with no EOL pass, so the buffered fix is a normalization right there. But `processSaxBuffer` (~line 362) has the same gap *and* a chunk-boundary problem: a literal `\r` ending one chunk with `\n` starting the next must not become two line breaks. There is already a holdback mechanism for exactly this class of problem — `safeTextSplit` (~line 342), which holds back a partial entity — and a CR holdback would mirror it.
[...]

Issue #492, opened 2026-08-13
Style-only cells inflate the materialized grid: 197 values become 45×16,126 nulls (727 KB of commas from 1.8 KB of data)


taoche:
## What happened

A real-world packing-list workbook (written by Excel, edited by humans over years) has a sheet where formatting was once applied across the full row width. Its worksheet XML:

- declared `<dimension ref="A4:WVF45"/>` — WVF is column **16,126**
- 145,512 `<c>` elements, of which **145,315 are self-closing style-only cells** (`<c r="…" s="…"/>`, no value)
- only **197 cells carry a value**; the right-most value sits in column **14**

`readXlsx` returns `rows` as 45 × 16,126 — every row padded with nulls to the bounding box. Piping that through `writeCsv` yields **727,211 bytes**, ~99.75% of it bare commas. The actual data is 1.8 KB.

## Why

`src/xlsx/worksheet.ts` bumps `maxCol`/`maxRow` on every `</c>`, including value-less style-only cells, then materializes the full bounding box (`colCount = maxCol + 1`).

This isn't a spec violation — the file is valid OOXML and the cells do exist. But under the default `readStyles: false`, those 145k cells contribute **nothing to the returned data**: their styles aren't read, so their only effect is null padding the caller cannot distinguish from never-written cells anyway.

`MAX_TOTAL_CELLS` guards the OOM case but not this one (45 × 16,126 ≈ 726k cells is under the limit). `range`/`maxRows` require knowing the data extent in advance, which is exactly what the caller doesn't know.

## Proposal

When `readStyles` is false, don't let a value-less cell extend `maxCol`/`maxRow`. Interior positions are untouched — a value at N5 still lands at `rows[4][13]` with nulls before it — so this is not the #394 class of bug (interior rows/cells must stay). Only the trailing bounding box shrinks to the last cell that actually carries data. With `readStyles: true` nothing changes: style-only cells carry information there and should keep counting.

Alternative if changing the default is unacceptable: an opt-in read option (`ignoreStyleOnlyCells` / `trimToData`).

## Repro

```python
# builds a minimal xlsx whose sheet has one value at A1 and one style-only cell at WVF45
import zipfile
[...]

PR #495, opened 2026-08-13
test: a corpus of files hucre did not write

### LINKED ISSUE #464: Test against files hucre did not write
Split out of #439 §G1. The single highest-value item on that list, and the one thing the 23 PRs could not do.

## The problem

```
$ find test -type f ! -name '*.ts' ! -name '*.snap' | wc -l
0
```

9,934 tests, 181 files, and **not one byte produced by Excel, LibreOffice, Google Sheets, SheetJS or ExcelJS** is ever parsed. Every assertion is hucre writing something and hucre reading it back.

A writer bug the reader mirrors is invisible. That is not hypothetical — three of the defects fixed this round were exactly that shape, and all three sat in `main` at 98.8% statement coverage:

| | why the suite could not see it |
| --- | --- |
| `xml:space` missing on inline strings (#441) | hucre's reader does not trim, so the round trip agreed with itself. Only Excel disagrees. |
| XLS/XLSB reading CJK dates as raw serials (#444) | the date-id table was 61% complete; no real `.xls` was ever opened |
| `saveXlsx` emitting a duplicate part (#460) | `ZipWriter` accepted the duplicate and `extract` returned the last one |

The XLS (BIFF8) and XLSB readers are the sharpest case: they exist *solely* to consume other tools' output, and are validated only against byte arrays the tests construct themselves.

## What would help

A small, licence-clean corpus under `test/fixtures/` — a dozen files is plenty — with a golden-model test per file. The shapes that matter, roughly in order:

- Excel, LibreOffice and Google Sheets each writing the same simple sheet: strings, numbers, dates, booleans, a formula with a cached result
- shared strings **and** inline strings, including values with leading/trailing whitespace
- a styled sheet: fonts, fills, borders, number formats, a column-level format
- merges, a frozen pane, a conditional rule
- a `.xls` and a `.xlsb` from a real Excel, ideally one with a CJK locale so the date block is exercised
- one file with a `<tfoot>`-style oddity or an unusual paper size

Licensing is the thing to get right: files someone on the project authored, or something already under a permissive licence. Not scraped documents.

## Why it is not in the batch

There is no Excel, LibreOffice or `openpyxl` in the environment those PRs were written in, so no genuinely third-party file could be produced. Generating fixtures with hucre would defeat the entire point.

erannave:
Closes #464.

The issue's point is that hucre's ~10,000 tests never parse a byte produced by another tool: every binary input under `test/` is constructed by the test that reads it, so a writer bug the reader mirrors is invisible. This adds a corpus that hucre did not write, and a test that reads it.

Test-only. No reader or writer behaviour changes.

## What is in it

`test/fixtures/` — eleven workbooks, 135 KB of binaries, 149 KB for the directory including the golden models and `PROVENANCE.md`.

| file | shape |
| --- | --- |
| `excel-basic.xlsx` / `.xls` / `.xlsb` | one authored sheet saved three ways: strings, numbers, a negative, a leap day, booleans, formulas with cached numeric/string/error results |
| `excel-strings.xlsx` / `.xlsb` | shared-string table: leading/trailing whitespace needing `xml:space="preserve"` (#441), a reused entry, XML metacharacters, an embedded newline, tab and NBSP, several scripts, a non-BMP codepoint, a rich string |
| `excel-styled.xlsx` | fonts, a fill, borders, alignment, number formats, and a format on a column with no cells |
| `excel-layout.xlsx` | merges across and down, a frozen pane, a conditional-formatting rule, a hidden row and a hidden column |
| `excel-pagesetup.xlsx` | A3 landscape, fit-to-pages, a non-default margin, print area, print titles |
| `excel-styleonly.xlsx` | whole-row formatting with no values out to the right — the #492 shape |
| `excel-dates.xls` | BIFF number formats: built-ins, a time, a date-time, thousands, percent, text-formatted digits, CJK date codes |
| `excel-empty.xlsx` | one sheet, nothing in it |

`test/real-files.test.ts` reads each through the public API — the explicit `readXlsx` / `readXls` / `readXlsb` **and** `read()`'s format detection, which must agree — and compares against a per-fixture golden model.

`scripts/fixtures/make-fixtures.vbs` and `test/fixtures/PROVENANCE.md` are how the corpus was produced and how to regenerate it.

## Provenance

| | |
| --- | --- |
| Producer | Microsoft Excel 16.0 (Microsoft 365, Windows 11 x64), one producer for all eleven files |
| Driven by | the committed `make-fixtures.vbs`, via `cscript.exe`, late-bound COM |
| Content | **synthetic, authored for this corpus by me.** No third-party or confidential document is involved, and nothing was scrubbed down from one — a real workbook carries `docProps`, defined names, `calcChain`, revision ids and comment authors, and reproducing an interesting *shape* is easier than scrubbing an original reliably |
| Licence | same as the repo (MIT) |

Author metadata is blanked at the source with `Workbook.RemovePersonalInformation = True` before `SaveAs`, which clears `dc:creator` / `cp:lastModifiedBy` in xlsx and xlsb and the author fields of the BIFF `SummaryInformation` stream. Worth recording for anyone who repeats this: `Application.UserName` does **not** do it — Excel 16 takes the author from the signed-in Office identity, and setting `UserName` still produced a real name in `docProps/core.xml`.

## Golden models, not snapshots

The expected models are hand-written from what the `.vbs` authored and from what the raw files contain (unzip the xlsx, dump the BIFF `FORMAT` records) — **not** from running hucre and pasting the result, which would rebuild exactly the closed loop this is meant to break. No vitest snapshots either: binary-derived `.snap` churn is unreadable in review.

The projection is deliberately narrow for the same reason. It carries what a person can check against Excel — values and types, merges, frozen panes, hidden rows and columns, number formats, page setup, and the fact that no author survived — and drops what a person cannot, notably Excel's autofit column widths (`10.33203125`) and its theme palette. Asserting those would be asserting that hucre agrees with hucre.

Each golden may declare `knownDefects`, which swap one field out of the green comparison and re-assert it, unswapped, as an `it.fails` naming the issue. So a known defect is recorded loudly without masking the rest of the fixture.

## What it found

Two defects, first pass, in code sitting at 98.8% coverage:

- **#493** — `readXlsx` does not apply XML line-ending normalization. Excel wrote a newline as a **literal CRLF** inside `<t>`; XML 1.0 §2.11 requires a processor to normalize that to a single LF before the application sees it, and `src/xml/parser.ts` does not. `src/xml/writer.ts:103` escapes CR as `&#13;` on output, so the writer is spec-aware and the parser is not. The same authored cell read from the XLSB fixture gives `\n` — one workbook, two containers, two different strings.
- **#494** — `readXls` and `readXlsb` return ragged rows where `readXlsx` pads to the sheet width, for `excel-basic.{xlsx,xls,xlsb}` — one authored sheet saved three ways. `readXlsx` has an explicit "Ensure all rows have consistent length" pass (`src/xlsx/worksheet.ts:1148`); the binary readers have no equivalent. `src/xls/reader.ts:190-191` states in its own comment that "`rows` is a dense rectangle" and sizes its cell-count guard on that, but never enforces it on output. `row.length` is therefore not the sheet width for XLS/XLSB, so caller code written against an XLSX read changes behaviour with the container. The same root cause also leaves `undefined` holes for rows with no cell records, which `CellValue` does not admit.

Both are marked with `it.fails` and their issue number rather than fixed here. This is the corpus PR; each fix belongs in its own change with its own failing-test-first, per `CONTRIBUTING.md`. Happy to follow up with either.

Everything else came back correct, including some things I expected to break: BIFF8 dates and times, cached error and string formula results, `xml:space` whitespace, the non-BMP codepoint through both string tables, the conditional-format rule with its `dxf`, the frozen pane, the column-level number format with no cells under it, A3 landscape with print titles, and the #492 style-only shape (no blow-up — two rows out, as it should be).

## What #464 still wants

- **The BIFF CJK built-in format ids (27–36, 50–58) from #444.** `excel-dates.xls` carries Japanese, Japanese-era and Korean date codes, but a non-CJK Windows Excel writes them as *custom* `FORMAT` records — I dumped the records and got ids 164–169, nothing in the CJK built-in range. Reaching those ids needs an Excel running under a CJK locale.
- **LibreOffice and Google Sheets**, both named in the issue. Everything here is Excel. `openpyxl` would also be a cheap second producer — it is one line of the generation script — and I left it out only to keep this PR one producer, one story. Say the word and I will add either.

The directory layout is not load-bearing; if you want the fixtures arranged differently that is cheap to change. The fixtures themselves are the expensive part and are layout-independent.

## Checks

- `pnpm test` — 198 files, 10,160 passed, 2 expected fail (the two marked defects above).
- `pnpm coverage` — statements 98.69, branches 96.66, functions 98.61, lines 99.12, against thresholds 98.5 / 96 / 98 / 99.
- `test/real-files.test.ts` is registered in both `tsconfig.cli.json` `include` and `tsconfig.json` `exclude`, since it reads from disk.
- Deleting a fixture makes its tests fail, so the suite really is reading the bytes and not silently skipping.


erannave:
Follow-up opened as #498 — a second producer (openpyxl 3.1.5) for the same corpus, stacked on this branch. It answers the "LibreOffice / Google Sheets stay uncovered" note above by at least making the corpus two-producer rather than one, and it surfaced two more defects: #496 (ISO-8601 `t="d"` cells read as strings) and #497 (cached formula results that are strings or errors are dropped, and lost on round-trip).

#497 is worth a look even if you would rather not take a second producer: it is a real round-trip loss on ordinary Excel files, and the fixture that demonstrates it is `excel-basic.xlsx` from *this* PR.

productdevbook:
Thank you, my Claude tokens were running low :)

productdevbook:
Superseded by #500, which is merged — its tree contains this one exactly (`comm -23` between the two file lists is empty) plus the chartsheet and sparse fixtures.

Thank you for this. The corpus is the single thing #464 asked for that could not be produced from inside the environment the earlier batch was written in: no Excel, no LibreOffice. The `.xls` and `.xlsb` files matter most — those readers exist solely to consume other tools' output and had never seen any.

Two things in it I want to note as good practice rather than let pass silently: writing the golden models by hand from the authored content and the raw file, instead of pasting hucre's output, is the only version of this that is worth anything; and marking the failures `it.fails` with an issue number each, rather than fixing them in the corpus PR, keeps the failing-test-first discipline intact for each fix.

#498 is still open for the openpyxl half.

Issue #497, opened 2026-08-13
readXlsx drops cached formula results that are strings or errors, losing them on round-trip


erannave:
Found while adding a second producer to the corpus from #464 — surfaced by starting to assert cached formula results, not by the new producer itself.

`Cell.formulaResult` is assigned in exactly one place, `src/xlsx/worksheet.ts:1778`, inside the `n`/default (numeric) arm of the cell-type switch. The `str` arm (a formula whose cached result is text) and the `e` arm (a formula whose cached result is an error) both set `value` and never set `formulaResult`. So the cached result survives only when it happens to be a number.

`test/fixtures/excel-basic.xlsx` — written by Excel, which always caches — has all three kinds:

```
E2: type=formula value=24          formula="B2*2"        formulaResult=24
B5: type=error   value="#DIV/0!"   formula="1/0"         formulaResult=undefined
C5: type=formula value="xy"        formula="\"x\" & \"y\"" formulaResult=undefined
```

Note `B5.type` is `"error"` rather than `"formula"` even though the cell has a formula, which is a second inconsistency in the same switch.

### It is a real round-trip loss, not just a missing field

The writer already supports string cached results — `src/xlsx/worksheet-writer.ts:981-982` branches on `typeof formulaResult === "string"`. Only the reader fails to produce them, so `readXlsx` → `writeXlsx` drops them:

```
--- after write+read ---
E2: type=formula value=24    formula="B2*2"          formulaResult=24
B5: type=formula value=null  formula="1/0"           formulaResult=null
C5: type=formula value=null  formula="\"x\" & \"y\"" formulaResult=null
```

The numeric result round-trips. The error and the string are gone — the output has `<f>` with no `<v>`, so anything that opens it without recalculating (including hucre itself, and #495's `openpyxl-formulas.xlsx` shows exactly what that looks like) sees empty cells where Excel showed `#DIV/0!` and `xy`.

Repro is the three cells above from `test/fixtures/excel-basic.xlsx`. A fix is to set `formulaResult` in the `str` and `e` arms as the numeric arm does, and to decide whether a formula cell with an error result should report `type: "error"` or `type: "formula"` — currently it reports `"error"` on the way in and `"formula"` on the way back out, which cannot both be right.

Not fixed in the corpus PRs; marked there with `it.fails` naming this issue, as with #493, #494 and #496.

Issue #496, opened 2026-08-13
ISO-8601 date cells (t="d") are read as strings — ST_CellType 'd' has no case in the reader


erannave:
Found while adding a second producer (openpyxl) to the corpus from #464, as a follow-up to #495.

ECMA-376 Part 1 §18.18.11 (`ST_CellType`) defines `d` — "Cell containing a date in the ISO 8601 format" — alongside `b`, `e`, `inlineStr`, `n`, `s` and `str`. hucre's cell-type switch in `src/xlsx/worksheet.ts` has cases for every one of those **except `d`**.

A `t="d"` cell therefore falls through to the `n`/default branch, where `Number("2024-03-17")` is `NaN`, and lands in this arm:

```ts
} else if (valueText !== "") {
  // Non-numeric value text (shouldn't happen, but be safe)
  value = valueText
  cellType = "string"
}
```

The comment is accurate about the author's expectations and wrong about the format: it does happen. So an ISO-8601 date cell comes back as a **string**, not a `Date`.

`test/fixtures/openpyxl-isodates.xlsx` carries its own control. openpyxl writes `t="d"` whenever `iso_dates=True`:

| cell | serialization | `type` | `value` |
| --- | --- | --- | --- |
| B1 | `<c t="d"><v>2024-03-17</v></c>` | `string` | `"2024-03-17"` |
| B2 | `<c t="d"><v>2024-03-17T13:45:30</v></c>` | `string` | `"2024-03-17T13:45:30"` |
| B4 | `<c><v>45368</v></c>` under `numFmt yyyy-mm-dd` | `date` | `Date(2024-03-17)` |

B1 and B4 are the same day under the same number format, and read back as different types.

Consequences beyond the type: `readObjects` and schema validation see a string where a date was declared, date-aware writers round-trip it as text, and a `t="d"` cell that also carries a date `numFmt` still does not become a `Date` because the style check only runs in the numeric branch.

A fix is a `case "d":` that parses the ISO value — note it is epoch-independent, so `date1904` must **not** be applied to it, unlike the serial path. Bare times (`<v>13:45:30</v>`, which openpyxl emits for a `datetime.time`) are not valid ISO 8601 date-times, and leaving those as text seems right to me; the fixture includes one so whatever is decided is pinned by a test.

Not fixed in the corpus PR — marked there with `it.fails` naming this issue, per the same convention as #493 and #494.

PR #498, opened 2026-08-13
test: a second producer for the corpus (openpyxl)

erannave:
**Stacked on #495.** GitHub will not let a fork PR target another fork's branch — the base has to be a branch in this repo — so the diff below carries #495's commit as well. Two ways to read just the new work:

- **[Incremental diff](https://github.com/erannave/hucre/compare/test/real-file-corpus...test/openpyxl-corpus)** — 1 commit, 26 files, none of #495's noise.
- Or merge #495 first: this PR's diff then collapses to that one commit on its own, with nothing to do here.

Refs #464. If you would rather have one PR than two, say so and I will fold this into #495 and close this.

Test-only. No reader or writer behaviour changes.

## Why a second producer

#495 answered #464 with one producer, Microsoft Excel. One producer is a single point of agreement. If hucre and Excel happen to share a misreading of the spec, a corpus made only of Excel output agrees with itself about it and stays green — the same shape of blind spot #464 is about, one level up.

openpyxl was written by different people from an independent reading of ECMA-376, and it emits four things Excel never does:

| shape | Excel |
| --- | --- |
| `<f>A1*3</f><v/>` — a formula with no cached result, and an *empty* `<v/>` rather than an absent one | always caches a result |
| `t="d"` cells holding an ISO-8601 date (`ST_CellType` 'd') | always writes a serial number |
| `date1904="1"` | reachable only through a workbook setting most Windows users never touch |
| `t="inlineStr"` with `<is><t>` and **no `sharedStrings.xml` at all** | always writes a shared string table |

The last one is the inline-string case #441 wanted, from a producer that actually emits inline strings.

## What is in it

Six fixtures, ~5 KB each, from openpyxl 3.1.5 via the committed `scripts/fixtures/make-openpyxl-fixtures.py`. Content synthetic and contributor-authored, licence-clean, creator blanked, same as before. openpyxl writes `.xlsx` only, so `.xls` and `.xlsb` still have Excel as their only producer — noted in `PROVENANCE.md` as the thinnest remaining part of the corpus.

| file | shape |
| --- | --- |
| `openpyxl-basic.xlsx` | the value cells of `excel-basic.xlsx`, authored again by the other producer |
| `openpyxl-formulas.xlsx` | formulas with no cached result |
| `openpyxl-isodates.xlsx` | `t="d"` date, date-time and time cells, with a serial-number date under the same format as an in-fixture control |
| `openpyxl-1904.xlsx` | the 1904 epoch — the two epochs are 1,462 days apart, so dropping the flag cannot fail quietly |
| `openpyxl-inline-strings.xlsx` | inline strings, no string table, with the whitespace and unicode cases |
| `openpyxl-styled.xlsx` | the same style facets as `excel-styled.xlsx` through openpyxl's different `styles.xml` |

**`openpyxl-basic.xlsx` re-authors `excel-basic.xlsx`**, and the suite reads both and asserts they agree. Where two independent producers disagree, one of them is wrong and neither gets a free pass. (Column E is compared as values: Excel cached its formula results and openpyxl cannot evaluate, so the Python writes the same numbers as literals. The formula-with-no-result case is its own fixture, where it is the subject rather than a confound.)

The projection gains a `formulas` facet — `A1 → formula text => cached result`. That is the half worth asserting once two producers disagree about whether a cached result exists at all. It is why the existing goldens are touched.

## What it found

- **#496 — ISO-8601 date cells (`t="d"`) are read as strings.** `ST_CellType` 'd' has no case in the reader's cell-type switch, so the value falls through to the numeric arm, `Number("2024-03-17")` is `NaN`, and it lands in the arm commented `// Non-numeric value text (shouldn't happen, but be safe)`. It does happen. `openpyxl-isodates.xlsx` carries its own control: B1 (`t="d"`, ISO text) and B4 (serial under the same `yyyy-mm-dd` format) are the same day and read back as different types.
- **#497 — cached formula results that are strings or errors are dropped.** `formulaResult` is assigned in one place, inside the numeric arm; the `str` and `e` arms set `value` and never set it. The writer already handles string results, so `readXlsx` → `writeXlsx` loses them and emits `<f>` with no `<v>`. Demonstrated on `excel-basic.xlsx`: `=1/0` cached `#DIV/0!` and `="x" & "y"` cached `xy` both come back `null` after a round-trip, while the numeric `=B2*2` survives. Found by projecting cached results — which the openpyxl formula fixture made worth doing — not by openpyxl output itself.

Both marked `it.fails` with their issue number, not fixed here, same convention as #493 and #494.

Everything else openpyxl produces read correctly, including the two I most expected to break: **`date1904` is handled exactly right** — all four dates, epoch and leap day included — and inline strings with no string table work. The formula-with-no-cached-result case is also handled well: `cells` carries `formula` and a `null` `formulaResult`, which is the honest answer.

## Checks

- `pnpm test` — 198 files, 10,173 passed, 6 expected fail (the four issues above; #493 and #497 account for two each).
- `pnpm coverage` — 98.69 / 96.67 / 98.61 / 99.12 against thresholds 98.5 / 96 / 98 / 99.
- `test/fixtures` is now 165 KB of binaries, 195 KB including goldens and `PROVENANCE.md`.

## Still uncovered

LibreOffice and Google Sheets, both named in #464. LibreOffice is the obvious third: it writes `.xls`, which is the one part of the corpus still reading a single producer's output.


productdevbook:
Merged. I rebased it onto `main` myself rather than asking you to — #500 had landed in between and the two shared the Excel half, so the first commit was already in. What is merged is only the openpyxl addition.

Three conflicts, all resolved by keeping both sides:

- `real-files.test.ts` — your cross-producer `it` and #500's chartsheet/sparse blocks are complementary; both are in.
- `PROVENANCE.md` — merged the two defect lists into one of six (#493, #494, #496, #497, #499, #501) and kept the real-corpus narrative from #500.
- I re-measured rather than adding the two counts: **19 binaries, 181 KB, 285 KB for the directory**. Both branches' figures were written before the other landed.

## What I checked before merging

Since these carry binaries and a generator script, I reviewed them as untrusted input rather than reading the description and merging:

- No `package.json`, lockfile, `.github/`, `CLAUDE.md`, `.npmrc` or config change in any of the three PRs. Nothing executes at install or test time.
- The `.vbs` and `.py` create only `Excel.Application` / `Scripting.FileSystemObject` and `openpyxl` respectively — no shell, no HTTP, no registry, no downloads. Neither is wired into CI.
- The fixtures are ordinary packages: no `vbaProject.bin`, no OLE objects, no embedded media, no external links.
- Provenance verified rather than taken on trust: `<Application>Microsoft Excel</Application>` / `AppVersion 16.0300` on the Excel files, `Openpyxl 3.1.5` on the others, and the `.xls` is a real OLE2 CFB.
- **Privacy verified**: `dc:creator` and `cp:lastModifiedBy` are empty on every xlsx/xlsb, and the BIFF `SummaryInformation` carries no author string. Good call asserting `hasAuthor: false` in the suite itself — a note in a markdown file would not have survived the next regeneration.

## What the corpus is worth

openpyxl reaches shapes Excel cannot be made to write — a formula with no cached result, an empty `<v/>`, `t="d"`, `date1904="1"`, inline strings with no `sharedStrings.xml` — and that is a better argument for the second producer than "more files".

Six reader defects out of code that was at 98.8% statement coverage, and two of them (#499, #501) are the kind that only turn up when you point the reader at files nobody authored for it. The note about the 538 instrument-exported workbooks is the most useful paragraph in the PR; the loop it describes — read real files, reproduce what breaks as a synthetic fixture — is the one worth keeping.

Each defect stays `it.fails` with its issue. I have not fixed any of them here, for the reason you gave.

Issue #503, opened 2026-08-13
readXlsx dies with a raw V8 "Cannot create a string longer than 0x1fffffe8" on worksheets over 512 MB


erannave:
Found by running hucre over a corpus of real instrument-exported workbooks (the exercise behind #495 / #500). Three files of ~600 hit it.

`readXlsx` on a large workbook fails with a raw V8 error:

```
Error: Cannot create a string longer than 0x1fffffe8 characters
```

Not a `ParseError`. Not wrapped. Nothing in it says "spreadsheet", names the part, or suggests what to do — and it arrives after ~9 seconds of work.

### Cause

`src/xlsx/reader.ts:70` decodes an entire ZIP entry into one JavaScript string:

```ts
return new TextDecoder("utf-8").decode(data)
```

V8's maximum string length is `0x1fffffe8` = 536,870,888 characters (~512 MB). A worksheet part above that cannot be turned into a string at all, so the reader dies before any parsing begins.

The three files here are 56–99 MB compressed. The one I measured has:

```
607505144  xl/worksheets/sheet1.xml     <- 607 MB uncompressed, over the cap
  7155882  xl/sharedStrings.xml
```

1,048,557 rows — essentially Excel's row limit — which is ordinary for instrument data logged at high sample rates.

### There is already a working path

`streamXlsxRows` reads the same file without complaint:

| | `readXlsx` | `streamXlsxRows` |
| --- | --- | --- |
| result | throws at 9.2 s | 1,048,557 rows in 29.7 s |
| peak RSS | — | 944 MB, flat |

So this is not a "hucre cannot read big files" problem. It is that the buffered reader has a hard ceiling it does not know about, and reports hitting it in the least useful way available.

### Suggested fix

Cheap and self-contained: catch the failure where the string is built and rethrow as a `ParseError` that names the part, the size, and the way out. Something like

> Worksheet part `xl/worksheets/sheet1.xml` is 607,505,144 bytes, over the 536,870,888-character maximum string length this runtime supports. Use `streamXlsxRows` to read this workbook.

The `maxTotalCells` error is the model to copy — it names the bound, the measurement, and three options. Compare with what this one currently produces.

Worth checking `readXlsb` and `readOds` for the same shape, and `src/zip/reader.ts:280` decodes file names the same way (harmless — names are short — but it is the same call).

### Not fixture-able

[...]

Issue #501, opened 2026-08-13
A sparse sheet (82k values over a 305M-slot bounding box) cannot be read: the dense rows model has no way out


erannave:
Found by running hucre over a corpus of real instrument-exported workbooks (the exercise behind #495 / #500). Distinct from #492 — see below.

A workbook Excel opens without complaint cannot be read by hucre at all:

```
ParseError: Sheet "…" spans 19959 rows x 15312 columns (305612208 cells), over the 20000000 limit.
Use the `range` or `maxRows` read option to bound the area read, or raise `maxTotalCells` if the
sheet really is this large.
```

### The sheet is sparse, not large

Measured from the worksheet XML:

| | |
| --- | --- |
| `<dimension ref>` (Excel's own) | `A1:VPX19959` — 15,312 cols × 19,959 rows |
| `<c>` elements | 81,846 |
| of which carry a `<v>` | **76,277** |
| self-closing / style-only | 5,569 |
| distinct columns actually used | **507** |
| right-most column carrying a *value* | 15,311 |

So the data is real and genuinely wide — this is not #492's style-only inflation, where the padding comes from value-less cells. It is a legitimately sparse sheet: **82k cells scattered over a 305,612,208-slot bounding box, a fill factor of 0.03%.**

`src/xlsx/worksheet.ts:1148` materializes `rows` as a dense rectangle, so reading it needs ~305M array slots (>2 GB for the backing arrays alone) to hold 82k values. `MAX_TOTAL_CELLS` stops that, correctly — the process would otherwise OOM.

### The suggested workarounds don't apply

The error names three, and none is usable here:

- **`raise maxTotalCells`** — this is the one the message points at when "the sheet really is this large". The sheet is not large, it is sparse; raising the bound trades a clean error for a multi-gigabyte allocation to hold 82k values.
- **`range`** — needs the caller to already know where the data is. The used columns are scattered across 507 of 15,312; no single rectangle covers them without covering nearly everything.
- **`maxRows`** — bounds rows, and the problem is columns.

So in practice the file is simply unreadable, with no option that makes it readable.

### Why this is worth separating from #492

#492 is about *value-less* cells inflating the grid, and its fix — don't let a style-only cell bump `maxCol` under `readStyles: false` — would not help here, because the far-right cells carry values. The shared root cause is that `Sheet.rows` is a dense rectangle; #492 is the cheap half of it and this is the half where the model itself is the limit.

[...]

Issue #499, opened 2026-08-13
A workbook containing a chart sheet cannot be read at all: "missing worksheet file for sheet"


erannave:
Found by running hucre over a corpus of real production workbooks (the follow-up to #464 the corpus PR #495 asks for). **4 of the first 53 files failed on this**, ~7.5%.

A workbook that contains a **chart sheet** — a chart on its own tab, not an embedded chart object — cannot be opened by hucre at all:

```
ParseError: Invalid XLSX: missing worksheet file for sheet "<chart sheet name>"
```

The whole workbook is refused. Not "the chart is dropped": every ordinary worksheet in the file is unreachable too, and there is no option to skip past it.

### Cause

`src/xlsx/reader.ts:362-368` builds the rId → part map from *only* worksheet relationships:

```ts
for (const rel of workbookRels) {
  if (matchesRelType(rel.type, "worksheet")) {
    sheetRelMap.set(rel.id, resolvePath(workbookDir, rel.target))
  }
}
```

`xl/workbook.xml`'s `<sheets>` element lists **every** sheet in the book, whatever its kind — ECMA-376 `CT_Sheet` covers worksheets, chart sheets, dialog sheets and macro sheets alike, and the *relationship type* is what distinguishes them. A chart sheet's relationship is

```
Type=".../officeDocument/2006/relationships/chartsheet"  Target="chartsheets/sheet1.xml"
```

so it never enters `sheetRelMap`, `sheetRelMap.get(info.rId)` returns undefined, and line 380 throws. (Excel 4 macro sheets happen to survive — Excel gives them a `worksheet` relationship pointing at `xl/worksheets/` — so chart sheets and dialog sheets are the cases that bite.)

`src/xlsx/stream-reader.ts:916` has the same shape and presumably the same behaviour.

### Reproduction

Minimal, 12.6 KB, written by Excel: one data worksheet + one chart sheet, produced with

```vbs
Set ch = wb.Charts.Add        ' a chart SHEET, not a ChartObject
ch.ChartType = 4
ch.SetSourceData ws.Range("A1:B4")
```

`read()` on it gives the error above. I can add this as `test/fixtures/excel-chartsheet.xlsx` in a follow-up to #495 if you want it in the corpus.

### Why nothing caught it

The chart tests in `test/` build their `<sheets>` and rels by hand and only ever emit `worksheet` relationships, so the reader has never been shown a workbook whose sheet list contains something else. This is exactly the failure mode #464 describes.

[...]

PR #511, opened 2026-08-13
test: cover the features real workbooks actually use

erannave:
Rebuilt on current `main` (`3159288`), so this is a single commit with no stacking and no conflict.

Test-only. No reader or writer behaviour changes, and no new `it.fails`.

## The finding is a negative one

After #504–#507 landed, I pointed hucre at a second real corpus: **270 production business workbooks** — invoices, offers, planning sheets — a different population from the instrument data that produced #499, #501 and #503.

**All 270 read successfully.** No failures, no warnings, no `NaN`, no crashes. The chart-sheet fix is holding on files it was never tested against: 13 of those workbooks carry charts and every one opens. The only anomaly in the whole set was six more `.xls` files with #494's `undefined` rows, two carrying 79 holes each — that evidence went onto the issue, and #509 has since fixed both halves of it.

The files are private. They were read locally, nothing committed, nothing uploaded, and the harness reported structure only — never a cell value, a sheet name, a formula or a file name. What is in this PR is one 11 KB workbook Excel wrote from `make-fixtures.vbs`.

## What it exposed instead

Not a defect — a mismatch between what the corpus tests and what real files contain. Ranked by how often those 270 files use a feature:

| feature | share of files | had an Excel-written fixture? |
| --- | --- | --- |
| merges | 50% | yes |
| named ranges | 43% | **no** |
| autofilter | 30% | **no** |
| images | 28% | **no** |
| cell comments | 18% | **no** |
| data validation | 16% | **no** |
| sheet protection | 14% | **no** |
| conditional formatting | 10% | yes |

hucre reads every one of them correctly today. I checked comments specifically rather than assuming, sampling twelve files and finding 133 comments where expected — the one file that returned none has an *empty* `xl/comments1.xml`, one author and zero `<comment>` elements, so returning none is right.

The problem is only that a regression in any of those would have been caught by files hucre wrote itself and nothing else. That is the hole #464 exists to close, and the corpus had been aimed at the defects already known rather than at what people's files contain.

## What is here

`test/fixtures/excel-features.xlsx`, 11 KB:

- a workbook-scoped defined name (`TaxRate=0.19`) and a sheet-scoped one (`Amounts=Features!$B$2:$B$4`)
- an autofilter over `A1:C4`
- a list validation on `C2:C4`
- sheet protection, with `B2:B4` left unlocked so both states exist
- a classic (non-threaded) cell comment on `A2`

Images are left out on purpose: an embedded picture costs more bytes than the rest of the corpus put together, and the drawing part is already exercised by `excel-chartsheet.xlsx`.

The projection in `real-files.test.ts` grows to match — `autoFilter`, `validations`, `protection`, `comments` per sheet, `namedRanges` per workbook — so these are *asserted*, not merely present. Every existing golden gains the new keys as empty, which is itself an assertion: those fixtures genuinely have none of these features, and the suite now says so rather than leaving it unstated.

## On the golden

Hand-written from what the `.vbs` authored, as with the rest. Two values were wrong when I wrote them and were corrected **against the file, not against hucre**: Excel's `ws.Protect` sets `objects` and `scenarios` in addition to `sheet`, which `xl/worksheets/sheet1.xml` confirms as `<sheetProtection sheet="1" objects="1" scenarios="1"/>`. Worth saying explicitly, because "the golden disagreed so I changed the golden" is the failure mode this whole approach exists to avoid — the check is the file, every time.

Also for the provenance story: `RemovePersonalInformation` blanks the comment author to the literal `Author`. A comment is one of the places a real name otherwise leaks into a fixture, and this is the first fixture here to have one.

## Checks

- `pnpm test` — 203 files, 10,268 passed, 1 expected fail (#501).
- 192 KB of binaries; 303 KB for the whole directory including golden models and `PROVENANCE.md`.
- `docProps/core.xml` empty `dc:creator` / `cp:lastModifiedBy`, same path as the rest.


productdevbook:
Merged, with one commit added on top.

`excel-features.xlsx` covers the five things the earlier fixtures did not — auto-filter, data validation, sheet protection, comments, named ranges — and it reads correctly on every one, which is the answer worth having either way. Same review as before: real Excel 16.0300, `dc:creator` empty, no macros or OLE objects, nothing sensitive touched.

The commit I added retires the last `it.fails`. #501 landed while this was open, so the sparse block is now four passing assertions instead: `sparse: true` reads it, streaming reads it, the default still refuses it — correctly, the grid genuinely cannot be built — and the error names both ways out.

**All six markers are now gone.** #493, #494, #496, #497, #499 and #501 are fixed and merged, each with its own failing test first, exactly the sequence you set up. PROVENANCE.md's narrative says so rather than still describing them as pending.

Worth recording what the corpus actually bought, since it is the clearest answer #464 has had:

| | found by |
| --- | --- |
| #493 line endings | one authored cell, two containers, two strings |
| #494 ragged rows | the same, three containers |
| #496 `t="d"` | the openpyxl producer |
| #497 cached results | starting to assert them |
| #499 chart sheets | pointing the reader at real files |
| #501 sparse sheets | the same |
| #503 the 512 MB ceiling | the same |

Seven defects, none of which the 10,000-test suite could see, in code at 98.8% statement coverage. Four of them needed a *second producer* rather than more Excel files, which is the part I would not have predicted.

PR #519, opened 2026-08-13
fix(read): the string-ceiling guard never fired on Node — closes #516

### LINKED ISSUE #516: The #503 fix never fires on Node: TextDecoder throws Error/ERR_STRING_TOO_LONG, not RangeError
#514 fixed #503 by converting the string-ceiling failure into a typed `ParseError`. On Node the conversion never happens: the guard tests for the wrong error type, so the raw V8 error still reaches the caller exactly as before.

### The guard

`src/_decode.ts`:

```ts
export function decodePart(data: Uint8Array, path: string): string {
  try {
    return new TextDecoder("utf-8").decode(data)
  } catch (error) {
    // A RangeError here is the string-length ceiling. Anything else is
    // not ours to reinterpret.
    if (error instanceof RangeError) throw tooLargeToDecode(path, data.length)
    throw error
  }
}
```

Node's `TextDecoder.decode` does not throw a `RangeError`. It throws a plain `Error` carrying `code: "ERR_STRING_TOO_LONG"`:

```js
const big = new Uint8Array(0x1fffffe8 + 1024).fill(0x41)
try { new TextDecoder("utf-8").decode(big) } catch (e) {
  e.constructor.name        // "Error"
  e instanceof RangeError   // false
  e.code                    // "ERR_STRING_TOO_LONG"
  e.message                 // "Cannot create a string longer than 0x1fffffe8 characters"
}
```

(Node 24.18.0. The `RangeError` assumption looks like it came from V8's own `RangeError: Invalid string length`, which is what plain string concatenation throws — but `TextDecoder` goes through Node's internal encoding layer, which wraps it as `ERR_STRING_TOO_LONG`.)

So `error instanceof RangeError` is false, the `throw error` branch runs, and the caller gets the untyped error the issue was about.

### Measured, not inferred

Re-running the corpus that produced #503 against `b723e80`, every file above the ceiling still fails identically:

```
10x Error: Cannot create a string longer than 0x1fffffe8 characters
```

Stack from one of them (a 162 MB workbook whose `xl/worksheets/sheet2.xml` is 589,208,551 bytes):

```
Error: Cannot create a string longer than 0x1fffffe8 characters
    at TextDecoder.decode (node:internal/encoding:494:28)
    at decodePart (.../_decode.ts)
    at decodeUtf8 (.../xlsx/reader.ts)
    at readXlsx (.../xlsx/reader.ts)
```

The frame is *inside* `decodePart` — it is being called, it just declines to convert.

### Why no test caught it

The comment above `tooLargeToDecode` says it:

> Separated from the decode so it can be tested: reproducing the condition needs a part over 512 MB, which is larger than this repository, so the message and the decision are what a test can reach.

The message factory is tested; the `catch` that decides whether to call it is not. That is the same closed loop #464 is about, reappearing inside the fix for a bug that was only found by breaking it — the test asserts what the code would do *if* the branch were taken.

It is reachable, though, and cheaply: the condition needs a large **buffer**, not a large **file**. A `Uint8Array` of `MAX_STRING_LENGTH + 1` is a ~537 MB `ArrayBuffer` allocation outside the JS heap, about a second, no fixture:

```ts
it("turns the string ceiling into a ParseError", () => {
  expect(() => decodePart(new Uint8Array(MAX_STRING_LENGTH + 1), "xl/worksheets/sheet1.xml"))
    .toThrow(ParseError)
})
```

That fails today.

### Suggested fix

Keep the `catch` — the reasoning for not comparing lengths up front is right, engines differ and a guess would refuse files JavaScriptCore could handle. Just widen what counts:

```ts
} catch (error) {
  // Node wraps V8's ceiling as ERR_STRING_TOO_LONG on a plain Error;
  // V8 raises RangeError directly; other engines differ again. Treat any
  // failure on a part that could not fit as the ceiling, and let a
  // genuinely malformed short part through untouched.
  const tooLong =
    error instanceof RangeError ||
    (error as { code?: string })?.code === "ERR_STRING_TOO_LONG" ||
    data.length > MAX_STRING_LENGTH
  if (tooLong) throw tooLargeToDecode(path, data.length)
  throw error
}
```

The `data.length > MAX_STRING_LENGTH` arm only ever runs when the decode has already failed, so it cannot refuse a file a roomier engine would have read — it just stops the classification depending on which wrapper a runtime chose.

Happy to send this as a PR with the test above if you want it.

productdevbook:
Closes #516. You are right on both counts, and the second one is the more useful correction.

## The guard tested the wrong type

Verified here rather than taken on trust:

```
ctor       : Error
RangeError : false
code       : ERR_STRING_TOO_LONG
message    : Cannot create a string longer than 0x1fffffe8 characters
```

`TextDecoder.decode` goes through Node's internal encoding layer, which wraps V8's failure rather than passing it through. My `RangeError` assumption came from plain string concatenation, which *does* throw one — so the fix for #503 shipped doing nothing on the runtime it was written for.

Three signals now, since no single one holds everywhere:

- Node's `code === "ERR_STRING_TOO_LONG"`
- other engines' `RangeError`
- **the byte length**, as the backstop that does not depend on guessing at engine internals: a string cannot exceed the ceiling in characters unless the buffer exceeds it in bytes, because UTF-8 uses at least one byte per character. So a failure on a buffer that large *is* this failure, whatever the engine chose to call it.

I kept the `catch` rather than switching to an up-front comparison, for the reason you agreed with — but the length now serves as evidence inside the catch instead of as a gate before it, which gets the engine-independence without the false positives.

## The claim about testing it was wrong

This is the part worth naming. #514 said:

> reproducing the condition needs a part over 512 MB, which is larger than this repository

and tested only the message factory. The condition needs a large **buffer**, not a large **file** — as you point out, one `new Uint8Array(MAX_STRING_LENGTH + 1)` reproduces it in about a second, outside the JS heap, with no fixture.

So the untested branch was the one with the bug in it, and the comment explaining why it could not be tested was doing the work of hiding that. Your framing is exact: it is the closed loop of #464 reappearing *inside* the fix for a bug that only breaking the loop had found.

The test is now there. It fails against `b723e80` and passes here, and it is the only case in the file that exercises the `catch` rather than the message it builds.

```
× turns the real ceiling into a ParseError
  AssertionError: expected error to be instance of ParseError
```

`pnpm lint`, `pnpm typecheck`, 10317 tests green; the new case adds ~1.2s.

### DIFF
--- a/src/_decode.ts
+++ b/src/_decode.ts
@@ -49,6 +49,31 @@ export function tooLargeToDecode(path: string, byteLength: number): ParseError {
 }
 
 /**
+ * Is this the string-length ceiling, and not some other failure?
+ *
+ * Three signals, because no one of them holds everywhere:
+ *
+ * - **Node throws a plain `Error`** carrying `code: "ERR_STRING_TOO_LONG"`.
+ *   `TextDecoder.decode` goes through Node's internal encoding layer,
+ *   which wraps V8's failure rather than passing it through. The first
+ *   version of this checked `instanceof RangeError` — which is what a
+ *   plain string concatenation throws — so on Node the guard never fired
+ *   and the raw error reached the caller exactly as before. See #516.
+ * - **Other engines throw a `RangeError`**, so that stays.
+ * - **The byte length is a necessary condition either way.** A string
+ *   cannot exceed {@link MAX_STRING_LENGTH} characters unless the buffer
+ *   exceeds it in bytes, since UTF-8 uses at least one byte per
+ *   character. So a failure on a buffer that large is this failure,
+ *   whatever the engine chose to call it — which is the backstop that
+ *   does not depend on guessing at engine internals.
+ */
+function isStringLengthCeiling(error: unknown, byteLength: number): boolean {
+  if (error instanceof RangeError) return true
+  if ((error as { code?: unknown } | null)?.code === "ERR_STRING_TOO_LONG") return true
+  return byteLength > MAX_STRING_LENGTH
+}
+
+/**
  * Decode a package part as UTF-8, or say why it cannot be.
  *
  * `path` is only used for the message, so a caller that has not resolved
@@ -58,9 +83,8 @@ export function decodePart(data: Uint8Array, path: string): string {
   try {
     return new TextDecoder("utf-8").decode(data)
   } catch (error) {
-    // A RangeError here is the string-length ceiling. Anything else is
-    // not ours to reinterpret.
-    if (error instanceof RangeError) throw tooLargeToDecode(path, data.length)
+    // Anything that is not the ceiling is not ours to reinterpret.
+    if (isStringLengthCeiling(error, data.length)) throw tooLargeToDecode(path, data.length)
     throw error
   }
 }
--- a/test/huge-part-error.test.ts
+++ b/test/huge-part-error.test.ts
@@ -17,11 +17,18 @@ import { ParseError, HucreError } from "../src/errors"
 // those files, in ~30s at a flat 944 MB. The buffered reader had a hard
 // ceiling it did not know about.
 //
-// The trigger is not reproducible here and that is worth stating rather
-// than working around: it needs a part larger than this repository.
-// Faking the decode failure would test the wrapper rather than the
-// condition, so what is tested is what has logic in it — the decision
-// and the message.
+// The first version of this file said the trigger was not reproducible —
+// that it needed a part larger than the repository — and tested only the
+// message factory. That was wrong, and the gap it left was a real bug:
+// the guard tested `instanceof RangeError`, Node throws a plain `Error`
+// with `code: "ERR_STRING_TOO_LONG"`, so on Node the conversion never
+// happened and the raw error reached the caller exactly as before (#516).
+//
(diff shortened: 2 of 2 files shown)


productdevbook:
Closing in favour of #518, which is merged — you found the bug and reported it, and your fix is at least as good as mine in every respect. We converged on the same three signals independently, which is reassuring about the shape of the fix.

Two things yours does better and I would have wanted anyway:

Your comment explains *why* the length backstop is safe rather than just that it is — "only consulted once the decode has already failed, so it cannot refuse a part a roomier engine would have decoded". Mine asserted the necessary-condition reasoning without spelling out that it is evidence inside the catch rather than a gate before it, which is the part someone reading it later actually needs.

And your test comment notes that zero bytes decode one-to-one, so the buffer lands exactly one character past the ceiling. Mine used the same size without saying why it is exact.

The correction I most want on the record is the one you made about my claim, not about my code: #514 said reproducing the condition needed a part larger than the repository, and tested only the message factory. That was wrong — a large *buffer*, not a large *file* — and the branch I declared untestable was the one with the bug in it. The paragraph explaining why the test could not exist was doing the work of hiding that. Worth remembering next time I am tempted to write one.

Issue #527, opened 2026-08-13
`sparse: true` is offered as the escape from `maxTotalCells` but throws `RangeError: Map maximum size exceeded` above 2^24 cells


erannave:
## What happens

`readXlsx(input, { sparse: true })` throws a raw

```
RangeError: Map maximum size exceeded
```

on any sheet with more than 2^24 = 16,777,216 filled cells. `Sheet.cells` is a
`Map<string, Cell>` and V8 caps a `Map` at that many entries, so the sparse path
has a hard ceiling of its own — one that is not a `HucreError`, names no sheet,
and says nothing about spreadsheets.

## Why it matters more than the raw error suggests

`sparse: true` is one of the two escapes the `maxTotalCells` error offers (#501,
added in #513):

```
  - streamXlsxRows(input) reads it a row at a time, whatever the box.
  - readXlsx(input, { sparse: true }) returns the cells and no grid.
```

For a sheet that is over `maxTotalCells` because its **box** is large and mostly
empty — #501's file, 82,000 values in a 305M-slot box, 0.03% filled — that advice
is right, and `sparse` is the answer it was built for.

For a sheet that is over `maxTotalCells` because it is genuinely **dense and
large**, the advice is a dead end: the cell count that blew the bounding-box
limit is the same cell count that blows the `Map` limit. The message does not
distinguish the two cases, so it recommends `sparse` in exactly the situation
where `sparse` cannot work.

## Measured

Reading ~600 instrument-exported workbooks, four land here. Structure only — no
content from these files is reproduced.

| bounding box | total slots | filled | filled cells | `sparse: true` |
| --- | ---: | ---: | ---: | --- |
| 579,151 × 35 | 20,270,285 | 96.25% | ~19.5M | `RangeError` |
| 637,888 × 35 | 22,326,080 | 88.14% | ~19.7M | `RangeError` |
| 658,025 × 35 | 23,030,875 | 85.70% | ~19.7M | `RangeError` |
| 916,449 × 33 | 30,242,817 | 93.95% | ~28.4M | `RangeError` |

All four are well past 16,777,216 filled cells, so all four fail.

**`streamXlsxRows` — the other escape the message names — works on all four**, at
579,151 / 637,888 / 658,025 / 916,449 rows in 40 / 53 / 46 / 53 s at 510 / 521 /
527 MB / 2.1 GB peak RSS. Its row counts match the bounding boxes the
`maxTotalCells` error reported, so the data is all there; only the `Workbook`
[...]

PR #536, opened 2026-08-13
fix(read): a worksheet part over the string ceiling now reads — closes #503

### LINKED ISSUE #503: readXlsx dies with a raw V8 "Cannot create a string longer than 0x1fffffe8" on worksheets over 512 MB

erannave:
## Scope

**Fixed:** `readXlsx` reads a workbook whose *worksheet* part is over V8's
0x1fffffe8-character string ceiling. Ten of the fourteen such workbooks in a
~600-file corpus now import; the other four clear the ceiling and land on
`maxTotalCells`, with numbers below.

**Not fixed, deliberately:** `readXlsb` and `readOds` still have the ceiling, and
so does every non-worksheet part of an XLSX — `sharedStrings.xml`, `styles.xml`, a
drawing. Those still throw the #514 `ParseError`. I scanned all ~600 central
directories: no non-worksheet part in the corpus is anywhere near the ceiling, so
this is untriggered here rather than tested-and-working. A workbook with a 512 MB
shared-string table would still fail.

#514 and #518 made the failure legible. Neither made a file load. This is the
capability fix.

Rebased onto `main` after #528-#531 landed. #530 is the fix for #527, which came
out of measuring this branch — so the four workbooks below that land on
`maxTotalCells` now get the error that describes their case, and the `Map`
ceiling they hit is `main`'s bound rather than something this PR has to explain.

## The change

`parseWorksheet`'s handlers were already SAX handlers — only their *driver*
needed the string.

- `src/xlsx/worksheet.ts` — the handlers and the finalisation are split into
  `worksheetParser(name, ctx)`, returning `{ handlers, finish }`. `parseWorksheet`
  drives them with `parseSax` over a string; the new `parseWorksheetStream` drives
  the *same* object with `parseSaxStream` over a `ReadableStream`. The diff is
  large but almost entirely re-indentation — `git diff -w` is small.
- `src/xlsx/reader.ts:71` — `readWorksheet()` chooses the driver.
- `src/zip/reader.ts` — `declaredSize(path)` exposes the central directory's
  uncompressed size (0 and the ZIP64 sentinel both report `undefined`, since
  neither means "small").
- `src/_decode.ts` — `isPartTooLargeToDecode()`, identity-based.
- `src/zip/reader.ts` — `extractEntryStream` now emits STORE and
  fallback-inflate entries in 1 MiB chunks (see "incidental fix").

`parseWorksheetStream` is intentionally **not** exported from `src/index.ts`. It is
an internal seam; `streamXlsxRows` remains the public streaming API.

## Decision 1 — automatic, no option

Both drivers run the same handler set and the same finalisation, so they cannot
disagree about a field of the model; a second model builder would be a second set
of answers for every field, which is the divergence `CONTRIBUTING.md` calls the
registers. With that established, automatic is safe, and the ceiling stops being
something a caller has to know about — which is the point.

Which driver runs is decided from the ZIP's declared uncompressed size, before
anything is decompressed. When the ZIP declares no size (a streaming producer
writes zero), the buffered read runs and the ceiling error it raises is the signal
to retry as a stream — the slow path, since it pays for the decompression twice.
That error is recognised by object identity, not by matching its message; matching
the text would be exactly the mistake #516 was.

The size gate compares *bytes* against a *character* ceiling, so a multibyte part
can be sent to the streaming path that the buffered one would have managed. That
costs nothing — same handlers, same `Sheet`.

**Two divergences closed rather than documented.** A multi-agent correctness
review of this branch found both; each has a test that fails without its fix.

- `parseSaxStream` *dropped* an unfinished construct at exactly the six places
  `parseSax` throws `XmlError`, so a worksheet truncated by a crashed producer
  would have read as a short `Sheet` with no error — silent data loss on
  precisely the files this PR is for. It now takes a `strict` option.
  `parseWorksheetStream` asks for it; the row streamers do not, so
  `streamXlsxRows` keeps the lenient contract
  `test/coverage-xml-parser.test.ts` pins. Making it the default would have
  changed that public behaviour for everyone, which is not this PR's to change.
- A CRLF split across a text-flush boundary normalised to *two* newlines
  (`normalizeEol` runs per emitted piece, and `safeTextSplit` guarded entities
  and surrogate pairs but not CR/LF). Only reachable in a run over 256 KiB with
  the chunk boundary on the CRLF, so no ordinary sheet shows it.

**On CRC-32.** The streaming ZIP path has no whole entry to check, so a
worksheet read that way is not CRC-32 verified — the same trade
`streamXlsxRows` has always made. It is bounded on the way in: a declared size
the compressed body could not have produced is not believed (DEFLATE tops out at
1032:1, a stored entry does not expand), so a 40 KB entry cannot claim 600 MB,
take the streaming route and skip the checksum. `docs/PARITY.md` says all of
this.

## Decision 2 — measured: ten read, four land on the cell bound

One process per workbook, `--max-old-space-size=8192`, timeout 15 min, concurrency
1. Files are labelled `F1`–`F14`; nothing from them is reproduced here. "before" is
`origin/main` at 05a7358 — time and RSS to *reach the error*, i.e. work spent to
fail.

| file | part bytes | before | after | rows x cols | before RSS | after RSS |
| --- | ---: | --- | --- | ---: | ---: | ---: |
| F1 | 588,490,362 | ParseError 6.1s | reads 32.7s | 546,462 x 62 | 2,890 MB | 1,586 MB |
| F2 | 588,490,362 | ParseError 5.7s | reads 30.2s | 546,462 x 62 | 2,891 MB | 1,585 MB |
| F3 | 588,939,986 | ParseError 4.1s | reads 33.9s | 800,001 x 15 | 2,011 MB | 1,808 MB |
| F4 | 588,951,686 | ParseError 4.2s | reads 24.6s | 800,001 x 15 | 1,981 MB | 1,227 MB |
| F5 | 589,208,551 | ParseError 4.1s | reads 38.2s | 800,001 x 15 | 2,015 MB | 2,036 MB |
| F6 | 589,255,109 | ParseError 3.9s | reads 36.7s | 800,001 x 15 | 2,013 MB | 2,029 MB |
| F7 | 607,505,144 | ParseError 4.0s | reads 25.5s | 1,048,557 x 13 | 2,266 MB | 749 MB |
| F8 | 787,192,563 | ParseError 6.0s | reads 115.1s | 501,982 x 35 | 2,644 MB | 2,040 MB |
| F9 | 860,154,420 | ParseError 7.4s | reads 75.0s | 458,225 x 41 | 3,648 MB | 1,239 MB |
| F10 | 880,986,535 | ParseError 6.1s | reads 76.2s | 566,155 x 35 | 2,944 MB | 2,286 MB |
| F11 | 886,914,347 | ParseError 6.4s | `maxTotalCells` 52.9s | — | 2,969 MB | 789 MB |
| F12 | 895,198,503 | ParseError 6.3s | `maxTotalCells` 90.9s | — | 3,008 MB | 811 MB |
| F13 | 900,010,877 | ParseError 6.0s | `maxTotalCells` 79.6s | — | 3,024 MB | 814 MB |
| F14 | 1,131,823,668 | ParseError 12.1s | `maxTotalCells` 131.2s | — | 5,323 MB | 2,630 MB |

Peak RSS is **lower after the change in twelve of the fourteen** — often much
lower (F7 2,266 → 749 MB, F9 3,648 → 1,239 MB, F14 5,323 → 2,630 MB) — and that
is while doing the whole parse the "before" never started. F5 and F6 are the
exceptions and are a wash: 2,015 → 2,036 MB and 2,013 → 2,029 MB, under 1%.
Buffering a 589 MB–1.13 GB part into a `Uint8Array` and then asking
`TextDecoder` for a string of it generally cost more than parsing the part a
chunk at a time costs.

"after" is the final branch; "before" is `origin/main` at 05a7358, where the
ceiling error is unchanged by #528–#531. Wall times are as measured, one process
each, on files served over OneDrive under WSL2, and **cold-read variance
dominates them**: F1 and F2 are byte-identical and differ run to run, and F8
measured 38 s and 115 s on two runs of the same build. Read the times as an order
of magnitude, not a benchmark; the RSS figures are stable.

The four that land on the cell bound, from the error itself:

| file | bounding box | filled |
| --- | --- | ---: |
| F11 | 579,151 × 35 = 20,270,285 | 96.25% |
| F12 | 637,888 × 35 = 22,326,080 | 88.14% |
| F13 | 658,025 × 35 = 23,030,875 | 85.70% |
| F14 | 916,449 × 33 = 30,242,817 | 93.95% |

Against `MAX_TOTAL_CELLS` = 20,000,000. Since #530 the error tells these four
apart from the sparse case and stops offering them `sparse`, e.g. F14:

```
  - streamXlsxRows(input) reads it a row at a time, whatever the box.
  - `sparse: true` cannot help here: 28412273 filled cells is past the 16777216 a Map can hold.
```

which is the right answer, and it is the one that works:

- **`streamXlsxRows` works** on all four — 579,151 / 637,888 / 658,025 / 916,449
  rows in 40 / 53 / 46 / 53 s at 510 / 521 / 527 MB / 2.1 GB peak RSS. Those row
  counts match the bounding boxes the cell-limit error reports exactly, which is
  a useful cross-check that the streaming worksheet parse sees the same data.
- **`sparse: true` does not**, which is what #527 was about and what #530 fixed:
  all four have 19–28 M filled cells against a `Map` cap of 16,777,216.

So all fourteen now either import or fail with an error that names an escape that
actually works. That is the whole of what this PR set out to change.

## Incidental fix, needed for this to work at all

`extractEntryStream` emitted a STORE entry (and the fallback inflate's output) as
a *single* `enqueue`. `parseSaxStream` does `buf += decoder.decode(chunk)`, so a
589 MB stored worksheet would have rebuilt the very 512 MB string the streaming
path exists to avoid — the new reader would have failed on the files it was added
for. Both paths now emit 1 MiB chunks; nothing is copied, each chunk is a
`subarray` view. `DecompressionStream` already chunked, so the deflate path
(every file above) is unchanged.

The review caught that `ZipStreamReader.entryStream` had the identical
one-chunk pattern in its no-`DecompressionStream` fallback — the branch the
Deno/browser/Workers claims rest on — so `streamXlsxRows` would still have died
on these files there. `chunkedStream` and its chunk size moved to
`src/zip/byte-limit.ts` and both readers now use it, rather than one being fixed
and its twin left behind.

## Bundle size

`readXlsx` grows **+2.7 KB minified / +1.0 KB gzipped (+2.8%)**, because it now
contains the streaming parser and the streaming ZIP path — `parseWorksheetStream`
reaches `parseSaxStream`, and the size gate reaches `extractStream`. Both were
already in the package; they were not in this entry point's graph.

`readOds` grows +0.3 KB / +0.2 KB (+2.2%) without gaining a feature: the CRLF
guard and the truncation contract live in `xml/parser.ts` and the chunked
emission in `zip/byte-limit.ts`, both shared. `csv` and `writeXlsx` are unchanged;
`everything` is +0.4%.

Re-baselined in its own commit with `node scripts/size.mjs --update`, as the
budget check asks. That command rewrites every budget to measurement + 5%, so the
scenarios that did not move are re-stated at the same headroom rather than left
alone — only `readXlsx` and `readOds` actually needed it.

## Tests

`test/huge-worksheet.test.ts`, 14 cases, ~3 s, fully in memory — no fixture, no
disk read, so no `tsconfig.cli.json` registration.

- **Fails before, passes after:** a workbook whose worksheet part is
  `MAX_STRING_LENGTH + 1024` bytes — two real rows with ignorable whitespace
  between them, in a hand-built STORE-only ZIP so nothing is spent deflating half
  a gigabyte. Re-verified against the final branch with the fix disabled: both
  cases throw the #514 `ParseError`. `test/huge-part-error.test.ts` set the
  precedent that a large *buffer* is affordable where a large *fixture* is not.
- **The undeclared-size path** has its own case: the ZIP writes zero for the
  uncompressed size, so the buffered read runs, hits the ceiling, and the retry
  is what has to work. The two cases **share one archive**, patching the declared
  size in place — two live 537 MB buffers made unrelated tests in other files
  miss their time budgets, which is a flaky suite for no gain. The file went from
  27 s to 3 s.
- **Divergence guard:** a worksheet exercising every handler that accumulates
  text across calls, parsed both ways and compared with `toEqual` — whole, and
  again with **every byte its own chunk**, plus the same in `sparse` mode.
- **Failure parity:** five documents ending mid-construct (opening tag, closing
  tag, comment, CDATA, PI), each asserted to throw `XmlError` on *both* drivers.
- **The CRLF flush cut:** a run over `SAX_TEXT_FLUSH_CHARS` with the chunk
  boundary placed exactly between the CR and the LF. Fails without the guard.
- **A lying declared size** is not believed, and the entry still reads by the
  buffered path it was trying to escape.

## Checks

- `corepack pnpm test` — green, twice in a row. 217 files, 10,442 tests.
- `corepack pnpm coverage` — 98.68 / 96.63 / 98.8 / 99.1 against 98.5 / 96 / 98 / 99.
- `test/real-files.test.ts` — passes. It has **zero** `it.fails` markers on current
  `main` (#501's was retired by #513); zero before this change, zero after.
- `docs/PARITY.md` updated in this PR: what reads and what still does not, the
  CRC-32 trade and its bound, and the truncation contract.

What I did not check: any runtime other than Node 24, and any workbook outside
this corpus. The no-`DecompressionStream` fallback fix is reasoned, not measured
— I have no runtime here that lacks `deflate-raw`.

Nothing from the corpus is reproduced anywhere in this PR — the harness reports
counts, byte sizes and part names only, and quoted literals in error messages are
redacted.

### DIFF
--- a/docs/PARITY.md
+++ b/docs/PARITY.md
@@ -122,24 +122,56 @@ would otherwise show `1E-07` — but only when that form is the same
 number, which it was not for the smallest values (`Number.MIN_VALUE` used
 to come out as `0.0`).
 
-### The buffered readers have a size ceiling
+### The string ceiling, and where it still is
 
 One JavaScript string cannot hold an arbitrarily large part. V8's limit
-is 536,870,888 characters — about 512 MB — so a worksheet above it cannot
-be turned into a string at all, and `readXlsx`, `readXlsb` and `readOds`
-fail before any parsing begins. Instrument logs at Excel's row limit
-reach it: three files in a corpus of ~600 were 56–99 MB compressed and
-607 MB expanded. See #503.
+is 536,870,888 characters — about 512 MB — so a part above it cannot be
+turned into a string at all, however much memory the machine has.
+Instrument logs at Excel's row limit reach it: in a corpus of ~600, 14
+workbooks had a worksheet part between 588 MB and 1.13 GB. See #503.
 
-That used to surface as a raw `Error: Cannot create a string longer than
-0x1fffffe8 characters` — not a `ParseError`, naming no part, saying
-nothing about spreadsheets. It is now a `ParseError` that names the part,
-the size, the bound, and `streamXlsxRows`; and it says the workbook is
-not damaged, because everything else a reader throws means the file is
-wrong and this one means it is large.
+**`readXlsx` reads a worksheet part of any size.** When the part is over
+the ceiling it is parsed from the ZIP entry's stream instead of from a
+string, by the same SAX handlers the buffered parse uses — so the `Sheet`
+is the same `Sheet`, and no option selects between them. Which path runs
+is decided from the size the ZIP declares, before anything is
+decompressed; when the ZIP declares no size, the buffered read runs and
+the ceiling error it raises is the signal to retry as a stream.
 
-**`streamXlsxRows` has no such ceiling** — it reads those same files in
-about 30 seconds at a flat 944 MB, because it never holds the part whole.
+Two things this does not do:
+
+- **The other buffered readers still have the ceiling.** `readXlsb` and
+  `readOds` are unchanged, as is every non-worksheet part of an
+  XLSX — `sharedStrings.xml`, `styles.xml`, a drawing. Those still fail
+  with the `ParseError` from #514: it names the part, the size, the
+  bound, and `streamXlsxRows`, and it says the workbook is not damaged,
+  because everything else a reader throws means the file is wrong and
+  this one means it is large. (None of the ~600 has a non-worksheet part
+  anywhere near the ceiling; a workbook with a 512 MB shared-string
+  table would still fail.)
+- **It does not lift the cell bound.** A part over the ceiling is
+  usually also a sheet with a very large bounding box, so clearing the
+  ceiling often lands on `maxTotalCells` instead — of the 14, ten read
+  and four hit that limit. See the next section.
+
+Two trades worth knowing:
+
+- The streaming ZIP path has no whole entry to check, so a worksheet read
+  this way is **not CRC-32 verified**. That is the same trade
+  `streamXlsxRows` has always made. It is bounded on the way in — a
+  declared size the compressed body could not have produced (DEFLATE tops
+  out at 1032:1) is not believed, so a small entry cannot claim to be
+  enormous and skip the checksum that way.
+- A truncated part is still an error. The row streamers have no error
+  contract for one — they drop the unfinished construct and let the
+  caller notice the missing rows — but the worksheet reader asks the
+  streaming parser for `strict`, so a part that ends mid-tag throws the
+  `XmlError` the buffered parser throws rather than returning a short
+  `Sheet`.
+
+**`streamXlsxRows` still has no ceiling either** and remains the lower-cost
+answer when you only need to walk the rows: it never builds a `Workbook`,
+so it is bounded by the row you are on rather than by the sheet.
 
 ### A sparse sheet has a way out
 
--- a/scripts/size-budget.json
+++ b/scripts/size-budget.json
@@ -4,23 +4,23 @@
     "gzip": 4038
   },
   "readXlsx (hucre/xlsx)": {
-    "min": 133188,
-    "gzip": 36967
+    "min": 138926,
+    "gzip": 39085
   },
   "read+write xlsx (hucre/xlsx)": {
-    "min": 266018,
-    "gzip": 72846
+    "min": 271860,
+    "gzip": 75215
   },
   "everything (hucre)": {
-    "min": 489277,
-    "gzip": 138764
+    "min": 500939,
+    "gzip": 142754
   },
   "writeXlsx (hucre)": {
-    "min": 137695,
-    "gzip": 39254
+    "min": 137825,
+    "gzip": 39445
   },
   "readOds (hucre/ods)": {
-    "min": 26941,
-    "gzip": 9438
+    "min": 28240,
+    "gzip": 9953
   }
 }
--- a/src/_decode.ts
+++ b/src/_decode.ts
@@ -30,6 +30,29 @@ import { ParseError } from "./errors"
 export const MAX_STRING_LENGTH = 0x1fffffe8
 
 /**
+ * The error a part over the string ceiling produces.
+ *
+ * A caller that can do something about the ceiling — the worksheet
+ * reader, which parses the part as a stream instead (#503) — has to
+ * recognise this one error among all the `ParseError`s a decode can
+ * raise. Matching its message would be the same mistake #516 was: the
+ * text is for people, and a reworded sentence would silently turn the
+ * recovery off. A class cannot be reworded, and it is how every other
+ * distinguishable condition in this library is told apart — see
+ * `errors.ts`.
+ *
+ * Unexported, and a `ParseError` first: nothing outside needs to name it
+ * yet, and every existing `catch (e) { e instanceof ParseError }` keeps
+ * working.
+ */
+class PartTooLargeError extends ParseError {}
+
+/** Whether `error` is the "part is over the string ceiling" `ParseError`. */
+export function isPartTooLargeToDecode(error: unknown): boolean {
+  return error instanceof PartTooLargeError
+}
+
+/**
  * The message for a part too large to become a string.
  *
  * Separated from the decode so it can be tested: reproducing the
@@ -39,7 +62,7 @@ export const MAX_STRING_LENGTH = 0x1fffffe8
  * measurement, and the way out.
  */
 export function tooLargeToDecode(path: string, byteLength: number): ParseError {
-  return new ParseError(
+  return new PartTooLargeError(
     `Part "${path}" is ${byteLength.toLocaleString("en-US")} bytes, over the ` +
       `${MAX_STRING_LENGTH.toLocaleString("en-US")}-character maximum string length ` +
       `this runtime supports, so it cannot be read into memory whole.\n` +
--- a/src/xlsx/reader.ts
+++ b/src/xlsx/reader.ts
@@ -21,7 +21,7 @@ import type {
   TimelineCache,
   ChartAnchor,
 } from "../_types"
-import { decodePart } from "../_decode"
+import { decodePart, isPartTooLargeToDecode, MAX_STRING_LENGTH } from "../_decode"
 import { parsePersons, parseThreadedComments } from "./threaded-comments-reader"
 import { parseExternalLink } from "./external-link-reader"
 import { assembleCellImages, parseCellImages, REL_CELL_IMAGES } from "./cell-images-reader"
@@ -37,7 +37,8 @@ import { parseContentTypes } from "./content-types"
 import { parseRelationships } from "./relationships"
 import { parseSharedStrings } from "./shared-strings"
 import { parseStyles } from "./styles"
-import { parseWorksheet } from "./worksheet"
+import { parseWorksheet, parseWorksheetStream } from "./worksheet"
+import type { WorksheetContext } from "./worksheet"
 import { parseDynamicArrayCellMetadata } from "./metadata"
 import type { ParsedStyles } from "./styles"
 import type { SharedString } from "./shared-strings"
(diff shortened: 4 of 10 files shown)


productdevbook:
Merged. Thank you — this is the capability #503 was missing, and #514 and #518 only ever made legible.

I reviewed it rather than taking CI's word for it. What I checked, and what I found:

## The claim I most wanted to test

That the streaming driver and the buffered one really do build the same `Sheet` — "one implementation, two drivers" is the whole argument for the refactor, and it is the kind of claim that is true when written and false a year later.

So I ran my own comparison rather than relying on the tests in the PR: **every worksheet part of all 32 real fixtures** in `test/fixtures` — Excel 16.0, openpyxl, ExcelJS — parsed both ways, with and without `readStyles`, compared as serialised `Sheet`s with `Map`s and `Date`s normalised. Byte-identical throughout.

The nicest result was accidental. `excel-sparse.xlsx` is over the bounding-box limit, so both paths throw — and they throw the *same* error, same class and same message. Agreeing about failure is as much a parity claim as agreeing about success, and your `endOfInput` comment says you already thought about that.

## The rest

- **No read regression.** 60,000 x 12, five runs each, medians: main 1124 ms, this 1097 ms. The ranges overlap heavily; there is no difference to see.
- **The merge, not just the branch.** CI ran on your base; seven of my changes landed after it. Only `docs/PARITY.md` overlapped and in a different section. Merged locally first: 224 files, 10,516 tests, format and typecheck clean.
- **`carriesData`.** Your branch carries the `t="inlineStr"` case from #528, which lands in the middle of the region you refactored. Checked specifically, because a clean textual merge is not the same as a correct one.

## The cost, stated plainly for whoever reads this later

Bundle size, gzipped: `readXlsx` 36,967 → 39,085 (+5.7%), `readOds` 9,438 → 9,953 (+5.5%), everything 138,764 → 142,754 (+2.9%). ODS grows because it shares `xml/parser.ts` and `zip/reader.ts`. That is a real trade for a capability most callers will never reach, and the re-baselined `size-budget.json` is where it is recorded.

## Two things I want to single out

**The `PartTooLargeError` subclass.** You cite #516 for why not to match on the message. That is exactly the lesson, and it was mine to learn — I guarded that condition on `instanceof RangeError` when Node throws a plain `Error`, and the guard never fired. Keeping it unexported and a `ParseError` first means every existing `catch` keeps working.

**The CRLF split.** A chunk boundary between a CR and its LF turning one line ending into two, on the streaming driver only, and only in a text run long enough to be flushed. That is not a bug anyone finds by reading; well caught.

The `strict` split is the right call too — `streamXlsxRows` has never had an error contract for a truncated document, and quietly giving it one would have been a behaviour change hidden inside a capability change.

Left open deliberately: the CRC-32 point you raise. A part read as a stream is not checked for corruption, which is the trade `streamXlsxRows` already makes, and your comment says so where someone will find it.

PR #545, opened 2026-08-13
fix(ods): a LibreOffice row ends at its data, not column 16,384

onmax:
Refs #464.

A five-cell row written by LibreOffice came back **16,384 columns wide**. No error, no warning — 16,379 trailing nulls on every row:

```
readOds        Name | Qty | Date | Active | Total | null … × 16,379
streamOdsRows  Name | Qty | Date | Active | Total
```

## What LibreOffice writes

LibreOffice ends each populated row with one empty cell repeated to the sheet boundary:

```xml
<table:table-cell table:style-name="Default"
                  table:number-columns-repeated="16379"/>
```

`readOds` already trimmed trailing empty entries, but **any** `table:style-name` stopped the trim. That happened even when `readStyles` was false, and when the style name did not resolve — so the entry carried no value or metadata but still expanded into 16,379 cells. The streaming reader ignores that unused style name and exposed the disagreement.

The trim now keeps a style-only entry only when the caller requested styles **and** the style resolves. Values, formulas, hyperlinks, spans and styles the reader can actually return are unchanged.

## Why the corpus found it

#464 named two shapes still missing from the ODS corpus: LibreOffice's `table:number-columns-repeated` and its error cells. `libreoffice-basic.ods` is LibreOfficeDev 26.8's conversion of the existing, project-owned `excel-basic.xlsx`; it contains both, and found this defect on the first read.

The fixture pins its producer and repeated tail, checks the widths with `readStyles` both off and on, preserves LibreOffice's `#DIV/0!`, and joins the existing buffered-versus-streaming sweep. Three assertions fail against main's `src` and pass with the fix.

Standalone published-package repros: [bug](https://stackblitz.com/github/onmax/repros/tree/repro/hucre-464/hucre-464?startScript=build), [pnpm-patched fix](https://stackblitz.com/github/onmax/repros/tree/repro/hucre-464/hucre-464-fix?startScript=build).

`pnpm exec vitest run test/ods-third-party.test.ts` green — 19 tests. `pnpm lint`, `pnpm typecheck`, `pnpm build` and the packed-consumer verification are green too.

The full local `pnpm test` is not green: 10,530 tests pass, then the existing `html-import` 24-million-slot resource-bound test exceeds its 20-second timeout. It also times out when run alone; this PR does not touch that path.


productdevbook:
Merged. Thank you — this is the fixture #464 has been asking for since it was opened, and it broke the reader on arrival, which is exactly what a corpus is for.

I reviewed it rather than taking CI's word (no checks ran on the fork branch), and verified the claim independently before reading your fix.

## The bug reproduces, and it is worse than "wide"

Reading your fixture on `main`:

```
rows: 5
row widths: 16384 16384 16384 16384 16384
trailing nulls in row 0: 16379
```

Four columns of data, 16,384 columns returned. On a five-row file that is a curiosity. On a real one it is not: the bounding box is rows × 16,384, so a 100,000-row LibreOffice sheet describes 1.6 **billion** slots and `maxTotalCells` refuses to read it at all. The file would not open.

I had named this exact construct on #464 as the thing SheetJS never emits and LibreOffice always does — `table:number-columns-repeated="16379"` closing every row — and said it needed someone with the tool. You had the tool.

## What I checked on the fix

- **Both style modes.** 16,384 → 5 with `readStyles` off *and* on. That is the part I most wanted to see, since the condition turns on whether the style resolves.
- **Values, not just widths.** Strings, numbers, dates, booleans, a cached formula result and an error cell all come back correctly.
- **That it does not over-trim.** A trailing *empty* cell carrying a real fill: dropped with `readStyles: false`, kept with the fill intact when `true`. That is the #492 contract for XLSX, now matched here, and it is the regression a trim like this invites.
- **The merge, not the branch.** Merged locally onto main first: 228 files, 10,548 tests, format and typecheck clean.

## The binary, since it is one

13,458 bytes compressed, 44,323 uncompressed — a ratio of 3.3, so nothing hiding. Nine ordinary ODF parts, no traversal in any entry name, and `meta.xml` says `LibreOfficeDev/26.8.0.0.alpha0$MacOSX_AARCH64`, which is the provenance claim standing up on its own.

## One thing worth writing down somewhere

The fix works with `readStyles: true` because LibreOffice's padding cell names a style hucre cannot resolve — it lives in `styles.xml`, and `PARITY.md` records that the ODS reader opens `content.xml` and `meta.xml` only. So `styleDefs.has(...)` is false and the tail is trimmed.

That is correct today and it is load-bearing. If the reader ever learns to read `styles.xml` — a reasonable thing to want, since it is where LibreOffice keeps named and default cell styles — that padding becomes resolvable and 16,384 columns come back with `readStyles: true`. Not a change request; a note for whoever opens that door.

Issue #407, opened 2026-08-04
hucre's own XLSX output is not fully readable by hucre — eight write-only fields


productdevbook:
Part of #352. From the read/write parity audit.

Each of these is written by `writeXlsx` and then ignored by `readXlsx`. Write a workbook, read it back, and the value is gone — hucre does not round-trip its own output.

| field | written at | read | consequence |
| --- | --- | --- | --- |
| `ConditionalRule.style` (dxf) | `worksheet-writer.ts:1366` → `styles-writer.ts:371,560` | **no** — `dxfId` parsed then discarded, `worksheet.ts:1410` says so outright | a red-fill "greater than 5" rule survives read+rewrite; **its colours do not** |
| `Workbook.defaultFont` | `styles-writer.ts:262-271` | **no** — never assigned in `reader.ts` | always `undefined` on read |
| `Workbook.activeSheet` | `workbook-writer.ts:82-94` | **no** — `<workbookView activeTab>` has no case | the tab a file opens on is lost |
| `SheetImage.width` / `.height` | `drawing-writer.ts:129-130,172` | **partial** — read only from `oneCellAnchor` (`reader.ts:836-837`), and hucre always writes `twoCellAnchor` (`drawing-writer.ts:179`) | never recovered from hucre's own files |
| `SheetTextBox.width` / `.height` | `drawing-writer.ts:202-203,235` | **no** — `a:ext` never extracted | same |
| `TableDefinition.showAutoFilter: false` | correctly omits `<autoFilter>` (`table-writer.ts:34,56`) | **wrong** — `reader.ts:1573` initialises `true` and never clears it | `false` reads back as `true` |
| `pageSetup.printArea` / `.printTitlesRow` / `.printTitlesColumn` | `writer.ts:606-629` | **no** — resurface only as raw `_xlnm.Print_Area` / `_xlnm.Print_Titles` entries in `namedRanges` | the same setting has two unconnected representations depending on direction |
| `Chart` alt-text / frame title | `drawing-writer.ts:346-347` | **no** — graphicFrame `cNvPr` descr/title never read | accessibility metadata is write-only |

None of it is documented.

Two more in the same family, on the write side rather than the read side:

[...]

PR #559, opened 2026-08-13
refactor: one copy of the helpers that were copied, and comments that match the code

productdevbook:
A cleanup pass with no behaviour change. The commit message has the full
inventory; this is what is worth knowing about it.

## The claim is checked, not asserted

"No behaviour change" is the easiest thing in a refactor to believe and
not verify — and `pnpm test` cannot settle it on its own, because a
refactor that breaks something the suite does not cover looks exactly
like one that breaks nothing.

So the digests of what the writers produce were compared across the
change: `writeXlsx` (styled cells, CJK, number formats, a checkbox, a
formula, print titles, an auto-filter, frozen panes, a table with a
total row, a bar chart with a legend and data labels), `saveXlsx`,
`writeOds`, `writeCsv`, plus `parseCsv` and `streamXlsxRows` on the read
side. **Every digest is identical before and after.**

Dates had to be pinned to get there — `docProps/core.xml` and
`meta:creation-date` carry the clock, so the first run of the probe
disagreed with itself before it could disagree with `main`.

The readers are covered by the golden JSON fixtures, which compare
against files Excel and openpyxl wrote rather than against hucre's own
output.

Also re-run: the ODF and OOXML schema validators (#552, #553) — all
parts still valid — `verify-package.mjs`, and `pnpm size`, all within
budget.

## What it did

**Nine byte-identical copies of `findChild`**, plus four of
`childElements`, three of `resolvePath` / `dirname` / `findRIdAttr`, and
`parseIntSafe` under a fifth name as `parseIntOr`. Two full copies of
`buildNamedRanges` and `computeTableRange` — and `applyPrintDefinedNames`
in the reader depends on every writer deriving the print names from
`pageSetup` the same way (#407), so that pair was the copy worth removing
first.

The FeaturePropertyBag GUID, namespace, content type, part path and
relationship type were declared as constants in one module and then typed
out again as literals in five others.

## Comments that had stopped being true

This is the half that matters more than the deduplication, since a wrong
comment costs more than no comment.

`_style.ts` listed `styles.resolveStyle` among the deep-copy callers. It
is deliberately the opposite — the parsed font/fill/border records are
*shared*, because copying them per cell takes peak memory from 407 MB to
787 MB over 720,000 styled cells, and `docs/PARITY.md` states the sharing
as a contract. A reader who trusted that header would have "fixed" a
documented decision.

`chart/util.ts` claimed a consolidation that had only half happened —
two chart modules still carried private copies. `styles-writer.ts` had
the #437 snapshot-on-registration note orphaned above a function that
does no snapshotting. `writer.ts` labelled the FeaturePropertyBag part
`xl/vbaProject.bin`. Two comments still called the library `defter`.

## Found and deliberately not fixed

Reported rather than folded in, because none of them is a cleanup:

1. **`src/ods/writer.ts:1401`** — every ODS hucre writes stamps
   `<meta:generator>defter</meta:generator>`, the pre-rename project
   name, and a test pins it. Shipped output, so it is someone's call, not
   a refactor's.
2. **`src/xlsx/pivot-writer.ts:648`** — a local `parseCellRef` that
   *throws* on a bad ref shadows the name of the exported, deliberately
   lenient one in `xlsx/worksheet.ts`. Same name, opposite error
   contract.
3. **`decodeUtf8` means two different things** — via `decodePart` (with
   the #514 `MAX_STRING_LENGTH` ceiling) in the buffered readers, a bare
   `TextDecoder` in the streaming ones. That matches PARITY
   ("`streamXlsxRows` still has no ceiling") and is intentional; the
   shared name hides it.

Left alone as too big for a cleanup pass: `parseWorkbookXml` exists in
both the buffered and streaming readers (a genuine subset, but the
`<sheets>` and `date1904` handling is duplicated and can drift), and the
`chart/*` modules carry three identical 49-line `build*SpPr` bodies, nine
identical font-family normalizers and nine tri-state clone resolvers.
Each of those carries host-specific documentation; consolidating them is
a project with its own risk, not a tidy-up.


productdevbook:
Correcting the verification described above, which claimed more coverage than it had.

The probe's styled sheet put its cells inline in `rows`:

```ts
rows: [[{ value: "Header", style: { font: { bold: true, size: 14 } } }, …]]
```

That shape is **dropped** by the buffered writers — the finding that became #562. So the sheet I described as "styled cells, CJK, number formats, a checkbox, a formula" was written as a grid of empty cells, and the digests I compared were digests of a nearly empty document. The style writer, the shared-string table and the checkbox path were not exercised at all.

## Re-run, with cells that land

Same comparison, `ad0f4a5` (the commit before this PR) against `db04ead` (the merge), with the styles moved into the `cells` map where the writer reads them. Verified they land first: `<cellXfs count="4">` and `s="1"` / `s="2"` / `s="3"` on the cells that asked.

**The digests are identical.** So the conclusion holds — but it holds on evidence now rather than on the evidence I cited.

## And a limitation worth stating

A mutation check on the shared `findChild` — `c.local` → `c.tag` — turns **1,409 tests red** but leaves the probe's digests unchanged: for unprefixed elements the two are the same string, and the fixtures the probe writes and re-reads are unprefixed.

So the probe pins the writers; the suite and the golden fixtures pin the readers. Neither covers the other, and the probe on its own would not have caught a broken `findChild`.

PR #571, opened 2026-09-04
feat(xlsx): support autoFilter and conditionalRules in stream writer

ryuhzk:
## Summary

The buffered XLSX writer (`writeXlsx`) already supports auto-filters and conditional formatting rules on worksheets, but the streaming writers (`writeXlsxStream` / `writeXlsxStreamSheets`) did not expose the same options. Streaming exports that need filter dropdowns or conditional formatting had to fall back to the buffered path.

This change brings stream-writer parity for those two sheet features:

- Export `serializeConditionalFormatting` from `worksheet-writer.ts` so the stream path can reuse the same XML serialization as the buffered writer.
- Add `autoFilter` and `conditionalRules` to `StreamWriterOptions` and `XlsxStreamSheet`.
- Pass those options through `writeXlsxStream` into the multi-sheet streaming path.
- Emit `<autoFilter>` and `<conditionalFormatting>` in the worksheet tail (first sheet part only), in ECMA-376 order relative to `mergeCells`.

## Test plan

- [x] Added coverage in `test/xlsx-stream-write.test.ts` asserting the streamed worksheet XML contains `<autoFilter ref="...">` and `<conditionalFormatting>`, and that `readXlsx` round-trips both.
- [x] `bun test test/xlsx-stream-write.test.ts test/stream-writer-multi-sheet.test.ts`
- [x] `pnpm exec vitest run` on the related stream / conditional-formatting suites

## Summary by CodeRabbit

* **New Features**
  * Streaming XLSX exports now support auto-filters at the workbook and worksheet levels.
  * Auto-filter column criteria are included in exported files.
  * Streaming XLSX exports now support conditional formatting rules.
  * Auto-filters and conditional formatting are preserved when exported files are read back.
* **Bug Fixes**
  * Improved serialization of worksheet filters and conditional formatting for more reliable exported files.

### DIFF
--- a/src/xlsx/stream-writer.ts
+++ b/src/xlsx/stream-writer.ts
@@ -33,6 +33,7 @@ import {
   cellRef,
   hasRowAttributes,
   rowAttributes,
+  serializeAutoFilter,
   serializeCell,
   serializeConditionalFormatting,
   type ResolvedCell,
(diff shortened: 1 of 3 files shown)


ryuhzk:
Addressed CodeRabbit autoFilter columns feedback in e325d2a: stream writer now reuses shared serializeAutoFilter (emits filterColumn children when configured) and has a regression test in test/xlsx-stream-write.test.ts.

